Issue #14, opened 2025-06-27
Proposal: Multiple standardized enviroments / Dockerfile's


ItsWendell:
The current Dockerimage provided for the Sandbox SDK only contains bun, sandboxes with a wider rage of tools / libraries available could be highly valuable too.

I personally created a draft of a extended Dockerfile that also includes Python, including a couple of common python packages often used by ChatGPT, Antropic / Google's sandbox for running python, and added utilities to instantly invoke given code

This is useful for people needing quick solution for code execution within their LLMs apps, even better, would an additional MCP server and client that runs within the Dockerfile or within the container's durable object for instant utility.

. Having a small set of presets for Dockerfiles could be useful for the Sandbox SDK.

Resources:
https://platform.openai.com/docs/assistants/tools/code-interpreter
https://ai.google.dev/gemini-api/docs/code-execution
https://docs.anthropic.com/en/docs/agents-and-tools/tool-use/code-execution-tool

Example Dockerfile:

```Dockerfile
FROM oven/bun:alpine AS builder

## Add build dependencies and Python
RUN apk add --no-cache \
    python3 \
    py3-pip \
    build-base \
    gcc \
    g++ \
    musl-dev \
    linux-headers \
    libffi-dev \
    openssl-dev \
    python3-dev \
    jpeg-dev \
    zlib-dev \
    freetype-dev \
    lcms2-dev \
    openjpeg-dev \
    tiff-dev \
    harfbuzz-dev \
    fribidi-dev \
    proj-dev \
    geos-dev \
    gdal-dev

## Install uv and Python packages with balanced optimization for LLM sandbox
COPY --from=ghcr.io/astral-sh/uv:latest /uv /uvx /bin/
RUN uv pip install --system --break-system-packages --no-cache --no-deps \
    # Essential data science core
    numpy pandas matplotlib scipy \
    # File processing essentials
    pillow openpyxl xlrd pyarrow \
    # Document generation (important for LLM tasks)
    fpdf pylatex reportlab PyPDF2 python-docx python-pptx \
    # Web and markup processing
    lxml jinja2 \
    # Image processing
    imageio \
    # Utilities
    python-dateutil pytz tqdm joblib \
    # Math and validation
    sympy jsonschema \
    # Basic packages
    attrs six packaging && \
    # Install packages with dependencies (useful for LLM tasks)
    uv pip install --system --break-system-packages --no-cache \
    # ML and stats (users might request these)
    scikit-learn seaborn \
[...]

PR #15, opened 2025-06-27
add multiple standardized Docker environments

whoiskatrin:
Implements https://github.com/cloudflare/sandbox-sdk/issues/14 by adding structured Docker environment options with Alpine-optimized multi-stage builds.

### DIFF
--- a/dockerfiles/README.md
+++ b/dockerfiles/README.md
@@ -1,0 +1,113 @@
+# Sandbox Docker Environments
+
+This directory contains multiple Docker environment configurations for different use cases.
+
+## Available Environments
+
+### Default Environment
+- **File**: `../packages/sandbox/Dockerfile`
+- **Size**: ~200MB
+- **Languages**: JavaScript, TypeScript
+- **Use Cases**: Basic JS/TS execution, lightweight applications
+
+### Python Data Science
+- **File**: `python-data-science.dockerfile`
+- **Size**: ~2.5GB
+- **Languages**: Python, JavaScript, TypeScript
+- **Packages**: numpy, pandas, matplotlib, scikit-learn, torch, transformers, jupyter
+- **Use Cases**: 
+  - Data analysis and visualization
+  - Machine learning model training
+  - Jupyter notebook execution
+  - AI/ML prototyping
+
+### Node Extended
+- **File**: `node-extended.dockerfile`
+- **Size**: ~1.2GB
+- **Languages**: JavaScript, TypeScript, Python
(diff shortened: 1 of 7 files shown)


ghostwriternr (commented):
I'm at least a little bit hesitant of this approach, mainly because the possibilities of what someone integrating with the sandbox might need is endless - in terms of languages, libraries, SDKs, toolkits, etc.

My preference would be to do 2 things:
1. We come up with an easy way for people to generate and use their own extensions of the base Docker image - actually, we already have a version of it. Just completely undocumented. The [`Dockerfile` used in the basic example](https://github.com/cloudflare/sandbox-sdk/blob/037c848f961b372de20c1403f8723cd9b3cb0e8f/examples/basic/Dockerfile) extends the base image — and possibly be customised however someone might need — and referenced in the [wrangler config](https://github.com/cloudflare/sandbox-sdk/blob/037c848f961b372de20c1403f8723cd9b3cb0e8f/examples/basic/wrangler.jsonc). Our current way to do this is to directly extend the Dockerfile.
2. We can then go a step further to simplify this process either using a CLI tool (like [E2B](https://e2b.dev/docs/quickstart/install-custom-packages)) or via the SDK itself (like [Daytona](https://www.daytona.io/docs/declarative-builder) or [Modal](https://modal.com/docs/guide/images)), which can be handy for on-the-fly enhancement of the images.

I do think it could still be handy to provide pre-configured images like this PR does that anyone can pick up and use as the base - these could possibly come from the most common use-cases sandboxes are being used for, and identifying the common toolkit there. But I'd avoid making this environment/configuration driven.

ghostwriternr:
I just went through the links mentioned in #14 from OpenAI, Google and Anthropic. All of them seem to be doing the same combination of Python environment + a finite list of pre-installed libraries. So I have a feeling that this combination is being used much more than I'm personally privy to, so having those pre-published images as a base seems like a good thing to have.

I think my previous comment still remains, but if that approach is one we can agree upon, I think this PR is good to merge once:
1. We document the existing method, so everyone is aware of our current already-supported way to customise the base image.
2. We update the README to remove the config file / environment variables, and instead we publish the new images and we document the repository names for them.

threepointone:
clsoing this PR while we figure out an alternate approach here

anthonywu:
Just bundle `uv` and maybe `cargo` and other reasonable package managers.

+1 to keeping the base image finely-scoped and just published _tested_ cookbooks of more _fully featured images_.

Perhaps first-party variants of `sandbox-python` `sandbox-go` `sandbox-rust` etc

PR #27, opened 2025-07-25
remove yarn/pnpm from the image

threepointone:
(just to have a conversation about things we could potentially trim from the image)

the additional package managers seem like overkill, and don't have much value over npm+bun (especially because we don't have persistent volumes/caches)

### DIFF
--- a/packages/sandbox/Dockerfile
+++ b/packages/sandbox/Dockerfile
@@ -56,8 +56,6 @@ RUN curl -fsSL https://bun.sh/install | bash \
     && mv /root/.bun/bin/bunx /usr/local/bin/bunx \
     && rm -rf /root/.bun
 
-# Install global npm packages as root
-RUN npm install -g yarn pnpm
 
 # Set up working directory
 WORKDIR /app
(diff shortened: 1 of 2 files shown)


packages/sandbox/Dockerfile
```diff
@@ -56,8 +56,6 @@ RUN curl -fsSL https://bun.sh/install | bash \
     && mv /root/.bun/bin/bunx /usr/local/bin/bunx \
     && rm -rf /root/.bun
```
threepointone, line 57: btw why are these binaries moved?

ghostwriternr: This comment points to an older version of the Dockerfile, but the reason we do this move in our latest version:
```Dockerfile
COPY --from=bun-source /usr/local/bin/bun /usr/local/bin/bun
COPY --from=bun-source /usr/local/bin/bunx /usr/local/bin/bunx
```
is because we get the pre-built bun binaries from bun's official docker image as base:
```Dockerfile
FROM oven/bun:latest AS bun-source
```

The bun image is otherwise not used, so we're just moving this into our 'layer'

ghostwriternr (approved):
Yea, these were just opinionated defaults that I had thrown in - but there is indeed no set of tools that we can bake in that will satisfy every use-case. This is why I quite like that developers can just extend our base docker image as they need.

In terms of trimming down further, maybe we should publish a `cloudflare/sandbox:<version>-base` image which only contains the bare minimum needed for the bun server to function and let the developer add whatever they need.

PR #35, opened 2025-07-26
fixes / formats

threepointone:
This PR looks big, but that's just because I ran prettier over it. The actual small changes are to the text dockerfile, and the readme, that I've marked out.

### DIFF
--- a/examples/basic/Dockerfile
+++ b/examples/basic/Dockerfile
@@ -12,6 +12,3 @@ FROM cloudflare/sandbox-test:0.1.3
 # expose any ports you might want to use (necessary for local dev)
 # EXPOSE 8080
 # EXPOSE 3001
-
-# # Run the same command as the original image
-# CMD ["bun", "index.ts"]
--- a/examples/basic/Dockerfile
+++ b/examples/basic/Dockerfile
@@ -8,8 +8,10 @@ FROM cloudflare/sandbox-test:0.1.3
 # arm64 build of the image.
 # FROM --platform=linux/arm64 cloudflare/sandbox-test:0.1.3
 
-EXPOSE 8080
-EXPOSE 3001
 
-# Run the same command as the original image
-CMD ["bun", "index.ts"]
+# expose any ports you might want to use (necessary for local dev)
+# EXPOSE 8080
+# EXPOSE 3001
+
+# # Run the same command as the original image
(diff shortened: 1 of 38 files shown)


examples/basic/Dockerfile
```diff
@@ -8,8 +8,7 @@ FROM cloudflare/sandbox-test:0.1.3
 # arm64 build of the image.
 # FROM --platform=linux/arm64 cloudflare/sandbox-test:0.1.3
 
-EXPOSE 8080
-EXPOSE 3001
 
-# Run the same command as the original image
-CMD ["bun", "index.ts"]
```
threepointone, line 15: this shouldn't be necessary anymore, yeah?

packages/sandbox/README.md
```diff
@@ -66,10 +66,10 @@ npm install @cloudflare/sandbox
 ```dockerfile
 FROM docker.io/cloudflare/sandbox:0.1.3
 
-EXPOSE 3000
+# expose any ports you might want to use
+# EXPOSE 8080
+# EXPOSE 5173
 
-# Run the same command as the original image
-CMD ["bun", "index.ts"]
```
threepointone, line 72: like the example, this shouldn't be necessary, yes?

packages/sandbox/README.md
```diff
@@ -284,7 +283,7 @@ The SDK handles:
 
 ```dockerfile
 # In your Dockerfile (only needed for local dev)
-FROM oven/bun:latest
```
threepointone, line 287: this should point to our image, yes?

ghostwriternr:
Closing this PR in favour of https://github.com/cloudflare/sandbox-sdk/pull/46 which addresses the comments. Only reason to not merge this as-is is to avoid massive merge conflicts with #43 when it lands next week.

examples/basic/Dockerfile
ghostwriternr: Good catch! I didn't realise it works without this.

packages/sandbox/README.md
ghostwriternr: 🙌

packages/sandbox/README.md
ghostwriternr: Woops, this was a typo. Fixed in #46.

PR #66, opened 2025-08-24
Determine if the port is specified in the URL.

### LINKED ISSUE #65: sandbox.fetch does not respect port in request.url.
Caveat: I'm making some assumptions about the code here and the intentions of the code. This bug report may not be entirely accurate, but the result is that I am unable to call a different port in my container from the sandbox class.. 

In the Sandbox's `fetch` method we have the following line:

https://github.com/cloudflare/sandbox-sdk/blob/603d05faf0d32f179aec37c4f78afe37cb30f830/packages/sandbox/src/sandbox.ts#L125

Which passes`request.url` to `determinePort`

https://github.com/cloudflare/sandbox-sdk/blob/603d05faf0d32f179aec37c4f78afe37cb30f830/packages/sandbox/src/sandbox.ts#L131-L141

As an example, imagine we have the following url: `http://localhost:8910/health`. One would expect that port `8910` is extracted, but unfortunately it'll be `3000`

peterp:
The port can be specified in the URL. This change allows us to extract the port and parse it to a number.

Closes #65.

### DIFF
--- a/packages/sandbox/src/sandbox.ts
+++ b/packages/sandbox/src/sandbox.ts
@@ -135,6 +135,10 @@ export class Sandbox<Env = unknown> extends Container<Env> implements ISandbox {
       return parseInt(proxyMatch[1]);
     }
 
+    if (url.port) {
+      return parseInt(url.port);
+    }
+
     // All other requests go to control plane on port 3000
     // This includes /api/* endpoints and any other control requests
     return 3000;


ghostwriternr (approved):
Sorry this has taken forever to review! This looks great — thank you for the fix 🙏

PR #90, opened 2025-10-13
set bun idletimeout to max value

eastlondoner:
Set bun idletimeout so that requests aren't dropped if they don't do anything for 10s.

This occurs when we use execStream to do stuff that needs thinking time

https://bun.com/docs/api/http#idletimeout

note - It would also be a very good idea to handle the case where the collection is closed by the server, that currently seems to fail essentially silently


eastlondoner:
see also: https://github.com/cloudflare/containers/pull/100

PR #106, opened 2025-10-16
fix examples failing to deploy and prevent committing node_modules

jahands:
When running this command from the docs:

```sh
npm create cloudflare@latest -- my-sandbox --template=cloudflare/sandbox-sdk/examples/minimal
```
we get this error:
> ✘ [ERROR] Image too large: needs 2346MB, but your app is limited to images with size 2000MB. Your need more disk for this image.

It also commits all node_modules due to lack of .gitignore's in these examples.

This PR fixes both of these issues

### DIFF
--- a/.changeset/angry-pants-tell.md
+++ b/.changeset/angry-pants-tell.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/sandbox-sdk": patch
+---
+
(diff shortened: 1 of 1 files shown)


ghostwriternr:
Thanks for trying so soon, and putting up a fix!

PR #114, opened 2025-10-18
Debloat base docker image (2.63GB → 1.03GB)

ghostwriternr:


packages/sandbox/Dockerfile
```diff
@@ -108,13 +82,6 @@ RUN curl -fsSL https://deb.nodesource.com/setup_20.x | bash - \
 
 # Install Bun runtime from official image
 COPY --from=oven/bun:1 /usr/local/bin/bun /usr/local/bin/bun
-COPY --from=oven/bun:1 /usr/local/bin/bunx /usr/local/bin/bunx
-
-# Install development tools globally
-RUN npm install -g \
```
whoiskatrin, line 114: if we remove them, it defeats the purpose of adding them. we should find a way to keep wrangler, vite and opencode

ghostwriternr: Agreed, but its taking up 500mb. I'd instead have prominent docs that support how easy it is to add them if they need.
OR, we could also do multiple versions of the base image. I want to do a `-minimal` version that does not have any of the python stuff in case the user doesn't want the `runCode` functionality. Similarly, we could do a `-full` version that includes all these deps.
The "cost" of adding these deps (which I didn't realise earlier) is that it doesn't fit inside the smallest image size on our containers platform, so this is making the sandbox more expensive to use from the get go.

threepointone: yeah it's fine to remove it for now, we don't have a full story yet for it, but it's a thing to solve immediately after

PR #121, opened 2025-10-20
store the session on the storage

whoiskatrin:
Changes
- Load session ID from storage on initialization 
- Persist session ID when created
- Gracefully handle existing sessions

### DIFF
--- a/.github/workflows/pkg-pr-new.yml
+++ b/.github/workflows/pkg-pr-new.yml
@@ -8,6 +8,7 @@ on:
   pull_request:
     types: [opened, synchronize, reopened]
 
+
 jobs:
   publish-preview:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/pkg-pr-new.yml
```diff
@@ -0,0 +1,44 @@
...
+        run: npm ci
+
+      - name: Build packages
+        run: npm run build
+
+      - name: Resolve workspace dependencies
+        run: npx tsx .github/resolve-workspace-versions.ts
+
+      - name: Publish to pkg.pr.new
+        run: npx pkg-pr-new publish './packages/sandbox'
```
ghostwriternr, line 44: Modify this workflow to also include the docker publish, so that preview packages also come with a corresponding image that can be used for testing.

ghostwriternr (approved):
Added a comment regarding the new preview package workflow, but the main change looks good and can be merged. We can fix the workflow in a separate PR.

PR #128, opened 2025-10-21
Use single instance when deploying minimal example

ghostwriternr:


ghostwriternr:
Without this, when trying out all the tutorials one by one, users run into account-level limits.

PR #136, opened 2025-10-23
Add missing LICENSE file

malthejorgensen:
This file is referenced in README.md but is missing in the repository.

_Unclear if this file should really be at `packages/sandbox/LICENSE` ([link to dir]) and at the root it should be a symlink to similar to how `README.md` works at the root?_

Okay – made a change – I've made it a symlink similar to how README.md works.

[link to dir]: https://github.com/cloudflare/sandbox-sdk/tree/main/packages/sandbox

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,1 @@
+packages/sandbox/LICENSE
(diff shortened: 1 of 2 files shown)


ghostwriternr:
Thanks for bringing this up! I'd love to formalise a LICENSE so there's no confusion for users and contributors. I'll close this PR in favour of #208 which updates this across the project.

PR #137, opened 2025-10-23
add keepAlive flag to prevent container from shutting down for long processes 

whoiskatrin:
keepAlive flag that prevents containers from sleeping while processes are running. If used this would require calling destroy() manually afterwards

### DIFF
--- a/.changeset/wet-falcons-hang.md
+++ b/.changeset/wet-falcons-hang.md
@@ -2,4 +2,4 @@
 "@cloudflare/sandbox": patch
 ---
 
-add auto timeout renewal and some health checks
+add keepAlive flag to prevent containers from shutting down


.github/workflows/claude-code-review.yml
```diff
@@ -22,46 +22,33 @@ jobs:
...
-               ```bash
-               gh pr comment ${{ github.event.pull_request.number }} --body "YOUR_REVIEW_CONTENT_HERE" --repo ${{ github.repository }}
-               ```
-
-            Start your review with "## Claude Code Review" heading so it can be identified and updated on subsequent runs.
-
-          # See https://github.com/anthropics/claude-code-action/blob/main/docs/usage.md
-          # or https://docs.claude.com/en/docs/claude-code/cli-reference for available options
-          claude_args: '--allowed-tools "Bash(gh issue view:*),Bash(gh search:*),Bash(gh issue list:*),Bash(gh pr comment:*),Bash(gh pr diff:*),Bash(gh pr view:*),Bash(gh pr list:*),Bash(gh api:*)"'
+    # steps:
```
ghostwriternr: Can we bring this back? The main branch has already made changes to:
1. Keep Claude Code reviews be a lot more succinct
2. Get Claude Code to keep updating the same comment instead of new comments on every new commit within a PR

packages/sandbox/src/sandbox.ts
```diff
@@ -220,6 +235,22 @@ export class Sandbox<Env = unknown> extends Container<Env> implements ISandbox {
     this.logger.error('Sandbox error', error instanceof Error ? error : new Error(String(error)));
   }
 
+  /**
```
ghostwriternr, line 238: This is actually v clean! Nice

tests/e2e/test-worker/index.ts
```diff
@@ -86,6 +93,51 @@ export default {
         });
       }
 
+      // Command execution with callback-based streaming (for long-running commands)
```
ghostwriternr: Is this still needed?

.github/workflows/claude-code-review.yml
whoiskatrin: Fixed

tests/e2e/test-worker/index.ts
whoiskatrin: Fixed

PR #140, opened 2025-10-23
fix: preserve newlines in heredoc commands for exec()

### LINKED ISSUE #139: Fix `exec` for heredoc commands
Currently any command that uses heredoc format will hang when using exec because of the line [here](https://github.com/cloudflare/sandbox-sdk/blob/main/packages/sandbox-container/src/session.ts#L711):
```ts
script += `  { ${command}; } > >(while IFS= read -r line || [[ -n "$line" ]]; do printf '\\x01\\x01\\x01%s\\n' "$line"; done >> "$log") 2> >(while IFS= read -r line || [[ -n "$line" ]]; do printf '\\x02\\x02\\x02%s\\n' "$line"; done >> "$log")\n`;
```

I think it's because bash expects the heredoc terminator to be alone on its line and we're putting the user command in a single physical line. So I think a simple fix would be to format the commands to make sure they start/end with new lines?

deathbyknowledge:
Fixes heredoc commands hanging in foreground execution by wrapping commands with newlines instead of placing them on a single line. Heredoc terminators must appear alone on their own line to work correctly.

Fixes #139

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/.changeset/ten-mangos-cover.md
+++ b/.changeset/ten-mangos-cover.md
@@ -1,0 +1,6 @@
+---
+"@repo/sandbox-container": patch
+"@cloudflare/sandbox": patch
+---
+
+fix: preserve newlines in heredoc commands for exec()


ghostwriternr:
Closing as obsolete. The heredoc bug this PR fixes was already resolved in PR #185, which replaced process substitution with direct file redirects. Thanks for the PR regardless!

PR #156, opened 2025-10-27
Fix WebSocket connections through exposed ports

ghostwriternr:
WebSocket upgrade requests through exposed sandbox ports were failing during the connection handshake. The routing layer was not correctly handling the WebSocket protocol upgrade.

This fixes the routing to support WebSocket connections to services running in sandboxes.

### DIFF
--- a/.changeset/fix-websocket-routing.md
+++ b/.changeset/fix-websocket-routing.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/sandbox": patch
+---
+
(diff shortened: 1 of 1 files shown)


riderx:
@whoiskatrin that mean the websocket is ready?

whoiskatrin:
> @whoiskatrin that mean the websocket is ready?

Half way there, there will be another PR in the next few days.

riderx:
Lovely !

PR #168, opened 2025-10-29
Fix type generation

### LINKED ISSUE #165: Type definitions from `@repo/shared` are not accessible to consumers
# Type definitions from `@repo/shared` are not accessible to consumers

## Description

The `@cloudflare/sandbox` package re-exports type definitions from `@repo/shared`, but this package is not available to consumers because it's listed as a `devDependency` rather than a regular dependency, and it's not published to npm.

## Current Behavior

In `dist/index.d.ts`:
```typescript
export * from '@repo/shared';
export { BaseExecOptions, ExecEvent, ExecOptions, ExecResult, FileChunk, FileMetadata, FileStreamEvent, GitCheckoutResult, ISandbox, LogEvent, PortCloseResult, PortExposeResult, PortListResult, Process, ProcessCleanupResult, ProcessInfoResult, ProcessKillResult, ProcessListResult, ProcessLogsResult, ProcessOptions, ProcessStartResult, ProcessStatus, StartProcessRequest, StreamOptions, isExecResult, isProcess, isProcessStatus } from '@repo/shared';
```

In `package.json`:
```json
{
  "devDependencies": {
    "@repo/shared": "^0.0.0"
  }
}
```

## Impact

When consuming the `@cloudflare/sandbox` package, TypeScript cannot resolve these re-exported types because `@repo/shared` is:
1. Not installed (it's a devDependency)
2. Not published to npm (it's an internal monorepo package)

This results in:
- Loss of type safety for methods like `listFiles()` which return types from `@repo/shared` (e.g., `ListFilesResult`)
- Consumers must use `any` types or manually recreate the type definitions
- TypeScript errors when trying to import re-exported types

## Steps to Reproduce

1. Install `@cloudflare/sandbox` in a project
2. Try to use types from the package:
   ```typescript
   import { Sandbox, ListFilesResult } from '@cloudflare/sandbox';

   const sandbox: Sandbox;
   const files: ListFilesResult = await sandbox.listFiles('/');
   ```
3. TypeScript will fail to resolve `ListFilesResult` and related types

## Expected Behavior

All type definitions should be accessible to consumers of the package without requiring unpublished dependencies.

## Environment

- `@cloudflare/sandbox` version: 0.4.11
- TypeScript version: 5.x
- Package manager: pnpm/npm/yarn

## Workaround

Currently, consumers must manually define the types locally:

```typescript
interface FileInfo {
  path: string;
  name: string;
  isDirectory: boolean;
  size?: number;
  modifiedAt?: number;
}
```

threepointone:
We inline types from `@repo/shared` so that it includes the types we reexport. Fixes #165

Updated tsup to tsdown, and normalized some deps. Added typecheck steps to example repos (except basic, so that CI passes for this one). I'll do more monorepo maintenance after this change lands.

### DIFF
--- a/examples/claude-code/package.json
+++ b/examples/claude-code/package.json
@@ -8,10 +8,11 @@
 		"deploy": "wrangler deploy",
 		"dev": "wrangler dev",
 		"start": "wrangler dev",
-		"types": "wrangler types"
+		"types": "wrangler types",
+		"typecheck": "tsc --noEmit"
 	},
 	"devDependencies": {
-		"@cloudflare/sandbox": "^0.4.3",
+		"@cloudflare/sandbox": "*",
 		"@types/node": "^24.9.1",
 		"typescript": "^5.8.3",
 		"wrangler": "^4.44.0"
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -10,5 +10,5 @@
   "access": "public",
   "baseBranch": "main",
   "updateInternalDependencies": "patch",
-  "ignore": ["@cloudflare/sandbox-*"]
+  "ignore": ["@cloudflare/sandbox-*", "@repo/*"]
 }
--- a/.changeset/giant-paths-enjoy.md
+++ b/.changeset/giant-paths-enjoy.md
@@ -1,0 +1,7 @@
+---
+"@cloudflare/sandbox": patch
+---
+
+Fix type generation
+
+We inline types from `@repo/shared` so that it includes the types we reexport. Fixes #165
--- a/examples/code-interpreter/package.json
+++ b/examples/code-interpreter/package.json
@@ -8,17 +8,18 @@
     "deploy": "wrangler deploy",
     "dev": "wrangler dev",
     "start": "wrangler dev",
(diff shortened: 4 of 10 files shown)


examples/claude-code/package.json
```diff
@@ -8,10 +8,11 @@
 		"deploy": "wrangler deploy",
 		"dev": "wrangler dev",
 		"start": "wrangler dev",
-		"types": "wrangler types"
+		"types": "wrangler types",
+		"typecheck": "tsc --noEmit"
 	},
 	"devDependencies": {
-		"@cloudflare/sandbox": "^0.4.3",
+		"@cloudflare/sandbox": "*",
```
ghostwriternr, line 15: Ideally, nothing under `examples` should use the sandbox package as a workspace dependency because we want users to be able to run these examples using C3 ([as mentioned in our docs](https://developers.cloudflare.com/sandbox/get-started/#1-create-a-new-project)).

ghostwriternr: Resolving because I've now realised `*` would just pick up latest version, which makes sense in this case.

PR #182, opened 2025-10-31
feat: use Node.js LTS (v24)

o-az:
As you know Node.js v24 ([LTS](https://nodejs.org/en/about/previous-releases#looking-for-the-latest-release-of-a-version-branch)) comes with builtin TypeScript support. In a project I'm currently working on I need to use Node.js and need to run TypeScript - so being able to run `.ts` files without needing extra dependencies like `tsx` would make a huge difference.

image

- I tested with [`node:lts-alpine`](https://hub.docker.com/layers/library/node/lts-alpine/images/sha256-a3955307c0ddda257f0c8c5f1442017838cfa6d0e77ea61912be6d38f4abcd6d) and Node.js v24 from [nodesource.com](https://nodesource.com/products/distributions) to verify it works.

### DIFF
--- a/.changeset/neat-drinks-own.md
+++ b/.changeset/neat-drinks-own.md
@@ -1,5 +1,0 @@
----
-"@cloudflare/sandbox": patch
----
-
-Fix foreground commands blocking on background processes
--- a/.changeset/rude-years-wink.md
+++ b/.changeset/rude-years-wink.md
@@ -1,5 +1,0 @@
----
-'@cloudflare/sandbox': patch
----
-
-update python to 3.11.14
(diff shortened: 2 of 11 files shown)


ghostwriternr:
Thanks for this PR, @o-az ! Upgrading from Node.js 20 makes sense given it's approaching EOL in April 2026, but at this time I'd prefer to upgrade to Node.js 22, not Node.js 24. The goal of this SDK is to give a safe default for _most user code_, while offering easy upgrade paths for anyone else. Considering the current LTS is 22, I think that perhaps offers a better default. But at this time, since your need is mainly around automatic Typescript support and I haven't seen a lot of other issues around 20 being the default, I'm leaning towards closing this PR.

If this reasoning feels off, feel free to open a separate issue and we can discuss more with other users too.

PR #183, opened 2025-10-31
build Python 3.11.14

whoiskatrin:
Downloading 3.11.14 directly from the source, image size is 1GB now


jsj:
Verified this fixed my problem 

```
docker run --rm cloudflare/sandbox:0.0.0-pr-183-4d9a1d2 bash -lc "python3 -m pip install --quiet openai-agents && python3 - <<'PY'
from agents import Agent
print('openai-agents import ok')
PY"
# openai-agents import ok
```

PR #186, opened 2025-11-03
expose deleteSession

whoiskatrin:
- expose `deleteSession(sessionId)` method
- prevent deletion of default session (throws error with guidance to use `sandbox.destroy()`)
- session cleanup kills all running commands in parallel before destroying shell
- return structured `SessionDeleteResult` with success status, sessionId, and timestamp

### DIFF
--- a/.changeset/wild-news-pretend.md
+++ b/.changeset/wild-news-pretend.md
@@ -1,7 +1,7 @@
 ---
-"@repo/sandbox-container": patch
-"@cloudflare/sandbox": patch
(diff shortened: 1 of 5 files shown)


ghostwriternr (changes requested):
On a quick scan, this is looking good so far! The only gap is perhaps that we're not yet killing running commands (see `packages/sandbox-container/src/session.ts`) and cleaning out all related resources (labellers and pid files). Much of that logic should be present in this one file.

Issue #81, opened 2025-09-23
Passing a custom timeout to exec does not work


ghostwriternr:

mgomes:
Confirmed I am also running into this issue.

whoiskatrin:
@mgomes should be fixed, releasing soon

Issue #16, opened 2025-06-27
Proposal: cwd support to sandbox class


tlgimenes:
Hi folks,

I'm building an online bundler. The idea is that I'll spin a container to write files, npm install and bundle. 

Currently, I'm creating a new folder for each build with the following code

```
const sandbox = getSandbox(c.SANDBOX, 'my-sandbox')

// copy files with sanbox.writeFile

// install deps and build
await sandbox.exec('bun', ['install', --cwd, "path/to/the/build/folder"])
await sandbox.exec('bun', ['build', --cwd, 'path/to/the/build/folder'])
```

This works great, however, I need to, somehow, have a way to `cwd` to a given folder for building so I can remove this `--cwd` to the bun command.

Is this the intended usage? Should I spin one container for each build instead or this folder approach is ok?

thanks!

ItsWendell:
Since the resources available for a given instance are limited (bigger instances available after beta), I would recommend a single instance for each build.

You could pass a random uuid `crypto.randomUuid()` as an build id for - example for each build.

I would also recommend discussions here on GitHub or the Cloudflare Discord for questions like these.

https://discord.gg/cloudflaredev

threepointone:
we already have lightweight support for sessions, we haven't exposed it correctly in the api just yet. so you should be able to call multipole commands in a row, including `cd`

whoiskatrin:
This should work now:
```
 await sandbox.exec('bun', ['install'], { cwd: 'path/to/the/build/folder' })
 await sandbox.exec('bun', ['build'], { cwd: 'path/to/the/build/folder' })
```

PR #204, opened 2025-11-11
add environment variables and working directory support to command exec

whoiskatrin:


ghostwriternr:
(Sharing some thoughts here about `env` and `cwd` as I go, before adding more specific feedback. Literally me "think out loud"-ing haha)

Related issues: #16, #144

Overall, I quite like the features this PR is adding and think this is an important one to clean up and get right. It's been fragile for a long time and I'm happy we're fixing this!

So with env, I generally want us to make sure there are 4 layers of env setting and all of them work cohesively:
1. ENV variable when extending our published `Dockerfile`
2. Using `setEnvVars` on the sandbox object returned by `getSandbox`
3. Using `setEnvVars` on the session object returned by `createSession`
4. Passing the `env` parameter on methods like `exec`, `execStream`, `startProcess`, etc.

And env is mainly critical for all the flows that involve the default session and the ones created explicitly using `createSession`, where we maintain persistent bash sessions. We inherit whatever is set using ENV on the Dockerfile via `process.env` when we create a shell session. And "intuitively", I think `setEnvVars` on the sandbox object (_global_) should possibly set the variables on all the sessions created explicitly, whereas `setEnvVars` on the session object (_local_) should only affect that particular bash session.

cwd is much easier to reason with as an option, and is definitely quite useful when working in the scope of either the default session or the manually created session, but sometimes developers just want to run a command elsewhere and don't want to do `cd x`, `command_to_run` then `cd back_to_where_I_was`, which can get annoying.

packages/sandbox-container/src/session.ts
```diff
@@ -639,6 +644,24 @@ export class Session {
...
+    if (env && Object.keys(env).length > 0) {
+      const exports = Object.entries(env)
+        .map(([key, value]) => {
+          // Escape the value for safe shell usage
+          const escapedValue = value.replace(/'/g, "'\\''");
+          return `export ${key}='${escapedValue}'`;
+        })
+        .join('; ');
+      // Wrap in subshell to isolate env vars (they don't persist in session)
+      commandWithEnv = `(${exports}; ${command})`;
```
ghostwriternr: There's probably a silent bug here, because if the command is a `cd`, it won't get persisted because we're running in a sub-shell. Probably a cleaner option is to remove the `export` bit, and change `commandWithEnv` to `${envPrefix} ${command};`? This is a standard shell pattern to keep variables scoped to just that command.

packages/sandbox/src/sandbox.ts
```diff
@@ -447,7 +447,13 @@ export class Sandbox<Env = unknown> extends Container<Env> implements ISandbox {
         );
       } else {
         // Regular execution with session
-        const response = await this.client.commands.execute(command, sessionId);
+        const response = await this.client.commands.execute(
```
ghostwriternr, line 1075: Minor nitpick, okay to skip - since this is an internal method, we can probably make options an object instead of extending the number of params.

tests/e2e/test-worker/index.ts
```diff
@@ -381,7 +389,13 @@ console.log('Terminal server on port ' + port);
 
       // Process start
       if (url.pathname === '/api/process/start' && request.method === 'POST') {
-        const process = await executor.startProcess(body.command);
+        const process = await executor.startProcess(body.command, {
+          processId: body.processId,
+          env: body.env,
+          cwd: body.cwd,
+          timeout: body.timeout,
+          autoCleanup: body.autoCleanup
```
ghostwriternr: Are the rest of these fields actually being used properly? I would say we should not add them to the scope of this PR, but if not addressed, we should create a separate issue to track. #81 was created a while ago for the timeout.

tests/e2e/exec-env-vars-repro.test.ts
```diff
@@ -0,0 +1,195 @@
+import {
```
ghostwriternr: Is this test intended to be checked-in? Seems to be a mix of testing pre-change behaviour vs new

ghostwriternr (changes requested):
Besides these (`commandWithEnv` is the only blocking one though), I think we could do a couple of things more:
1. Add `setEnvVars` to the `ISandbox` type. I just realised this is not there currently, but should be.
2. In `createSession`, IF you think my earlier comment about `setEnvVars` behaviour on the global sandbox object vs the session object makes sense, we can merge `this.envVars` with the session-specific ones before creating a session from the session object.

packages/sandbox-container/src/session.ts
whoiskatrin: fixed

tests/e2e/exec-env-vars-repro.test.ts
whoiskatrin: removed

PR #213, opened 2025-11-13
Add normalizeId option to fix preview URLs

ghostwriternr:
## Summary
Adds opt-in `normalizeId` option to `getSandbox()` to fix preview URL routing issues with mixed/uppercase sandbox IDs.

## Problem
Sandbox IDs with uppercase letters cause preview URL requests to route to different Durable Object instances because hostnames are case-insensitive and thus the URL class automatically lowercases them (similar to DNS resolution). This results in INVALID_TOKEN errors.

## Solution
Add opt-in notmalizeId option that normalizes IDs to lowercase. This is done in order to prevent existing services from not being able to access sandbox instances started by an older version. We instead raise warnings in logs and errors right now so there is no breaking change for anyone, but giving developers sufficient time to transition to the new behaviour before we make this the default in a future version.

## Example Usage
```typescript
getSandbox(ns, "MyProject-123", { normalizeId: true })  // Creates DO with key "myproject-123"
```

## Breaking Change
This change in itself is non-breaking, but paves the way for a smooth transition to a breaking change in future. The goal is that we raise the logs & warning messages enough times so users opt-in to the new behaviour in the meantime.

### DIFF
--- a/.changeset/fix-turbo-prune-lockfile.md
+++ b/.changeset/fix-turbo-prune-lockfile.md
@@ -1,0 +1,7 @@
+---
+'@cloudflare/sandbox': patch
+---
+
+Fix Docker build failures caused by turbo prune lockfile mismatch
+
+Remove @cloudflare/vite-plugin from root devDependencies to avoid turbo prune bug with nested optionalDependencies. The vite-plugin is only used by examples which are excluded from Docker builds and already have it in their own package.json.
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,7 +18,6 @@
         "@biomejs/biome": "^2.3.2",
         "@changesets/changelog-github": "^0.5.1",
         "@changesets/cli": "^2.29.7",
-        "@cloudflare/vite-plugin": "^1.14.1",
         "@cloudflare/vitest-pool-workers": "^0.10.2",
         "@cloudflare/workers-types": "^4.20251014.0",
         "@types/bun": "^1.3.1",
@@ -201,7 +200,6 @@
       "integrity": "sha512-e7jT4DxYvIDLk1ZHmU/m/mB19rex9sv0c2ftBtjSBv+kVM/902eh0fINUzD7UwLLNR+jU585GxUJ8/EBfAM5fw==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@babel/code-frame": "^7.27.1",
         "@babel/generator": "^7.28.5",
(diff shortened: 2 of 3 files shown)


packages/sandbox/src/sandbox.ts
```diff
@@ -1031,20 +1047,29 @@ export class Sandbox<Env = unknown> extends Container<Env> implements ISandbox {
...
+      throw new SecurityError(
+        `Preview URLs require lowercase sandbox IDs. Your ID "${effectiveId}" contains uppercase letters.\n\n` +
+          `To fix this:\n` +
+          `1. Create a new sandbox with: getSandbox(ns, "${effectiveId}", { normalizeId: true })\n` +
+          `2. This will create a sandbox with ID: "${effectiveId.toLowerCase()}"\n\n` +
+          `Note: Due to DNS case-insensitivity, IDs with uppercase letters cannot be used with preview URLs.`
+      );
+    }
+
+    const sanitizedSandboxId = sanitizeSandboxId(sandboxId).toLowerCase();
```
ghostwriternr, line 1063: **Redundant toLowerCase after validation**

The validation at lines 1050-1061 already throws an error for uppercase IDs without normalizeId. So this line only executes for:
- Lowercase IDs (toLowerCase is no-op)
- Uppercase IDs with normalizeId=true (correct)

For code clarity, make it conditional:
```typescript
const sanitizedSandboxId = this.normalizeId 
  ? sanitizeSandboxId(sandboxId).toLowerCase()
  : sanitizeSandboxId(sandboxId);
```

This makes the intent explicit: normalize only when the flag is set.

packages/sandbox/src/sandbox.ts
```diff
@@ -133,11 +148,12 @@ export class Sandbox<Env = unknown> extends Container<Env> implements ISandbox {
     });
   }
 
-  // RPC method to set the sandbox name
-  async setSandboxName(name: string): Promise<void> {
+  async setSandboxName(name: string, normalizeId?: boolean): Promise<void> {
```
ghostwriternr, line 151: **Potential race condition in setSandboxName**

Two concurrent calls could both pass the `if (!this.sandboxName)` check before either writes to storage. Wrap the entire block in concurrency protection:

```typescript
async setSandboxName(name: string, normalizeId?: boolean): Promise<void> {
  if (!this.sandboxName) {
    await this.ctx.blockConcurrencyWhile(async () => {
      // Double-check inside critical section
      if (!this.sandboxName) {
        this.sandboxName = name;
        this.normalizeId = normalizeId || false;
        await this.ctx.storage.put('sandboxName', name);
        await this.ctx.storage.put('normalizeId', this.normalizeId);
      }
    });
  }
}
```

Impact is mitigated since getSandbox() always passes consistent values, but the code doesn't enforce this structurally.

whoiskatrin:
LGTM!

Issue #240, opened 2025-11-24
Cannot make preview url work with WS, neither in prod or local


riderx:
I'm trying to make a vite server run in the sandbox and preview it in an iframe.
`Failed to load resource: A server with the specified hostname could not be found.`
The only way to make the preview url work is to re proxy them myself from my worker.

Otherwise Each preview url i create is always having errors and i have no idea how to debug it, could you make an example with this ?
i'm not even at the step with the socket yet ^^

whoiskatrin:
@riderx can you check if you are accidentally using sandbox Id with spaces in it? try without

riderx:
i dont have space in names, i'm honeslty a bit lost, i try all and it fail. also the doc in the website is wrong i made PR to fix it https://github.com/cloudflare/sandbox-sdk/pull/241

ghostwriternr:
Hey @riderx! Thanks for the PR - that definitely seems like a miss at our end on the website and looks good to merge.
But in order to debug the underlying issue - can you share a very minimal repro for us to look at, or share some more details about your setup (specifically, what sandboxId you're using, how you are exposing ports, what does the preview url returned by exposePort look like, and when do you see the error you've shared). We're currently unable to reproduce the issue at our end but likely could be that our setup looks different from yours.

riderx:
@ghostwriternr @whoiskatrin here is simple reproduction repo with latest sandbox of what i try to achieve: https://github.com/riderx/reproduction-vite

I get the same issue:
Failed to load resource: Could not connect to the server.
While in the log i can see `VITE_READY` so the vite server is ready for sure.
The url returned by expose : http://3333-vite-echo-server-h5cyq-adv739zj65.localhost
Is not working
and if i go to http://localhost:55031 the vite server in sandbox is running, i do see the page working.
So i dont get why expose return me this URL who dont' work.
And in deployed env i do have the same issue as well

riderx:
with the reproduction it was way easier to debug myself, i found out the url returned by `exposePort`
cannot be used like that but if i proxy it with the worker i manage to see it like in this branch https://github.com/riderx/reproduction-vite/tree/debug
but the ws wont work, so i'm not sure to understand at all preview URL.

riderx:
I tried to deploy to a domain to see if that was just local but i confirm i cannot make the websocket work either: https://appmi.store
reload manually update but not HMR
from the branch https://github.com/riderx/reproduction-vite/tree/deploy-debug

[...]

PR #241, opened 2025-11-24
fix: doc websocket connection method

riderx:
this will help on #240 but still not solving it


ghostwriternr:
Thanks for the change! This was indeed missed at our end

PR #246, opened 2025-11-25
fix JS and TS executors to await Promise in context execution

whoiskatrin:
top level await, i played a lot with just wrapping all code and tbh, i wasn't happy with it, felt hacky

so now we parse code with acord, then we check if last AST node is an `ExpressionStatement`, if the answer is yes: wrap code with `return` on last expression, if no: wrap code without return (returns `undefined`)

### DIFF
--- a/tests/e2e/code-interpreter-workflow.test.ts
+++ b/tests/e2e/code-interpreter-workflow.test.ts
@@ -671,6 +671,214 @@ describe('Code Interpreter Workflow (E2E)', () => {
   }, 120000);
 
   // ============================================================================
+  // Top-Level Await (without IIFE wrappers)
+  // ============================================================================
+
+  test('should support top-level await without IIFE wrapper', async () => {
+    currentSandboxId = createSandboxId();
+    const headers = createTestHeaders(currentSandboxId);
+
+    // Create JavaScript context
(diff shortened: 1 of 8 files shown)


tests/e2e/code-interpreter-workflow.test.ts
```diff
@@ -367,6 +367,309 @@ describe('Code Interpreter Workflow (E2E)', () => {
     );
   }, 120000);
 
+  // ============================================================================
+  // Async/Await Promise Handling (Issue #206)
+  // ============================================================================
+
+  test('should resolve async IIFE and return the value', async () => {
```
ghostwriternr, line 374: Just to validate - all the tests are wrapped in an IIFE. Would this change work if the user passed code not wrapped in an IIFE? If not, is the answer then to wrap the underlying implementation in an async IIFE?

whoiskatrin: @ghostwriternr to support top-level await we'd need to wrap the code in an async IIFE, which adds complexity (need to handle "last expression" capture for multi-statement code) and changes. For this PR, I'd suggest keeping as is since it solves the issue. We can open a separate issue for top-level await? or should we just do it properly here?

whoiskatrin:
going to start fresh to do the top level await, not just this fix

PR #258, opened 2025-11-26
Fix executor mutex race condition and memory leak

ghostwriternr:
## Summary

Fixes race condition in `getExecutorLock()` where concurrent calls could create duplicate mutex instances, defeating serialization. Also fixes memory leak where executor mutexes weren't cleaned up on process exit.

## Test plan

- [x] E2E tests pass locally
- [x] Improved isolation test now verifies state across 12 contexts using `globalThis`


ghostwriternr:
@claude can you fix the issues raised in the code review comments

PR #261, opened 2025-11-28
top-level await 

whoiskatrin:
top level await, i played a lot with just wrapping all code and tbh, i wasn't happy with it, felt hacky

so now we parse code with acord, then we check if last AST node is an ExpressionStatement, if the answer is yes: wrap code with return on last expression, if no: wrap code without return (returns undefined)

### DIFF
--- a/packages/sandbox-container/src/runtime/executors/shared/code-transformer.ts
+++ b/packages/sandbox-container/src/runtime/executors/shared/code-transformer.ts
@@ -133,6 +133,10 @@ function processFunctionDeclaration(
   decl: FunctionDeclaration,
   source: string
 ): HoistedDeclaration {
+  // Anonymous functions (e.g., export default function() {}) have no id
+  if (!decl.id) {
+    return { names: [], assignment: '' };
+  }
   const name = decl.id.name;
   const funcText = source.slice(decl.start, decl.end);
 
@@ -151,6 +155,10 @@ function processClassDeclaration(
   decl: ClassDeclaration,
   source: string
 ): HoistedDeclaration {
+  // Anonymous classes (e.g., export default class {}) have no id
+  if (!decl.id) {
+    return { names: [], assignment: '' };
+  }
   const name = decl.id.name;
   const classText = source.slice(decl.start, decl.end);
 
@@ -287,8 +295,13 @@ export function transformForAsyncExecution(code: string): string {
     }
 
     // Add the async IIFE with transformed body
-    const bodyCode = bodyParts.join(';\n');
-    parts.push(`(async () => {\n${bodyCode};\n})()`);
+    if (bodyParts.length > 0) {
+      const bodyCode = bodyParts.join(';\n');
+      parts.push(`(async () => {\n${bodyCode};\n})()`);
+    } else {
+      // No body statements (e.g., just "let x;") - empty IIFE
+      parts.push('(async () => {})()');
+    }
 
     return parts.join('\n');
   } catch {
--- a/packages/sandbox-container/tests/runtime/code-transformer.test.ts
+++ b/packages/sandbox-container/tests/runtime/code-transformer.test.ts
@@ -56,8 +56,8 @@ describe('transformForAsyncExecution', () => {
 
     it('hoists declaration without initializer', () => {
       const result = transformForAsyncExecution('let x;');
-      // No assignment, just hoisting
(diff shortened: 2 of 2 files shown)


tests/e2e/code-interpreter-workflow.test.ts
```diff
@@ -668,13 +811,13 @@ console.log('Sum:', sum);
...
     // Execute different code in each context concurrently
-    // Each context sets its own unique value
+    // Each context sets its own unique value using globalThis for persistence
     const executionPromises = contexts.map((context, i) =>
       fetch(`${workerUrl}/api/code/execute`, {
         method: 'POST',
         headers,
         body: JSON.stringify({
-          code: `const value = ${i}; value;`,
+          code: `globalThis.value = ${i}; globalThis.value;`,
```
ghostwriternr: Okay this is probably a red-flag, and the one that threw me off the most. This E2E test modification should not be needed, and from what I understand, it's perhaps needed because the current implementation does not solve for variable scoping/persistence - which is a bug. Ideally, regular const/let/var declarations must persist across code executions.

packages/sandbox-container/src/runtime/executors/javascript/node_executor.ts
```diff
@@ -83,7 +85,14 @@ rl.on('line', async (line: string) => {
         options.timeout = timeout;
```
ghostwriternr, line 85: Is this timeout getting used for sync only or async as well?

packages/sandbox-container/src/runtime/executors/shared/code-transformer.ts
```diff
@@ -0,0 +1,43 @@
...
+      // Remove trailing semicolon if present
+      const cleanedExpr = lastExpr.replace(/;$/, '');
+      return `(async () => {\n${beforeLast}return ${cleanedExpr};\n})()`;
+    }
+
+    // For other statement types (declarations, loops, etc.), just wrap
+    return `(async () => {\n${trimmed}\n})()`;
+  } catch {
+    // If parsing fails, wrap anyway and let runtime report the error
+    return `(async () => {\n${trimmed}\n})()`;
```
ghostwriternr, line 313: Not sure I understand this one - where do such errors get handled?

packages/sandbox-container/tests/runtime/thenable.test.ts
```diff
@@ -0,0 +1,100 @@
...
+      const promiseAll = Promise.all([Promise.resolve(1)]);
+      expect(isThenable(promiseAll)).toBe(true);
+    });
+
+    it('Promise.race result', () => {
+      const promiseRace = Promise.race([Promise.resolve(1)]);
+      expect(isThenable(promiseRace)).toBe(true);
+    });
+  });
+});
```
ghostwriternr, line 18: Nit: This is a fairly verbose test with lots of repetitive checks, which we may not need.

```suggestion
describe('isThenable', () => {
  it('returns true for Promise', () => {
    expect(isThenable(Promise.resolve(42))).toBe(true);
  });

  it('returns true for custom thenable', () => {
    expect(isThenable({ then: () => {} })).toBe(true);
  });

  it('returns false for non-thenable', () => {
    expect(isThenable(null)).toBe(false);
    expect(isThenable(42)).toBe(false);
    expect(isThenable({ then: 'not a function' })).toBe(false);
  });
});
```

tests/e2e/code-interpreter-workflow.test.ts
ghostwriternr: I'd suggest having a few more scenarios here to ensure we're getting the implementation right:
1. Variables defined with await should persist across executions
2. We should be able to resolve Promise without the await keyword
3. The IIFE pattern from  the original issue

packages/sandbox-container/src/runtime/executors/javascript/node_executor.ts
whoiskatrin: good point, was only applying to sync, but fixed now for async as well

packages/sandbox-container/src/runtime/executors/shared/code-transformer.ts
whoiskatrin: node_executor.ts lines 109-112.

  } catch (error: unknown) {
    const err = error as Error;
    stderr += err.stack || err.toString();
    success = false;
  }

PR #263, opened 2025-11-30
Fix /api/ping endpoint to return health status

PavanKalisetti:
- This PR adds a case for the /api/ping route within the MiscHandler. It was previously unhandled, causing it to fall through to the default error case.
- Currently, any service trying to use /api/ping as a health or readiness check receives a 500 error. This causes deployment and monitoring workflows (like the initial sandbox readiness probe) to fail. This change ensures the endpoint behaves as expected by aliasing it to the existing /api/health functionality.

### DIFF
--- a/.changeset/fix-process-callbacks.md
+++ b/.changeset/fix-process-callbacks.md
@@ -1,0 +1,5 @@
+---
+'@cloudflare/sandbox': patch
+---
+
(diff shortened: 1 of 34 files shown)


ghostwriternr:
Thanks for the fix, @PavanKalisetti !

PR #264, opened 2025-12-01
Fix type error when extending Sandbox class

### LINKED ISSUE #262: Friction log
Tried doing some basic stuff with sandbox, faced some problems and couldn't proceed. Made a repro with the big ones, repo here: https://github.com/threepointone/repro-sandbox-friction

- using any class that extends Sandbox throws a type error in getSandbox(env.Sandbox, name)

Image

- ok then I cast it as what I think it should be, but it's not picking up the custom method I've exposed on the class. Not even sure this is a sandbox problem, but I feel stupid for not knowing how to fix this...

Image

- ok so I expose a methos that runs "echo 123" as a simple process, and add logs on onOutput, onError, onExit, onStart. Only onStart is triggered, but without a process id (because the process dies?) But I'd expect onOutput to show 123 

- or I'd expect await process.getLogs() to show me 123 but apparently that's empty too

Image

- I tried running opencode on this sanebox and sending it requests, but I don't think it's receiving requests. Or maybe it is! I can't say lol. 

(Btw, this is for me running opencode on a sandbox, and I have more questions around proxyToSandbox and passing environment variables, but I'm blocked with the above atm)

ghostwriternr:
Users extending `Sandbox` to add custom methods would get a type error when passing their namespace to `getSandbox()`. TypeScript's generic invariance meant the function couldn't accept subclass namespaces. `getSandbox` and `SandboxEnv` are now generic to properly support extended classes.

Fixes #262

### DIFF
--- a/.changeset/generic-sandbox-types.md
+++ b/.changeset/generic-sandbox-types.md
@@ -1,0 +1,5 @@
+---
+'@cloudflare/sandbox': patch
+---
+
+Fix type error when extending Sandbox class by making getSandbox and SandboxEnv generic


packages/sandbox/src/sandbox.ts
```diff
@@ -45,11 +45,11 @@ import {
 import type { MountInfo } from './storage-mount/types';
 import { SDK_VERSION } from './version';
 
-export function getSandbox(
-  ns: DurableObjectNamespace<Sandbox>,
+export function getSandbox<T extends Sandbox<any>>(
```
threepointone, line 48: should this have a default = Sandbox<unknown>

threepointone (commented):
generally should we use unknown instead of any?

packages/sandbox/src/sandbox.ts
ghostwriternr: This should already be what gets inferred by TS

PR #265, opened 2025-12-01
Use correct image tag in the example

jinjiaKarl:
Since https://github.com/cloudflare/sandbox-sdk/pull/259 splits docker images, we need to use the python image.

`{"output":"","error":"bash: line 15: python3: command not found","exitCode":127,"success":false}%`

### DIFF
--- a/examples/minimal/Dockerfile
+++ b/examples/minimal/Dockerfile
@@ -1,4 +1,4 @@
-FROM docker.io/cloudflare/sandbox:0.6.1-python
+FROM docker.io/cloudflare/sandbox:0.6.1
 
 # Required during local development to access exposed ports
 EXPOSE 8080


examples/code-interpreter/src/index.ts
```diff
@@ -43,7 +43,7 @@ async function callCloudflareAPI(
 
 async function executePythonCode(env: Env, code: string): Promise<string> {
   const sandboxId = env.Sandbox.idFromName('default');
-  const sandbox = getSandbox(env.Sandbox, sandboxId.toString());
+  const sandbox = getSandbox(env.Sandbox, sandboxId.toString().slice(0, 63));
```
jinjiaKarl, line 46: [ERROR] Sandbox execution failed: SecurityError: Sandbox ID must be 1-63 characters long.
 
 
 Reference: https://github.com/cloudflare/sandbox-sdk/blob/main/packages/sandbox/src/security.ts#L54

examples/minimal/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM docker.io/cloudflare/sandbox:0.6.0
+FROM docker.io/cloudflare/sandbox:0.6.0-python
```
ghostwriternr: Do we need to update this for the minimal example too? I'd prefer this to be the base image because the default code does not use it. But instead this is a good note to ensure the docs get updated to mention the tutorials like code interpreter mention that the base dockerfile should be updated.

examples/code-interpreter/src/index.ts
ghostwriternr, line 46: Is this needed here? We do indeed validate this here because some browsers stop being able to resolve overly long IDs, but in this particular example, this shouldn't be needed right?

examples/minimal/Dockerfile
jinjiaKarl: I agree that using the base image makes sense for a minimal example — it aligns well with the “minimal” concept.
However, we should update the command at that line https://github.com/cloudflare/sandbox-sdk/blob/main/examples/minimal/src/index.ts#L14 command here to something else, as the base image doesn't have python3.

examples/code-interpreter/src/index.ts
jinjiaKarl: I added it because I was seeing this error when running it locally.  The sandboxId I received was 64 characters long.  https://developers.cloudflare.com/durable-objects/api/id/#description
` [ERROR] Sandbox execution failed: SecurityError: Sandbox ID must be 1-63 characters long.`

kamath:
+1, faced this issue as well and was going to open my own PR haha

examples/code-interpreter/src/index.ts
ghostwriternr: Didn't realise this! All good then. Thankss

examples/minimal/Dockerfile
ghostwriternr: Great catch - I've updated the command

ghostwriternr:
Thanks for the fix, @jinjiaKarl ! 🧡

Issue #196, opened 2025-11-06
Listing files in a hidden directory requires `includeHidden`


NuroDev:
## The issue

Currently if you are trying to list files that are located inside of a hidden directory, you have to enable the `includeHidden` option, even if you are not looking for hidden files include the nested directory.

## Example

```ts
const sandbox = getSandbox(env.Sandbox, 'my-sandbox');

await sandbox.mkdir('/workspace/.hidden/foo/bar/', { recursive: true });

const works = await sandbox.listFiles('/workspace/.hidden/foo/', { includeHidden: true });
console.log("Works", works.files);

const doesntWork = await sandbox.listFiles('/workspace/.hidden/foo/');
console.log("Doesn't work", doesntWork.files);
```

whoiskatrin:
@NuroDev do you mind trying the package in this PR https://github.com/cloudflare/sandbox-sdk/pull/203 ? this should solve the issue

NuroDev:
Sure! Just tested a previous project that was showing this issue and yes #203 looks to fix this issue.

ghostwriternr:
Thanks for confirming, @NuroDev ! The fix should be available in the next release version.

PR #266, opened 2025-12-02
refactor e2e

deathbyknowledge:
Rewrote E2E tests to share a single sandbox container across all test files. Tests now run in parallel using sessions for isolation (~30s vs ~280s before).

Mainly did:
- Add global-setup.ts and helpers/global-sandbox.ts for shared container
- Add comprehensive-workflow.test.ts for consolidated happy path tests
- Convert all test files to shared sandbox pattern
- Update vitest config for parallel execution with global setup
- Add ports 9090-9092 to Dockerfiles for process-readiness tests
- Fix first-run failures by warming both base and Python containers

### DIFF
--- a/tests/e2e/environment-workflow.test.ts
+++ b/tests/e2e/environment-workflow.test.ts
@@ -3,18 +3,24 @@ import {
   getSharedSandbox,
   createUniqueSession
 } from './helpers/global-sandbox';
-import type { ExecResult } from '@repo/shared';
+import type { ExecResult, ExecEvent } from '@repo/shared';
+import { parseSSEStream } from '../../packages/sandbox/src/sse-parser';
 
 /**
- * Environment Edge Case Tests
+ * Environment Variable Tests
  *
- * Tests edge cases for environment and command execution.
- * Happy path tests (env vars, persistence, per-command env/cwd) are in comprehensive-workflow.test.ts.
+ * Tests all ways to set environment variables and their override behavior:
+ * - Dockerfile ENV (base level, e.g. SANDBOX_VERSION)
+ * - setEnvVars at session level
+ * - Per-command env in exec()
+ * - Per-command env in execStream()
  *
- * This file focuses on:
- * - Commands that read stdin (should not hang)
+ * Override precedence (highest to lowest):
+ * 1. Per-command env
+ * 2. Session-level setEnvVars
+ * 3. Dockerfile ENV
  */
-describe('Environment Edge Cases', () => {
+describe('Environment Variables', () => {
   let workerUrl: string;
   let headers: Record<string, string>;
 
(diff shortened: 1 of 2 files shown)


tests/e2e/git-clone-workflow.test.ts
```diff
@@ -1,318 +1,64 @@
...
-
-      // Git clone should fail with appropriate error
-      expect(cloneResponse.status).toBe(500);
-      const errorData = (await cloneResponse.json()) as ErrorResponse;
-      expect(errorData.error).toBeTruthy();
-      // Should mention repository not found or doesn't exist
-      expect(errorData.error).toMatch(
-        /not found|does not exist|repository|fatal/i
-      );
+describe('Git Clone Error Handling', () => {
```
ghostwriternr, line 19: Is it intentional that we've removed happy path tests here?

tests/e2e/keepalive-workflow.test.ts
```diff
@@ -1,277 +1,115 @@
...
  *
- * This validates that:
- * - The keepAlive interval properly renews activity timeout
- * - Containers don't auto-timeout when keepAlive is enabled
- * - Manual cleanup via destroy() works correctly
+ * What we verify:
+ * 1. keepAlive header is accepted and enables the mode
+ * 2. Multiple commands work with keepAlive enabled
+ * 3. File and process operations work with keepAlive
+ * 4. Explicit destroy works (cleanup endpoint)
```
ghostwriternr, line 18: I don't think these tests are validating the core feature: `keepAlive`  is primarily focused on the lifecycle of the container itself instead of what happens within the container. Ideally, this e2e test should be focused on ensuring that setting `keepAlive` keeps the container alive until it's explicitly destroyed.

ghostwriternr: Thinking a bit more about this... I also wonder how we'd do this, considering the default `sleepAfter` is 10 minutes. Probably the older tests didn't test this feature properly either. Does using `setSleepAfter` to a much shorter duration help?

tests/e2e/process-lifecycle-workflow.test.ts
```diff
@@ -1,826 +1,246 @@
...
-      workerUrl = result.url;
-      runner = result.runner;
-    });
-
-    afterEach(async () => {
-      // Cleanup sandbox container after each test
-      if (currentSandboxId) {
-        await cleanupSandbox(workerUrl, currentSandboxId);
-        currentSandboxId = null;
+describe('Process Lifecycle Error Handling', () => {
```
ghostwriternr, line 25: Same comment as before, why get rid of happy paths?

ghostwriternr: Also, imo, this test file is _the most important_ of all our e2e tests, because it goes into a lot of the session implementation behaviour. The git history might shed some light on how and where it's come handy.

tests/e2e/streaming-operations-workflow.test.ts
```diff
@@ -1,680 +1,139 @@
...
+ * Basic streaming tests are in comprehensive-workflow.test.ts.
  *
- * This validates the execStream() method which provides SSE-based
- * streaming for real-time command output.
+ * This file focuses on:
+ * - Command failures with non-zero exit codes
+ * - Nonexistent commands (exit code 127)
+ * - Long-running commands (15s+, 60s+)
+ * - High-volume streaming
+ * - Intermittent output gaps
```
ghostwriternr: I like these focus areas, but looks like we've removed most of these

ghostwriternr (changes requested):

tests/e2e/environment-workflow.test.ts
```diff
@@ -1,374 +1,69 @@
...
-      // Get test worker URL (CI: uses deployed URL, Local: spawns wrangler dev)
-      const result = await getTestWorkerUrl();
-      workerUrl = result.url;
-      runner = result.runner;
+  getSharedSandbox,
+  createUniqueSession
+} from './helpers/global-sandbox';
+import type { ExecResult } from '@repo/shared';
+
+/**
```
ghostwriternr, line 9: can/should we expand this test to cover all the ways in which it's possible to set envs and whether they're getting overridden properly?
1. Via ENV in dockerfile
2. Via setEnvVars at the sandbox level
3. Via setEnvVars at the session level
4. Passing `env` to exec/execStream/startProcess

tests/e2e/file-operations-workflow.test.ts
```diff
@@ -366,833 +68,71 @@ describe('File Operations Workflow (E2E)', () => {
...
-      method: 'POST',
-      headers,
-      body: JSON.stringify({
-        path: '/workspace/large.txt',
-        content: largeContent
-      })
-    });
-
-    // Stream the file
-    const streamResponse = await fetch(`${workerUrl}/api/read/stream`, {
```
ghostwriternr, line 771: This functionality doesn't seem to be tested anywhere post-refactor

tests/e2e/git-clone-workflow.test.ts
deathbyknowledge: Yes, covered in `comprehensive-workflow.test.ts`

tests/e2e/process-lifecycle-workflow.test.ts
deathbyknowledge: Yes, covered in `comprehensive-workflow.test.ts`

tests/e2e/environment-workflow.test.ts
deathbyknowledge: IMO not in the e2e tests. We can ensure the logic works properly on "cheaper" tests

tests/e2e/file-operations-workflow.test.ts
```diff
@@ -366,833 +68,71 @@ describe('File Operations Workflow (E2E)', () => {
...
-      })
-    });
-
-    expect(notExistsResponse.status).toBe(200);
-    const notExistsData = (await notExistsResponse.json()) as FileExistsResult;
-    expect(notExistsData.success).toBe(true);
-    expect(notExistsData.exists).toBe(false);
-  }, 90000);
-
-  test('should list files in hidden directories without includeHidden flag', async () => {
```
ghostwriternr, line 1096: I remember this test being introduced as part of a fix to #196. pls ensure we still have this.

tests/e2e/keepalive-workflow.test.ts
deathbyknowledge: Oof... I'll try updating `sleepAfter`. Is there a way we can confirm the container went to sleep? Seems like a hard thing to test for from here

tests/e2e/file-operations-workflow.test.ts
```diff
@@ -366,833 +68,71 @@ describe('File Operations Workflow (E2E)', () => {
...
-
-    expect(readData.success).toBe(true);
-    expect(readData.content).toBe('Hello, World! This is a test.');
-    expect(readData.encoding).toBe('utf-8');
-    expect(readData.isBinary).toBe(false);
-    expect(readData.mimeType).toMatch(/text\/plain/);
-    expect(readData.size).toBeGreaterThan(0);
-  }, 90000);
-
-  test('should write and read binary files with base64 encoding', async () => {
```
ghostwriternr, line 679: We should have a bunch of tests to support our `encoding` field. see the both the read & write methods and how encoding is used and where it needs to be tested.

tests/e2e/file-operations-workflow.test.ts
deathbyknowledge: True, I covered exec streaming and log streaming but missed the file streaming. Adding one back

tests/e2e/file-operations-workflow.test.ts
deathbyknowledge: Added this one back for reading. Since writing is simpler and already covered in the unit tests I don't think we need to add, but let me know if you think otherwise

tests/e2e/file-operations-workflow.test.ts
deathbyknowledge: Got it, added it back

tests/e2e/file-operations-workflow.test.ts
ghostwriternr: yeah read is where I expect this to be used more anyway. All good.

tests/e2e/environment-workflow.test.ts
ghostwriternr: Hmm, but where else though? I don't think unit tests can capture this due to the granularity? Esp. considering the 4-level hierarchy (tbh, i can't remember if levels 2 & 3 override each other - so we can check that).

PR #272, opened 2025-12-03
fix for trusted publishing

whoiskatrin:


.github/workflows/release.yml
```diff
@@ -184,7 +193,9 @@ jobs:
           target: default
           platforms: linux/amd64
           push: true
-          tags: cloudflare/sandbox:${{ needs.unit-tests.outputs.version }}
+          tags: |
+            cloudflare/sandbox:${{ needs.unit-tests.outputs.version }}
+            cloudflare/sandbox:latest
```
ghostwriternr: I don't know if our Dockerhub account is configured to allow mutating tags (i.e. on every new version release, the new one should get tagged as 'latest' and the old one automatically get untagged from being the 'latest'). We should test this, or keep this part of the change as a separate PR.

.github/workflows/release.yml
```diff
@@ -211,5 +224,58 @@ jobs:
...
+        uses: docker/build-push-action@v6
+        with:
+          context: .
+          file: packages/sandbox/Dockerfile
+          target: default
+          platforms: linux/amd64
+          push: true
+          tags: |
+            cloudflare/sandbox:${{ steps.beta-version.outputs.version }}
+            cloudflare/sandbox:beta
```
ghostwriternr: Same question as earlier

Issue #270, opened 2025-12-02
Add an option/callback/other to allow users to wait for some event/log/custom check after startProcess before they continue


ghostwriternr:
Right now, this looks a problem users are solving this repeatedly at their end

ghostwriternr:
Fixed by #273

PR #273, opened 2025-12-03
process readiness 

whoiskatrin:
fixes this https://github.com/cloudflare/sandbox-sdk/issues/270

### DIFF
--- a/.changeset/mighty-squids-count.md
+++ b/.changeset/mighty-squids-count.md
@@ -1,15 +1,15 @@
 ---
-"@cloudflare/sandbox": patch
+'@cloudflare/sandbox': patch
 ---
+
 Add process readiness detection with port and log pattern waiting
 The `Process` object returned by `startProcess()` now includes readiness methods:
 
-  - `process.waitForPort(port, options?)`: Wait for process to listen on a port
-    - Supports HTTP mode (default): checks endpoint returns expected status (200-399)
-    - Supports TCP mode: checks port accepts connections
-    - Container-side checking via `/api/port-check` endpoint
-    - Configurable timeout, interval, path, and expected status
+- `process.waitForPort(port, options?)`: Wait for process to listen on a port
+  - Supports HTTP mode (default): checks endpoint returns expected status (200-399)
+  - Supports TCP mode: checks port accepts connections
+  - Configurable timeout, interval, path, and expected status
 
-  - `process.waitForLog(pattern, options?)`: Wait for pattern in process output
-    - Supports string or RegExp patterns
-    - Returns matching line and capture groups
+- `process.waitForLog(pattern, options?)`: Wait for pattern in process output
+  - Supports string or RegExp patterns
+  - Returns matching line and capture groups
--- a/package-lock.json
+++ b/package-lock.json
@@ -401,7 +401,6 @@
       "resolved": "https://registry.npmjs.org/picomatch/-/picomatch-4.0.3.tgz",
       "integrity": "sha512-5gTmgEY/sqK6gFXLIsQNH19lWb4ebPDLA4SdLP7dsWkIXHWlG66oPuVvXSGFPppYZz8ZDZq0dYYrbHfBCVUb1Q==",
       "license": "MIT",
-      "peer": true,
       "engines": {
         "node": ">=12"
       },
@@ -423,7 +422,6 @@
       "resolved": "https://registry.npmjs.org/vite/-/vite-6.4.1.tgz",
       "integrity": "sha512-+Oxm7q9hDoLMyJOYfUYBuHQo+dkAloi33apOPP56pzj+vsdJDzr+j1NISE5pyaAuKL4A3UD34qd0lx5+kfKp2g==",
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "esbuild": "^0.25.0",
         "fdir": "^6.4.4",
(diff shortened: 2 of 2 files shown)


packages/sandbox/README.md
```diff
@@ -92,11 +92,61 @@ export default {
       return Response.json({ content: file.content });
     }
 
-    return new Response('Try /run or /file');
+    // Start a server and wait for it to be ready
```
ghostwriternr: Let's keep the README simple with only the basic example. All advanced patterns should go into the docs.

packages/sandbox/src/sandbox.ts
```diff
@@ -1239,10 +1248,375 @@ export class Sandbox<Env = unknown> extends Container<Env> implements ISandbox {
...
 
+  /**
+   * Wait for a condition to be met for a process
+   * Supports log patterns (string/regex) and port availability
+   */
+  private async waitForCondition(
+    processId: string,
+    command: string,
+    condition: ReadyCondition,
+    timeout: number = 30_000
```
ghostwriternr: As a general pattern, I've been moving away from having any built-in default timeouts in our SDK. We used to have a bunch of timeouts with the understanding that it's a good safeguard, but more often than not, users were hitting these timeouts and wondering how to get over it than thinking - oh it's nice to have this timeout. I like the idea of the user being able to pass an optional timeout here, but there should be no defaults. If the user by default provides no default, we will just wait indefinitely.

packages/sandbox/src/sandbox.ts
```diff
@@ -1239,10 +1248,375 @@ export class Sandbox<Env = unknown> extends Container<Env> implements ISandbox {
...
+                return result;
+              }
+            } else {
+              collectedStderr += data;
+              // Check accumulated buffer for pattern (handles patterns split across chunks)
+              const result = this.matchPattern(collectedStderr, pattern);
+              if (result) {
+                return result;
+              }
+            }
```
ghostwriternr, line 1383: minor: Should we debounce this check, since the stream could be coming in rather fast and we're doing a string/regex match on every chunk?

packages/shared/src/errors/contexts.ts
```diff
@@ -48,6 +48,27 @@ export interface ProcessErrorContext {
...
+}
+
+export interface ProcessExitedBeforeReadyContext {
+  processId: string;
+  command: string;
+  condition: string;
+  exitCode: number;
+  stdout?: string;
+  stderr?: string;
+}
```
ghostwriternr, line 66: `stdout` and `stderr` could be huge - should we really be returning them as part of the error object? I like all the other fields.

packages/sandbox/src/index.ts
```diff
@@ -96,6 +98,15 @@ export type {
   ExecutionCallbacks,
   InterpreterClient
 } from './clients/interpreter-client.js';
+export type {
+  ProcessExitedBeforeReadyContext,
+  ProcessReadyTimeoutContext
+} from './errors';
```
ghostwriternr: Do end users need these?

packages/sandbox/src/sandbox.ts
```diff
@@ -1239,10 +1246,370 @@ export class Sandbox<Env = unknown> extends Container<Env> implements ISandbox {
...
+      try {
+        const execTimeoutMs =
+          timeout !== undefined
+            ? Math.min(execTimeout, timeout - (Date.now() - startTime))
+            : execTimeout;
+
+        const result = await this.exec(
+          `bash -c 'echo > /dev/tcp/localhost/${port}' 2>/dev/null`,
+          { timeout: execTimeoutMs }
+        );
```
ghostwriternr: okay these 3 lines are a bit deceptive but perhaps force us to think about this design some more. So when developers say `waitForPort`, they're basically saying - _"Block until this port is serving requests properly"_.

That could mean different things, based on what developers are starting. For example, properly could mean a health endpoint returning 200, or a database is now accepting connections, or some other custom logic too. But the current check essentially just validates if something is bound and listening to the port or if a TCP handshake works, but all of this might happen **much before** the port is actually ready to be used.

And so we need to improve this check. But also, improve our interface to our now separate `waitForPort` method. What if, we define an interface like so:
```ts
interface WaitForPortOptions {
  // Check mode (default: 'http')
  mode?: 'http' | 'tcp';

  // HTTP mode options
  path?: string;              // default '/' -> this is useful because even for the sandbox sdk, the health check port does not determine if the interpreter is up, due to delayed startup
  status?: number; // default 200-399 -> in case the health check endpoint actually returns 201 or something

  // Timing
  timeout?: number;           // default no timeout
  interval?: number;          // default 500
}
```
For TCP, we could just do `Bun.connect` and for HTTP we can do a `fetch`

AND expose a third method `waitForCommand` that covers all advanced use-cases:
```ts
interface WaitForCommandOptions {
  // this command is 'successful' if it exits with code 0
  command: string;

  // Timing
  timeout?: number;           // default no timeout
  interval?: number;          // default 500
}
```
useful to do things like `redis-cli ping | grep PONG`. And we can even run this in the same session.

And most importantly - all of these ideally are implemented within sandbox-container, i.e. on the container side and exposed as new simple API(s) that the DO side polls.

ghostwriternr: For this PR though - we can just do `waitForPort` and keep `waitForCommand` as a future enhancement, to prevent scope creep. And might keep the PR clean too, because `waitForCommand` will use the session (although I don't think it'll be messy at all).

ghostwriternr:
and changeset too pls

.changeset/mighty-squids-count.md
```diff
@@ -0,0 +1,15 @@
+---
+"@cloudflare/sandbox": patch
+---
+Add process readiness detection with port and log pattern waiting
+The `Process` object returned by `startProcess()` now includes readiness methods:
+
+  - `process.waitForPort(port, options?)`: Wait for process to listen on a port
+    - Supports HTTP mode (default): checks endpoint returns expected status (200-399)
+    - Supports TCP mode: checks port accepts connections
+    - Container-side checking via `/api/port-check` endpoint
```
ghostwriternr: nit: This is an internal implementation, we can remove from changeset to avoid confusions for users.

PR #289, opened 2025-12-10
fix workspace bug

whoiskatrin:


ghostwriternr (commented):
Good core fix, but the tests are overly verbose and redundant. We should be able to have the same level of guards with much more minimal tests.

packages/sandbox-container/tests/session.test.ts
```diff
@@ -496,6 +543,92 @@ describe('Session', () => {
...
+        `rm -rf ${workspaceDir} && ln -sf ${backupDir} ${workspaceDir}`
+      );
+
+      // Try a subsequent command - should continue working
+      const afterSymlink = await session.exec('echo "after symlink"');
+      expect(afterSymlink.exitCode).toBe(0);
+      expect(afterSymlink.stdout.trim()).toBe('after symlink');
+    });
+
+    it('should be recoverable by cd-ing to valid directory after cwd deletion', async () => {
```
ghostwriternr: This test can be removed imo - seems to be testing a no-op

packages/sandbox-container/tests/session.test.ts
```diff
@@ -496,6 +543,92 @@ describe('Session', () => {
       expect(result.exitCode).toBe(1);
       expect(result.stderr).toContain('Failed to change directory');
     });
+
+    it('should continue working after session cwd is deleted (issue #288)', async () => {
```
ghostwriternr: nitt: Lets not mention github issue numbers in tests

tests/e2e/workspace-deletion-workflow.test.ts
```diff
@@ -0,0 +1,105 @@
+/**
+ * Workspace Deletion Workflow Tests (Issue #288)
+ *
+ * Tests that new sessions fall back to the home directory when /workspace
+ * doesn't exist, preventing "Unknown Error, TODO" failures.
+ *
+ * @see https://github.com/cloudflare/sandbox-sdk/issues/288
+ */
```
ghostwriternr: okay you did suggest this already, but on second thoughts, I think we can indeed remove these. the unit tests already cover these exact scenarios, and its more of an implementation detail that we don't need to worry about at the e2e level.

whoiskatrin: done done

PR #299, opened 2025-12-12
Add waitForExit() method to Process interface

### LINKED ISSUE #297: Add waitForExit() method to Process objects
## Problem

When using `startProcess()` for tasks that exit on their own, there's no clean way to wait for completion. Users must either:

1. Use `waitForLog()` with a known output pattern
2. Manually wrap `onExit` callback in a Promise

```typescript
// Current workaround
let resolveExit: (code: number | null) => void;
const exitPromise = new Promise<number | null>((r) => resolveExit = r);
const process = await sandbox.startProcess('npm run build', { onExit: resolveExit });
await exitPromise;
```

## Proposal

Add `waitForExit(timeout?)` to Process objects, complementing existing `waitForLog()` and `waitForPort()`:

```typescript
const process = await sandbox.startProcess('npm run build');
const exitCode = await process.waitForExit(60000);
```

## Use Cases

- Run multiple processes concurrently, wait for specific ones to finish
- Start a process, do other work, then wait for completion
- Start a process that might run indefinitely, but check if it crashed early

whoiskatrin:
Implemented waitForExit(timeout?) method to Process objects

also replaced the arbitrary setTimeout(resolve, 3000) workaround with waitForExit():
- Now uses the waitForExit() API to wait for the background process to complete
- Validates exit code and verifies stdout output from the WaitForExitResult
- More reliable and deterministic than arbitrary delays

fixes #297

### DIFF
--- a/packages/sandbox/src/sandbox.ts
+++ b/packages/sandbox/src/sandbox.ts
@@ -1561,7 +1561,7 @@ export class Sandbox<Env = unknown> extends Container<Env> implements ISandbox {
         for await (const event of parseSSEStream<LogEvent>(stream)) {
           if (event.type === 'exit') {
             return {
-              exitCode: event.exitCode ?? 0
+              exitCode: event.exitCode ?? 1
             };
           }
         }


packages/shared/src/types.ts
```diff
@@ -110,6 +110,18 @@ export interface WaitForLogResult {
...
+/**
+ * Result from waiting for process exit
+ */
+export interface WaitForExitResult {
+  /** Process exit code */
+  exitCode: number;
+  /** Accumulated stdout output */
+  stdout: string;
+  /** Accumulated stderr output */
+  stderr: string;
```
ghostwriternr: Similar to the other waitFor methods, I think we can leave out accumulated logs from being returned in the response. If users need it, they anyway have access to the various log fetching utils right now. This method has a simple purpose that already gets solved without these.

packages/sandbox/src/sandbox.ts
```diff
@@ -1521,6 +1526,135 @@ export class Sandbox<Env = unknown> extends Container<Env> implements ISandbox {
...
+  ): Promise<WaitForExitResult> {
+    const startTime = Date.now();
+    let collectedStdout = '';
+    let collectedStderr = '';
+
+    // First check if process already exited by getting existing logs and status
+    try {
+      const existingLogs = await this.getProcessLogs(processId);
+      collectedStdout = existingLogs.stdout;
+      collectedStderr = existingLogs.stderr;
```
ghostwriternr: If we remove the logs from output, we can perhaps completely get rid of all log-related logic from here.

packages/sandbox/src/sandbox.ts
```diff
@@ -1521,6 +1526,135 @@ export class Sandbox<Env = unknown> extends Container<Env> implements ISandbox {
...
+    } catch (error) {
+      // Process might not exist yet or other error, continue to streaming
+      this.logger.debug('Could not get existing process info, will stream', {
+        processId,
+        error: error instanceof Error ? error.message : String(error)
+      });
+    }
+
+    // Stream logs and wait for exit event
+    const stream = await this.streamProcessLogs(processId);
```
ghostwriternr: Same as above

packages/sandbox/tests/process-readiness.test.ts
```diff
@@ -660,6 +660,309 @@ data: {"type":"exit","exitCode":127,"timestamp":"${new Date().toISOString()}"}
     });
   });
 
+  describe('waitForExit() method', () => {
```
ghostwriternr, line 663: Tests can be more concise. Lots of overlap with current setup.

tests/e2e/process-readiness-workflow.test.ts
```diff
@@ -321,6 +326,148 @@ await Bun.sleep(60000);
     );
   }, 60000);
 
+  test('should wait for process to exit and return exit code with output', async () => {
```
ghostwriternr: no-op: v nice e2e tests

ghostwriternr:
Minor import order merge conflicts too, sorry about that

.changeset/clean-pans-switch.md
```diff
@@ -1,5 +1,5 @@
 ---
-"@cloudflare/sandbox": patch
+'@cloudflare/sandbox': patch
```
ghostwriternr: This is modifying an existing changeset. We need a new one for this change.

packages/sandbox/src/sandbox.ts
```diff
@@ -1521,6 +1526,53 @@ export class Sandbox<Env = unknown> extends Container<Env> implements ISandbox {
...
+      if (isTerminalStatus(processInfo.status)) {
+        return {
+          exitCode:
+            processInfo.exitCode ?? (processInfo.status === 'completed' ? 0 : 1)
+        };
+      }
+
+      // Wait before next poll
+      await new Promise((resolve) => setTimeout(resolve, pollInterval));
+    }
```
ghostwriternr: Thanks for cleaning out the logs parts! But if you won't hate me, I have a proposal haha.

So, right now, this method is purely polling based which is a LOT of requests to our container - especially if the process is a long running one.  How about, we use `streamProcessLogs` here, but don't use it for the actual logs but instead we just use it to listen for the exit status. There's 2 things I love about this:
1. This way, there's only a single request to the container, and our durable object infrastructure is far more robust to incoming SSE events from the container.
2. The `streamProcessLogs` also takes care of the situations wherein the process exited before the method is called and it'll send you the exit event right away, so don't need special handling for that case either.

What do you think of this? Rough change (might need small tweaks):
```suggestion
    const stream = await this.streamProcessLogs(processId);

    // Set up timeout if specified
    let timeoutId: ReturnType<typeof setTimeout> | undefined;
    let timeoutPromise: Promise<never> | undefined;

    if (timeout !== undefined) {
      timeoutPromise = new Promise<never>((_, reject) => {
        timeoutId = setTimeout(() => {
          reject(
            this.createReadyTimeoutError(processId, command, 'process exit', timeout)
          );
        }, timeout);
      });
    }

    try {
      const streamProcessor = async (): Promise<WaitForExitResult> => {
        for await (const event of parseSSEStream<LogEvent>(stream)) {
          if (event.type === 'exit') {
            return {
              exitCode: event.exitCode ?? (event.data?.includes('completed') ? 0 : 1)
            };
          }
        }

        // Stream ended without exit event - shouldn't happen, but handle gracefully
        throw new Error(
          `Process ${processId} stream ended unexpectedly without exit event`
        );
      };

      if (timeoutPromise) {
        return await Promise.race([streamProcessor(), timeoutPromise]);
      }
      return await streamProcessor();
    } finally {
      if (timeoutId) {
        clearTimeout(timeoutId);
      }
    }
```

packages/sandbox/src/sandbox.ts
```diff
@@ -1521,6 +1526,63 @@ export class Sandbox<Env = unknown> extends Container<Env> implements ISandbox {
...
+        }, timeout);
+      });
+    }
+
+    try {
+      const streamProcessor = async (): Promise<WaitForExitResult> => {
+        for await (const event of parseSSEStream<LogEvent>(stream)) {
+          if (event.type === 'exit') {
+            return {
+              exitCode: event.exitCode ?? 0
```
ghostwriternr: nit: should the fallback be 0 or 1? Assuming failure when we don't have a clean exit code might be better no?

PR #310, opened 2025-12-19
Add PTY terminal passthrough for browser clients

whoiskatrin:
## Summary

Enables browser-based terminal UIs to connect to sandbox shells via WebSocket. The `terminal()` method proxies WebSocket connections to the container's PTY, with output buffering for replay on reconnect.

## What's included

- **`sandbox.terminal(request)`** — proxy a WebSocket upgrade to a container PTY.
- **Multiple terminals per sandbox** — each session can have its own terminal with isolated working directory and environment, so users can run separate shells side-by-side in the same container.
- **`@cloudflare/sandbox/xterm`** — `SandboxAddon` for xterm.js with automatic reconnection (exponential backoff + jitter), buffered output replay, and resize forwarding.

## Usage

```typescript
// Worker: proxy WebSocket to container terminal
return sandbox.terminal(request, { cols: 80, rows: 24 });

// Multiple isolated terminals in the same sandbox
const dev = await sandbox.getSession('dev');
return dev.terminal(request);
```

```typescript
// Browser: xterm.js addon
import { SandboxAddon } from '@cloudflare/sandbox/xterm';

const addon = new SandboxAddon({
  getWebSocketUrl: ({ sandboxId, origin }) =>
    `${origin}/ws/terminal?id=${sandboxId}`,
  onStateChange: (state, error) => updateUI(state),
});
terminal.loadAddon(addon);
addon.connect({ sandboxId: 'my-sandbox' });
```


whoiskatrin:
this has been also tested by @sdan in his work, he has positive feedback

packages/sandbox/src/clients/transport/types.ts
```diff
@@ -74,4 +74,24 @@ export interface ITransport {
...
+
+  /**
+   * Register PTY data listener (WebSocket only)
+   */
+  onPtyData(ptyId: string, callback: (data: string) => void): () => void;
+
+  /**
+   * Register PTY exit listener (WebSocket only)
+   */
+  onPtyExit(ptyId: string, callback: (exitCode: number) => void): () => void;
```
ghostwriternr: These methods probably don't belong here. The more I thought about it, it feels like a very natural fit for the PTY support is to use websockets because of the continuous bidirectional communication on every input (keystroke) and continuous output - and this is something you've already done (looking mainly at the changes here and to `ws-transport.ts` (and the unimplemented stubs on the http transport).

But the above is also exactly why the way to achieve this is to not add PTY-specific components to the transport layer itself, which serves a more foundational purpose that both the HTTP & WS transports implement. But instead, the PTY feature should unconditionally use the websocket transport and use the standard interface provided by the transport instead.

I'd imagine we could instead perhaps have (these are just illustrative, could think about whether the existing methods are useful too):
```ts
abstract sendMessage(message: object): void;
abstract onStreamEvent(
  streamId: string,
  event: string,
  callback: (data: string) => void
): () => void;
```
and have both transports implement this. Remove all pty specific code like listeners from the ws-transport implementation and use a more standardised streamEventListeners perhaps, and pty-client could possibly then do things like:
```ts
transport.sendMessage({ type: 'pty_input', ptyId, data });
transport.onStreamEvent(ptyId, 'pty_data', callback);
```

Right now, this sort of very tight unidirectional coupling is defeating the purpose of having a common transport.

ghostwriternr: Other files in this PR that are related to the same comment: `ws-transport.ts`, `http-transport.ts`, `base-transport.ts`, `pty-client.ts`

packages/sandbox/src/clients/pty-client.ts
```diff
@@ -0,0 +1,479 @@
...
+   * of an existing session.
+   *
+   * @param sessionId - Session ID to attach to
+   * @param options - PTY options (terminal size)
+   * @returns PTY handle for interacting with the terminal
+   *
+   * @example
+   * const pty = await client.attach('session_123', { cols: 100, rows: 30 });
+   */
+  async attach(sessionId: string, options?: AttachPtyOptions): Promise<Pty> {
```
ghostwriternr: We seem to have a lot of code around the implementation for this attach feature, and while I initially got excited by the idea of it, the more I looked at this, the less clear I was on the exact purpose of this one. When I read "attach", I assumed it implies an ongoing relationship between PTY and `session`, but in reality the PTY gets its own bash process that diverges immediately after creation. And this part feels fair - because our session implementation uses FIFO-based I/O and primarily designed for a non-interactive use-case, wherein PTY uses TTY-based I/O and is for interactive use-cases. But what this means, is that attach is primarily just a way to avoid the user having to write
```ts
const pty = await sandbox.pty.create({
  cwd: '/workspace/project',
  env: { NODE_ENV: 'development' }
});
```  
which is something we entirely support already. And furthermore, when you look at the fact that we currently use `getInitialCwd` and `getInitialEnv`, it's also not the latest state of the session but is only the initial state - making this entire attach feature even less enticing/useful.

And the existence of this feature has also lead to implementation of exclusive control, which doesn't make sense because the PTY and the session are divergent anyway so there is no reason to block one or the other.

So my personal recommendation is to remove the attach feature entirely unless there are strong reasons to keep it. We could get rid of the attach() method, handleAttach(), getSessionInfo, sessionToPty mapping, hasActivePty(), getBySessionId(), checkPtyExclusiveControl(), the attach route, AttachPtyOptions type, and PTY_EXCLUSIVE_CONTROL error code. And any other related code too.

packages/sandbox-container/src/managers/pty-manager.ts
```diff
@@ -0,0 +1,449 @@
...
+      this.logger.warn('Write to unknown PTY', { ptyId: id });
+      return { success: false, error: 'PTY not found' };
+    }
+    if (session.state !== 'running') {
+      this.logger.warn('Write to exited PTY', { ptyId: id });
+      return { success: false, error: 'PTY has exited' };
+    }
+    try {
+      // Handle Ctrl+C (ETX, 0x03) - send SIGINT to process group
+      if (data === '\x03') {
```
ghostwriternr: This comment is maybe relevant to the other signals too. When I tested using the example app you have deployed to your account, it looks like Ctrl+C/Z send signals to the shell process only, but not to the foreground process group. So for instance, if you run `sleep 100` and press Ctrl+C/Z, nothing happens because the SIGINT gets sent to bash but is ignored and thus the sleep never gets SIGINT.

Can you remind me why we need this special handling here? Naively, it feels like this method could just be:
```ts
write(id: string, data: string): { success: boolean; error?: string } {
  const session = this.sessions.get(id);
  if (!session) return { success: false, error: 'PTY not found' };
  if (session.state !== 'running') return { success: false, error: 'PTY has exited' };

  session.terminal.write(data);
  return { success: true };
}
```
But there's certainly a reason why this was added that I'm missing in my testing.

packages/sandbox-container/src/managers/pty-manager.ts
```diff
@@ -0,0 +1,449 @@
...
+        );
+      }
+      return () => {};
+    }
+
+    session.exitListeners.add(callback);
+    return () => session.exitListeners.delete(callback);
+  }
+
+  startDisconnectTimer(id: string): void {
```
ghostwriternr: Looks like these methods are never used (start and cancel), but are only being used in tests. We could remove them? And by extension, I noticed the `disconnectTimer`/`disconnectTimeout` option defined elsewhere seems to be unused as well. And fwiw - following the general pattern in this SDK - I lean towards not having any automatic timeouts at all in the SDK. We _could_ provide an optional option if users ask, but for simplicity, I'd just remove it because it's guaranteed to surprise users if we implement this properly and PTY just disconnects after 30 seconds of inactivity by default and users have to read source code/docs to understand how to not have that happen. Instead, we already expose `kill` to users so if they need timeouts, they can achieve it using very little code anyway and they have much more control over the process.

packages/sandbox-container/src/handlers/ws-adapter.ts
```diff
@@ -345,22 +424,92 @@ export class WebSocketAdapter {
...
+
+  /**
+   * Register PTY output listener for a WebSocket connection
+   * Returns cleanup function to unsubscribe from PTY events
+   *
+   * Auto-unsubscribes when send fails to prevent resource leaks
+   * from repeatedly attempting to send to a dead connection.
+   * Also tracked per-connection for cleanup when connection closes.
+   */
+  registerPtyListener(ws: ServerWebSocket<WSData>, ptyId: string): () => void {
```
ghostwriternr: This method seems unused as well and only used by tests - and is not exposed to developers either. We can delete this and all related code if not needed?

packages/sandbox-container/src/routes/setup.ts
```diff
@@ -196,6 +196,63 @@ export function setupRoutes(router: Router, container: Container): void {
...
+    handler: async (req, ctx) => container.get('ptyHandler').handle(req, ctx),
+    middleware: [container.get('loggingMiddleware')]
+  });
+
+  router.register({
+    method: 'POST',
+    path: '/api/pty/{id}/resize',
+    handler: async (req, ctx) => container.get('ptyHandler').handle(req, ctx),
+    middleware: [container.get('loggingMiddleware')]
+  });
```
ghostwriternr: Possibly related to my comment about the transport layer cleanup - right now, I think these endpoints exist as special methods only because of how the transport layer got implemented. But feels like we could get rid of these and still have everything working?

packages/sandbox/src/clients/pty-client.ts
```diff
@@ -0,0 +1,479 @@
...
+            rows
+          }
+        );
+        throw new Error(`PTY resize failed: ${message}`);
+      }
+      return;
+    }
+
+    // HTTP: await the response to surface errors
+    const response = await this.transport.fetch(`/api/pty/${this.id}/resize`, {
```
ghostwriternr: +related to ws transport cleanup. We could get rid of this fallback logic I think.

packages/sandbox/src/clients/pty-client.ts
```diff
@@ -0,0 +1,479 @@
...
+            ptyId: this.id
+          }
+        );
+        throw new Error(`PTY write failed: ${message}`);
+      }
+      return;
+    }
+
+    // HTTP: await the response to surface errors
+    const response = await this.transport.fetch(`/api/pty/${this.id}/input`, {
```
ghostwriternr: +related to ws transport cleanup. We could get rid of this fallback logic I think.

packages/sandbox/src/clients/transport/http-transport.ts
```diff
@@ -98,4 +98,36 @@ export class HttpTransport extends BaseTransport {
...
+  }
+
+  onPtyExit(_ptyId: string, _callback: (exitCode: number) => void): () => void {
+    // HTTP transport doesn't support real-time PTY exit events.
+    // Exit must be detected via SSE stream (GET /api/pty/:id/stream).
+    this.logger.warn(
+      'onPtyExit() has no effect with HTTP transport. Use WebSocket transport for real-time events.'
+    );
+    return () => {};
+  }
```
ghostwriternr: +related to ws transport cleanup

packages/sandbox/src/sandbox.ts
```diff
@@ -2507,4 +2525,90 @@ export class Sandbox<Env = unknown> extends Container<Env> implements ISandbox {
...
+  // These methods return serializable data for RPC compatibility
+  // ============================================================================
+
+  /**
+   * Create a new PTY session
+   *
+   * @param options - PTY creation options (terminal size, command, cwd, etc.)
+   * @returns PTY info object (use getPtyById to get an interactive handle
+   */
+  async createPty(
```
ghostwriternr: All the top-level methods being here are maybe an artefact of leftovers from before the decision to namespace all pty related methods under `pty` was made. `createPty`, `writeToPty`, `resizePty`, `killPty`, `listPtys`, `getPtyInfo`, `attachPty` all seem to be redundant to the `sandbox.pty.*` methods and can be removed, so we give users a single, clean way to do these?

packages/sandbox/src/clients/transport/types.ts
ghostwriternr: And rough pseudocode for how pty-client (?) changes as a result could be:
```ts
private ptyTransport: WebSocketTransport | null = null;

private async getPtyTransport(): Promise<WebSocketTransport> {
  if (!this.ptyTransport) {
    this.ptyTransport = new WebSocketTransport({...});
    await this.ptyTransport.connect();
  }
  return this.ptyTransport;
}

async create(options?: CreatePtyOptions): Promise<Pty> {
  const transport = await this.getPtyTransport();
  // ... create PTY, pass transport to PtyHandle
}
```

packages/sandbox-container/src/server.ts
```diff
@@ -33,7 +33,8 @@ async function createApplication(): Promise<{
   setupRoutes(router, container);
 
   // Create WebSocket adapter with the router for control plane multiplexing
-  const wsAdapter = new WebSocketAdapter(router, logger);
+  const ptyManager = container.get('ptyManager');
+  const wsAdapter = new WebSocketAdapter(router, ptyManager, logger);
```
ghostwriternr: +related to ws cleanup

packages/sandbox-container/src/managers/pty-manager.ts
```diff
@@ -0,0 +1,449 @@
...
+
+/**
+ * Minimal interface for Bun.Terminal (introduced in Bun v1.3.5+)
+ * Defined locally since it's only used in the container runtime.
+ * @types/bun doesn't include this yet, so we define it here.
+ */
+interface BunTerminal {
+  write(data: string): void;
+  resize(cols: number, rows: number): void;
+}
```
ghostwriternr: Here, and maybe a few other places, I noticed we're re-defining some bun types ourselves, but possibly the much cleaner approach would be to use `"@types/bun": "^1.3.5"` which should have all these types and use them directly instead.

packages/sandbox/src/clients/pty-client.ts
```diff
@@ -0,0 +1,479 @@
...
+      method: 'POST',
+      headers: { 'Content-Type': 'application/json' },
+      body: JSON.stringify({ data })
+    });
+
+    // Use handleResponse to properly parse ErrorResponse on failure
+    await this.handleResponse<{ success: boolean }>(response);
+
+    this.logSuccess('PTY input sent', id);
+  }
```
ghostwriternr: Any reason why this exists here, as well as in `PtyHandle` (where it does seem more appropriate)? We can probably delete these?

packages/sandbox-container/src/managers/pty-manager.ts
```diff
@@ -0,0 +1,449 @@
...
+      this.logger.error(
+        'Failed to kill PTY',
+        error instanceof Error ? error : undefined,
+        { ptyId: id, signal }
+      );
+      return { success: false, error: message };
+    }
+  }
+
+  killAll(): void {
```
ghostwriternr: nit: this should probably be called from the `cleanup` function in server.ts?

packages/shared/src/errors/codes.ts
```diff
@@ -45,6 +45,7 @@ export const ErrorCode = {
 
   // Session Errors (409)
   SESSION_ALREADY_EXISTS: 'SESSION_ALREADY_EXISTS',
+  PTY_EXCLUSIVE_CONTROL: 'PTY_EXCLUSIVE_CONTROL',
```
ghostwriternr: Nit: This gets eliminated based on the comment re: attach feature (regardless of whether attach itself gets removed, this is bound to go I think). But perhaps the right set of error codes might be `PTY_NOT_FOUND`, `PTY_WRITE_FAILED`, `PTY_ALREADY_EXITED` that gets used appropriately. Right now, this PR does not use our overall error handling system and instead uses vanilla `Error` which doesn't surface up in useful manner.

packages/sandbox-container/src/managers/pty-manager.ts
whoiskatrin: It's due to this bug in bun.terminal https://github.com/oven-sh/bun/pull/25834

PR #313, opened 2025-12-23
build.ts: resolve bun PATH robustly in Ubuntu snap

sourman:
update build to grab the bun path before spawn

bg: I have `bun` installed on my wsl ubuntu flavour.
The build would fail because it did not find bun executable on path.
In my Ubuntu flavoured wsl, bun is installed using snap


ghostwriternr:
Hey @sourman ! Thanks for the fix. Sorry about the delay - just catching up on PRs after the break - highly appreciate the PR : )

PR #324, opened 2026-01-02
feat: add file watching capabilities with inotify support

whoiskatrin:
## User problems this solves
This feature is designed for concrete workflows, not just API surface:
 
1. **Hot reload / dev servers** — Trigger rebuilds or reloads when source files change.
2. **Incremental indexing / analysis** — Re-index only changed files instead of rescanning whole trees.
3. **Test reruns** — Re-run affected tests when test files or dependencies change.
 
## API
 
```ts
import { parseSSEStream } from "@cloudflare/sandbox";
import type { FileWatchSSEEvent } from "@cloudflare/sandbox";
 
// watch() resolves only after the watcher is established on the filesystem.
// Callers can immediately perform actions that depend on the watch being active.
const stream = await sandbox.watch("/app/src", {
  recursive: true,
  include: ["*.ts", "*.js"],
});
 
// Consume the SSE stream with parseSSEStream
const controller = new AbortController();
for await (const event of parseSSEStream<FileWatchSSEEvent>(
  stream,
  controller.signal
)) {
  if (event.type === "event") {
    console.log(event.eventType, event.path);
    // e.g. "create" "/app/src/new-file.ts"
  }
}
 
// Cancel by aborting the signal — cleans up the watcher server-side
controller.abort();
```
 
### Key behaviors
- **Deterministic readiness**: `watch()` blocks until the filesystem watcher is established. No need to scan for a `watching` event before performing actions.
- **Stream-based**: Returns `ReadableStream<Uint8Array>` consumed via `parseSSEStream`. This is the only viable approach given the RPC layer (AsyncIterable cannot cross the boundary).
- **Cancellation via AbortSignal**: `parseSSEStream` registers an abort listener that calls `reader.cancel()`, which propagates server-side cleanup. No explicit stop endpoint needed.
- **`include`/`exclude` are mutually exclusive**: Validated at the container layer since `inotifywait` does not support both simultaneously.
 
---
 
## Event flow
```
inotifywait (container)
        ↓
WatchService emits SSE frames
        ↓
parseSSEStream yields typed events
        ↓
User iterates with for-await
```
 
---
 
## Hardening included in this PR
 
- **Watch setup correctness** — Fails watch startup on setup timeout/error (does not emit false `watching`). `watch()` throws if the watcher fails to establish.
- **Request validation** — Validates `path`, `recursive`, `include`, `exclude`. Rejects both `include` and `exclude` together.
- **Path safety** — Normalizes watch paths and enforces `/workspace` boundary.
- **Backpressure protection** — Coalesces events and caps pending queue at 1000; logs dropped-event pressure.
- **Stream cancellation cleanup** — Client-side `reader.cancel()` or `AbortSignal` abort propagates to kill the `inotifywait` process server-side. No orphaned watchers.
- **WS per-connection stream ownership** — `onClose` only cancels streams belonging to the closing WebSocket connection, not all active streams.
- **Glob behavior clarity** — Supported syntax is intentionally limited (`*`, `**`, `?`, path separators). Unsupported complex tokens are rejected via validation.
 
---
 
### 1) SDK watch client + readiness
- `packages/sandbox/src/clients/watch-client.ts`
- `packages/sandbox/src/sandbox.ts`
- Focus: `waitForReadiness()` reads chunks until `watching` event, returns wrapper stream that replays buffered data
 
### 2) SSE parser abort handling
- `packages/sandbox/src/sse-parser.ts`
- Focus: abort listener calls `reader.cancel()` to unblock idle streams; cleanup removes listener
 
### 3) Container watch routing/validation
- `packages/sandbox-container/src/handlers/watch-handler.ts`
- `packages/sandbox-container/src/routes/setup.ts`
- Focus: request validation, `include`/`exclude` mutual exclusivity, path normalization
 
### 4) Watch process + stream robustness
- `packages/sandbox-container/src/services/watch-service.ts`
- Focus: setup-timeout behavior, backpressure controls, stderr/watch process cleanup
 
### 5) WebSocket adapter stream ownership
- `packages/sandbox-container/src/handlers/ws-adapter.ts`
- Focus: per-connection `activeStreams` tracking, scoped cancellation on close
 
### 6) End-to-end coverage
- `tests/e2e/file-watch-workflow.test.ts`
- `tests/e2e/test-worker/index.ts`
- Focus: watch establishment, include/exclude, stream cancellation, error paths. Uses `parseSSEStream` + `AbortSignal.timeout` with no hardcoded delays.
 
---
 
## Key design decisions
1. **Stream-based, not handle-based** — Raw SSE stream + `parseSSEStream` is the only approach that works across the RPC boundary. No `onEvent` callback or `watcher.stop()`.
2. **Deterministic readiness** — `watch()` blocks until established so callers know the watcher is active before performing filesystem actions.
3. **Cancellation via signal, not stop endpoint** — Aborting the `AbortSignal` or canceling the stream reader propagates cleanup. No `/api/watch/stop` needed.
4. **Operationally safe defaults** — Timeouts, cleanup, bounded buffering (max 1000 pending events).
5. **Explicit validation and constrained pattern support** — Predictable behavior over silent surprises.'

### DIFF
--- a/packages/sandbox-container/src/handlers/watch-handler.ts
+++ b/packages/sandbox-container/src/handlers/watch-handler.ts
@@ -1,11 +1,12 @@
 import { posix as pathPosix } from 'node:path';
 import type { Logger, WatchRequest } from '@repo/shared';
 import { ErrorCode } from '@repo/shared/errors';
+import { CONFIG } from '../config';
 import type { RequestContext } from '../core/types';
 import type { WatchService } from '../services/watch-service';
 import { BaseHandler } from './base-handler';
 
-const WORKSPACE_ROOT = '/workspace';
+const WORKSPACE_ROOT = CONFIG.DEFAULT_CWD;
 
 /**
  * Handler for file watch operations
--- a/packages/sandbox/src/clients/watch-client.ts
+++ b/packages/sandbox/src/clients/watch-client.ts
@@ -1,4 +1,9 @@
-import type { WatchRequest } from '@repo/shared';
+import {
+  type FileWatchSSEEvent,
+  parseSSEFrames,
+  type SSEPartialEvent,
+  type WatchRequest
+} from '@repo/shared';
 import { BaseHttpClient } from './base-client';
 
 /**
@@ -42,26 +47,57 @@ export class WatchClient extends BaseHttpClient {
     const reader = stream.getReader();
     const bufferedChunks: Uint8Array[] = [];
     const decoder = new TextDecoder();
-    let accumulated = '';
+    let buffer = '';
+    let currentEvent: SSEPartialEvent = { data: [] };
+    let watcherReady = false;
+
+    const processEventData = (eventData: string) => {
+      let event: FileWatchSSEEvent | undefined;
+      try {
+        event = JSON.parse(eventData) as FileWatchSSEEvent;
+      } catch {
+        return;
+      }
+
+      if (event.type === 'watching') {
+        watcherReady = true;
+      }
+
+      if (event.type === 'error') {
+        throw new Error(event.error || 'Watch failed to establish');
+      }
+    };
 
     try {
-      while (true) {
+      while (!watcherReady) {
         const { done, value } = await reader.read();
         if (done) {
+          const finalParsed = parseSSEFrames(`${buffer}\n\n`, currentEvent);
+          for (const frame of finalParsed.events) {
+            processEventData(frame.data);
+            if (watcherReady) {
+              break;
+            }
+          }
+
+          if (watcherReady) {
+            break;
+          }
+
           throw new Error('Watch stream ended before watcher was established');
         }
+
         bufferedChunks.push(value);
-        accumulated += decoder.decode(value, { stream: true });
+        buffer += decoder.decode(value, { stream: true });
+        const parsed = parseSSEFrames(buffer, currentEvent);
+        buffer = parsed.remaining;
+        currentEvent = parsed.currentEvent;
 
-        if (accumulated.includes('"type":"watching"')) {
-          break;
-        }
-
-        if (accumulated.includes('"type":"error"')) {
-          const match = accumulated.match(/"error"\s*:\s*"([^"]+)"/);
-          const msg = match?.[1] || 'Watch failed to establish';
-          reader.cancel().catch(() => {});
-          throw new Error(msg);
+        for (const frame of parsed.events) {
+          processEventData(frame.data);
+          if (watcherReady) {
+            break;
+          }
         }
       }
     } catch (error) {


parssak:
hey! I noticed there are [docs for this](https://developers.cloudflare.com/sandbox/guides/file-watching/) but doesn't seem like the changes are live yet

would love to use this API, any timelines on when this'll get merged?

whoiskatrin:
/bonk resolve conflict and review this PR

whoiskatrin:
> hey! I noticed there are [docs for this](https://developers.cloudflare.com/sandbox/guides/file-watching/) but doesn't seem like the changes are live yet
> 
> would love to use this API, any timelines on when this'll get merged?

really keen to get this in this week or next week. sorry about the docs, someone got them merged ahead of time

whoiskatrin:
/bonk fix these code review issues that you listed

whoiskatrin:
/bonk investigate tests failures and fix if needed

parssak:
@whoiskatrin any rough estimates as to when this'll be in the next release? 🙏🏻

whoiskatrin:
@parssak this will get in this week

parssak:
That's great to hear! Tysm

payamfakoorziba:
@whoiskatrin We'd also love to get this feature!

tests/e2e/test-worker/index.ts
```diff
@@ -997,6 +997,78 @@ console.log('Terminal server on port ' + port);
...
+          recursive: body.recursive,
+          include: body.include,
+          // inotifywait rejects include+exclude together. Keep include-only
+          // requests compatible with both current and older container images.
+          exclude: body.include ? [] : body.exclude,
+          sessionId: sessionId ?? undefined
+        };
+
+        const upstream = await sandbox.containerFetch(
+          new Request('http://localhost:3000/api/watch', {
```
ghostwriternr: This is probably a sign that we're unable to use the SDK methods directly from the worker, which is an anti-pattern we want to avoid. The highest likelihood for why this could be happening is that `WatchHandle` is currently not designed to be serializable over our runtime RPC layer, and fixing this might require a lot of changes in the DO layer of this implementation.

packages/sandbox-container/src/handlers/ws-adapter.ts
```diff
@@ -62,6 +82,16 @@ export class WebSocketAdapter {
...
     });
+
+    for (const [requestId, stream] of this.activeStreams) {
+      void stream.cancel().catch((error) => {
+        this.logger.debug('Failed to cancel stream on socket close', {
+          requestId,
+          error: error instanceof Error ? error.message : String(error)
+        });
+      });
+      this.activeStreams.delete(requestId);
```
ghostwriternr, line 99: This would close all file watchers that are open at a given time I think - worth testing

packages/sandbox/src/clients/transport/ws-transport.ts
```diff
@@ -365,75 +371,133 @@ export class WebSocketTransport extends BaseTransport {
...
+        );
+        if (firstMessageReceived) {
+          streamController?.error(error);
+        } else {
+          rejectStream(error);
+        }
+      }, timeoutMs);
+
+      // Create the stream but don't return it until we get the first message
+      const stream = new ReadableStream<Uint8Array>({
```
ghostwriternr, line 395: I'm generally super wary of timeout races and try to avoid using it if we could, but in this particular case, it feels like it's only partially handling the cleanup upon reject too because this stream object is left as-is. Can we add some more testing around this, especially since changes at the transport layer affect _all operations_ and we're changing a fundamental behaviour here (although I think I understand and like the rationale) by waiting for for the first message till we return.

packages/sandbox-container/src/handlers/ws-adapter.ts
```diff
@@ -198,39 +291,65 @@ export class WebSocketAdapter {
...
-    const reader = response.body.getReader();
     const decoder = new TextDecoder();
     let buffer = '';
+    // Track partial event state across chunks
+    let currentEvent: SSEPartialEvent = { data: [] };
+    let chunkCount = 0;
 
     try {
       while (true) {
+        this.logger.debug('Waiting for stream chunk', {
```
ghostwriternr: Nit: this (and many other debug logs on this file) will spam local dev logs for all users, and maybe can be trimmed a bit

packages/sandbox-container/src/services/watch-service.ts
```diff
@@ -0,0 +1,671 @@
...
+            droppedEvents++;
+            if (droppedEvents === 1 || droppedEvents % 100 === 0) {
+              logger.warn('Dropping watch events due to backpressure', {
+                watchId,
+                droppedEvents,
+                pendingCount: pendingEvents.size
+              });
+            }
+            return;
+          }
```
ghostwriternr, line 329: Is this intentional? Do we want to be dropping events? This would mean we can't offer strong guarantees on this new API.

packages/sandbox-container/src/services/watch-service.ts
```diff
@@ -0,0 +1,671 @@
...
+        }
+      }
+    };
+
+    try {
+      const result = await Promise.race([
+        readLoop(),
+        new Promise<'timeout'>((resolve) =>
+          setTimeout(() => resolve('timeout'), WATCH_SETUP_TIMEOUT_MS)
+        )
```
ghostwriternr: Nit: this timeout won't get cleared if readLoop wins the race

ghostwriternr (changes requested):
Biggest issue is perhaps confirming the API usage works as expected + some minor issues

tests/e2e/test-worker/index.ts
ghostwriternr: I'd written this earlier - probably resolved with newer commits

packages/sandbox-container/src/services/watch-service.ts
whoiskatrin: We should treat watch() as a notification stream, not a guaranteed audit log. Given that model, bounded queue + drop-on-overflow is intentional for reliability.

tests/e2e/file-watch-workflow.test.ts
```diff
@@ -0,0 +1,471 @@
...
+
+  afterAll(async () => {
+    await cleanupIsolatedSandbox(sandbox);
+    sandbox = null;
+  });
+
+  /**
+   * Helper to start a watch that allows performing actions after the watch is established.
+   */
+  async function watchWithActions<T>(
```
ghostwriternr, line 56: This helper is doing a _very heavy lift here_ in place of the SDK. It is currently manually parsing SSE, manages reader lifecycle, detects readiness, handles timeouts, and collects typed events - all things `parseSSEStream` + `watch()` should provide out of the box.

I'd encourage comparing this to how something like `execStream` works: users call `sandbox.execStream()`, pipe through `parseSSEStream()`, and iterate typed events with `for await`, which is ~5 lines. The equivalent for watch() is users needing to write some version of this 90-line function.

If the goal is to provide the DX the original PR description promises (which also needs updating since the API surface has changed a bit due to the RPC constraints), `watch()` should return something users can consume with `for await` and an abort signal, just like `execStream` + `parseSSEStream`.

tests/e2e/file-watch-workflow.test.ts
```diff
@@ -0,0 +1,471 @@
...
+    });
+
+    if (!response.ok || !response.body) {
+      throw new Error(`Watch request failed: ${response.status}`);
+    }
+
+    const events: FileWatchSSEEvent[] = [];
+    let watchId: string | null = null;
+    let actionResult: T;
+    const reader = response.body.getReader();
```
ghostwriternr: Our SDK currently exports `parseSSEStream` and even the JSDoc on the new `watch()` method recommends using it. But this test can't, because `parseSSEStream` doesn't currently support cancellation of long-lived streams. Its abort check only runs between `reader.read()` calls and on an idle watch stream, `reader.read()` blocks indefinitely, so the signal is never checked.

So maybe the hint here is to get `parseSSEStream` to register an abort listener that calls `reader.cancel()` when the signal fires, rather than polling `signal.aborted` between reads. It's probably just a one-line fix (`signal.addEventListener('abort', () => reader.cancel())`) (pls test properly though, I haven't tried this yet) that would make the entire `watchWithActions` helper unnecessary.

tests/e2e/file-watch-workflow.test.ts
```diff
@@ -0,0 +1,471 @@
...
+                actionsExecuted = true;
+                // Pause the timeout while executing actions so the full
+                // timeoutMs window is available for collecting events afterward
+                clearTimeout(timeoutHandle);
+                await new Promise((r) => setTimeout(r, 1500));
+                actionResult = await actions();
+                // Give events time to propagate through the SSE stream
+                await new Promise((r) => setTimeout(r, 3000));
+                // Reset the timeout so we have the full window to read events
+                timeoutHandle = setTimeout(() => reader.cancel(), timeoutMs);
```
ghostwriternr: This is a manual reimplementation of AbortSignal.timeout() that this test has to do because passing `AbortSignal.timeout(5000)` to `parseSSEStream` doesn't actually cancel the stream (due to reasons I shared above). The "cancellable + resettable timeout" pattern is a major footgun because every user building a watch UI will need to solve the same "how do I stop listening after N seconds of inactivity" problem.

Once `parseSSEStream` handles abort correctly, this likely just becomes `AbortSignal.timeout(timeoutMs)` and need zero custom code.

tests/e2e/file-watch-workflow.test.ts
```diff
@@ -0,0 +1,471 @@
...
+              watchId = event.watchId;
+              if (!actionsExecuted) {
+                actionsExecuted = true;
+                // Pause the timeout while executing actions so the full
+                // timeoutMs window is available for collecting events afterward
+                clearTimeout(timeoutHandle);
+                await new Promise((r) => setTimeout(r, 1500));
+                actionResult = await actions();
+                // Give events time to propagate through the SSE stream
+                await new Promise((r) => setTimeout(r, 3000));
```
ghostwriternr: As per the latest design, `watch` resolves immediately with a raw ReadableStream before the watcher is established on the filesystem. So this test has to manually scan the SSE stream for the watching event to know it's safe to create files — which is a bit absurd.

IMO, `watch` should either not resolve until the watcher is established (like how `exec` doesn't resolve until the command finishes), so that users don't need to write their own readiness-detection loop.

ghostwriternr: Any time we are using magic timeouts, that's a big code/design smell imo.

ghostwriternr: In fact, the more I'm reading this test file, I'm noticing there are 6 different hardcoded delays, all compensating for the lack of a readiness signal and event delivery guarantees. This 1500ms pre-action sleep is especially telling because the test already received the watching event but still doesn't trust that the watcher is ready. And the 3000ms one suggests we don't have a way to even confirm event delivery?

These are the kind of things that don't just make our CI flaky, but also user production code.

tests/e2e/file-watch-workflow.test.ts
```diff
@@ -0,0 +1,471 @@
...
+  }, 30000);
+
+  test('should filter events with include pattern', async () => {
+    testDir = uniqueTestPath('watch-filter');
+    await createDir(testDir);
+
+    // Start watch and create files after watch is confirmed ready
+    const { events } = await watchWithActions(
+      testDir,
+      { include: ['*.ts'], timeoutMs: 10000, stopAfterEvents: 10 },
```
ghostwriternr, line 285: This comment applies here, as well as the `should exclude patterns from events` test below.

These two tests carefully avoid combining both `include` and `exclude`. And the reason is because the test worker has this line:
```ts
exclude: body.include ? [] : body.exclude,
```

The test worker silently drops `exclude` whenever `include` is present because `inotifywait` doesn't support both simultaneously. But `WatchOptions` accepts both `include` and `exclude` and nothing in the type specification or elsewhere tells users they're mutually exclusive.

We should either make the type honest: use a discriminated union or overloads so include and exclude can't be specified together. OR - and this is what I'd prefer - implement the combination at the SDK/container layer by maybe filtering in the SSE emitter if `inotify` can't handle it.

We shouldn't be hiding these behaviours in the test worker.

tests/e2e/file-watch-workflow.test.ts
```diff
@@ -0,0 +1,471 @@
...
+    const decoder = new TextDecoder();
+
+    // Wait for watch to establish
+    let watchingReceived = false;
+    while (!watchingReceived) {
+      const { done, value } = await reader.read();
+      if (done) break;
+      const text = decoder.decode(value, { stream: true });
+      watchingReceived = text.includes('"type":"watching"');
+    }
```
ghostwriternr: One more sign of even the simplest test to start a watch, then stop it can't use `parseSSEStream`. It has to manually read chunks and do string matching instead of iterating typed events. Further evidence that `parseSSEStream's` abort handling is the cleanest place to improve things across the board.

tests/e2e/file-watch-workflow.test.ts
```diff
@@ -0,0 +1,471 @@
...
+        streamError = e instanceof Error ? e : new Error(String(e));
+      } finally {
+        clearTimeout(timeout);
+        reader.cancel().catch(() => {});
+      }
+
+      // Should contain an error message in either the text or the stream error
+      const errorContent = text || streamError?.message || '';
+      expect(errorContent).toMatch(/error|not found|does not exist/i);
+    }
```
ghostwriternr: This test checks for errors via HTTP status code, SSE stream content and stream exception. If this is intentional and necessary, a user calling `sandbox.watch('/nonexistent')` has to defend against all three paths and maybe a hint at the error handling story for watch not being properly done too.

whoiskatrin:
/bonk check if I missed anything based on Naresh's last review

.changeset/empty-poets-serve.md
```diff
@@ -0,0 +1,16 @@
...
+
+Use `sandbox.watch()` to monitor directories for create, modify, delete, and rename events with native inotify:
+
+```typescript
+const watcher = await sandbox.watch('/app/src', {
+  recursive: true,
+  include: ['*.ts', '*.js'],
+  onEvent: (event) => console.log(event.type, event.path)
+});
+await watcher.stop();
```
ghostwriternr:
```suggestion
`sandbox.watch()` returns an SSE stream of create, modify, delete, and move events using native inotify. The stream can be proxied directly to a client or consumed server-side with `parseSSEStream`:

```typescript
// Stream events to a browser client
const stream = await sandbox.watch('/workspace/src', {
  recursive: true,
  include: ['*.ts', '*.js']
});
return new Response(stream, {
  headers: { 'Content-Type': 'text/event-stream' }
});

// Or consume server-side
for await (const event of parseSSEStream<FileWatchSSEEvent>(stream)) {
  console.log(event.type, event.path);
}
```

packages/sandbox-container/src/handlers/watch-handler.ts
```diff
@@ -0,0 +1,242 @@
+import { posix as pathPosix } from 'node:path';
+import type { Logger, WatchRequest } from '@repo/shared';
+import { ErrorCode } from '@repo/shared/errors';
+import type { RequestContext } from '../core/types';
+import type { WatchService } from '../services/watch-service';
+import { BaseHandler } from './base-handler';
+
+const WORKSPACE_ROOT = '/workspace';
```
ghostwriternr: Should maybe use `CONFIG.DEFAULT_CWD`, considering the default cwd is configurable by the user?

packages/sandbox/src/clients/watch-client.ts
```diff
@@ -0,0 +1,93 @@
...
+        if (accumulated.includes('"type":"watching"')) {
+          break;
+        }
+
+        if (accumulated.includes('"type":"error"')) {
+          const match = accumulated.match(/"error"\s*:\s*"([^"]+)"/);
+          const msg = match?.[1] || 'Watch failed to establish';
+          reader.cancel().catch(() => {});
+          throw new Error(msg);
+        }
```
ghostwriternr: Why string based matching, which can be flaky? I feel like for this whole loop we could be using one of our parsing utils considering it's just a readable stream?

PR #326, opened 2026-01-02
add depth option for shallow cloning in git operations

whoiskatrin:
Add git clone support with branch selection and shallow clone options

- Added GitClient in SDK with checkout() method for cloning repositories
- Added GitManager in container for pure business logic (URL parsing, command building, error handling)
- Added GitService in container for orchestrating clone operations via SessionManager
- Added GitHandler for /api/git/checkout endpoint
- Exposed gitCheckout() method on the Sandbox class


whoiskatrin:
Tested by the user as well, all is good.

Screenshot 2026-01-03 at 10 13 47

PR #329, opened 2026-01-03
Add custom token support to exposePort for stable preview URLs

mikenomitch:
Adds support for custom tokens in exposePort. - Allows for consistent URLs across sessions


whoiskatrin (approved):
LGTM!

PR #336, opened 2026-01-08
fix: handle potential undefined value for envVars

roerohan:
```sh
✘ [ERROR] Uncaught TypeError: Cannot read properties of undefined (reading 'replace') Error

      at setEnvVars

[wrangler:error] TypeError: Cannot read properties of undefined (reading 'replace')
    at async Object.fetch
```

If the value of the environment variable is undefined, it throws the above error. It's not easy to understand what is causing it, adding an optional chaining operator avoid the error. It is worth noting that the `exportCommand` will be effectively become `export ${key}='undefined'`.

### DIFF
--- a/packages/sandbox/src/sandbox.ts
+++ b/packages/sandbox/src/sandbox.ts
@@ -286,7 +286,7 @@ export class Sandbox<Env = unknown> extends Container<Env> implements ISandbox {
     if (this.defaultSession) {
       // Set environment variables by executing export commands in the existing session
       for (const [key, value] of Object.entries(envVars)) {
-        const escapedValue = value.replace(/'/g, "'\\''");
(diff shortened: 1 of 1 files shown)


ghostwriternr:
Hey @roerohan ! In terms of making this a very complete fix, I opened #342 - what do you think about the approach there? I realised thanks to your PR that there was no clear way to unset env variables anyway and maybe it should be supported proper by doing this?

roerohan:
> Hey @roerohan ! In terms of making this a very complete fix, I opened #342 - what do you think about the approach there? I realised thanks to your PR that there was no clear way to unset env variables anyway and maybe it should be supported proper by doing this?

Looks good! I'm closing this one, let's merge #342 instead!

PR #337, opened 2026-01-08
fix: Propagate SIGTERM to child processes in killProcess

pandemicsyn:
This fixes an issue where killProcess returns success but the actual process continues running.

When killProcess is called, it sends SIGTERM to the PID stored in the PID file. However, this PID is the subshell wrapper, not the actual command.

Added trap 'kill 0' TERM inside the background subshell. When the subshell receives SIGTERM, the trap fires and sends SIGTERM to all processes in the current process group (the subshell and its children).

fixes https://github.com/cloudflare/sandbox-sdk/issues/338

### DIFF
--- a/.changeset/olive-lands-do.md
+++ b/.changeset/olive-lands-do.md
@@ -1,0 +1,5 @@
+---
+'@cloudflare/sandbox': patch
+---
+
+Fill killProcess not terminating child processes


ghostwriternr:
Hey @pandemicsyn! Thanks for identifying this issue - The bug report is indeed accurate - `killProcess()` wasn't terminating child processes.

However, trap 'kill 0' TERM seems to have a problem: in non-interactive bash (which we use), all processes share the same process group. kill 0 would terminate the entire session, not just the target command. I was able to validate this with an E2E test too. I've instead opened #339 with an alternative fix that uses bash job control (`set -m`) to give each background command its own process group, then kills the group with `kill(-pid)`. This felt more reliable since it's handled by the kernel rather than depending on bash signal handlers.

To get this closed soon, I'm closing this PR in favour of #339 - thanks again for raising this!

pandemicsyn:
@ghostwriternr no worries , thanks for getting this squared away so quickly!

PR #353, opened 2026-01-14
fix opencode terminal connection

TimoWilhelm:
Fix WebSocket Proxy + CSP headers for terminal support in opencode web ui

Also fixes the home path to fix file listing

### DIFF
--- a/.changeset/many-deers-dance.md
+++ b/.changeset/many-deers-dance.md
@@ -1,5 +1,0 @@
----
-'@cloudflare/sandbox': patch
----
-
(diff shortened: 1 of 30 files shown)


scuffi:
Hey @TimoWilhelm,

Thanks for the PR, looks good & safe to merge if you can resolve the conflict?

Thanks!

scuffi:
/bonk add a changeset, test coverage and update the openCode example in the repo from these changes

scuffi:
@TimoWilhelm bonk seems to be having some trouble, instead, could you please push a changeset, some test coverage and update the OpenCode example now this method is async? 

Everything else looks good to merge after those changes!

Thanks for the PR :)

PR #362, opened 2026-01-25
Add stdin support to exec()

### LINKED ISSUE #357: Add stdin support to exec()
## Summary

The `exec()` method currently runs commands through a shell and doesn't support passing stdin data or per-command environment variables. This makes it difficult to safely pass arbitrary user input without shell injection risks.

## Current behavior

```typescript
// SDK v0.0.7 signature
exec(command: string, args: string[], options?: { stream?: boolean })
```

- Commands appear to run through a shell
- No stdin support
- No per-command environment variables (despite being documented)

## Proposed API

```typescript
const result = await sandbox.exec("claude", ["-p", "-"], {
  stdin: prompt,           // Pass data to process stdin
  env: { API_KEY: "..." }, // Per-command env vars (already documented)
});
```

## Use case

We're building a Discord bot that runs Claude Code in a sandbox. User messages need to be passed as prompts, but they can contain arbitrary characters (quotes, backticks, $, etc.). Currently we have to:

1. Write the prompt to a file via `writeFile()`
2. Pipe it through cat: `cat /tmp/prompt.txt | claude -p -`

With stdin support, this becomes:
```typescript
await sandbox.exec("claude", ["-p", "-"], { stdin: userMessage });
```

## Benefits

- **Security**: Direct process spawn without shell interpretation
- **Simplicity**: No need for file-based workarounds
- **Consistency**: Per-command env is already documented but not implemented

## Documentation reference

The [environment variables docs](https://developers.cloudflare.com/sandbox/configuration/environment-variables/) show per-command env support:

```typescript
await sandbox.exec("node app.js", {
  env: { NODE_ENV: "production" }
});
```

But this isn't implemented in SDK v0.0.7.

whoiskatrin:
## Summary

Adds `stdin` option to `exec()` method, enabling users to pass arbitrary input to commands without shell injection risks.

- Input is written to a temp file and redirected to the command's stdin
- Temp file is cleaned up after command execution
- Passes stdin through all layers: SDK → Handler → Service → Session

## Usage

```typescript
const result = await sandbox.exec('cat', { stdin: 'hello world' });
console.log(result.stdout); // "hello world"
```

Closes #357

### DIFF
--- a/.changeset/stdin-support.md
+++ b/.changeset/stdin-support.md
@@ -1,0 +1,5 @@
+---
+'@cloudflare/sandbox': minor
+---
+
+Add stdin support to exec() for passing arbitrary input to commands without shell injection risks. Enable with `stdin` option: `sandbox.exec('cat', { stdin: 'hello world' })`.


ghostwriternr:
As discussed on chat a few weeks ago, closing this in favour of doing a proper PTY API implementation that will be far more flexible and handle all sorts of ways to pass and read from stdin rather than the specific approach this implementation takes.

PR #393, opened 2026-02-13
Fix: Opencode <> Cloudflare AI Gateway integration

AshishKumar4:
Fixes the bug where SDK's expected provider string 'cloudflareAIGateway' (camelCase) isn't recognised by opencode, and thus the opencode example fails to work with ai gateway.

Also improved the example code with instructions for ai gateway via unified billing

### DIFF
--- a/packages/sandbox/src/opencode/opencode.ts
+++ b/packages/sandbox/src/opencode/opencode.ts
@@ -277,8 +277,7 @@ async function startOpencodeServer(
  *         options: { apiKey: env.ANTHROPIC_KEY }
  *       },
  *       // Or use Cloudflare AI Gateway (with unified billing, no provider keys needed).
- *       // The SDK remaps 'cloudflareAIGateway' to 'cloudflare-ai-gateway' automatically.
- *       // cloudflareAIGateway: {
+ *       // 'cloudflare-ai-gateway': {
  *       //   options: {
  *       //     accountId: env.CF_ACCOUNT_ID,
  *       //     gatewayId: env.CF_GATEWAY_ID,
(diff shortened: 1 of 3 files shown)


packages/sandbox/src/opencode/opencode.ts
```diff
@@ -153,6 +153,34 @@ async function ensureOpencodeServer(
   }
 }
 
+/**
+ * The SDK accepts 'cloudflareAIGateway' (camelCase) as a convenience alias,
+ * but the OpenCode CLI recognizes 'cloudflare-ai-gateway' (hyphenated).
+ * Remap the provider key so the server picks it up correctly.
+ */
+function normalizeConfigForOpencode(config: Config): Config {
```
ghostwriternr: Should we even support `cloudflareAIGateway`? If its not supported by opencode, I'm okay with breaking this instead of keeping a patch.

ghostwriternr:
Minor comment, looks good otherwise. Thanks for fixing this!

packages/sandbox/src/opencode/opencode.ts
AshishKumar4: the only thing is, we would have to put 'cloudflare-ai-gateway' in quotes then, so its like a convenience method

ghostwriternr: (because if the camel case version doesn't work, no one is using this anyway)

AshishKumar4: Done

ghostwriternr: Thanks! Looks like it still needs to be removed from the comments/docs etc., seeing from the final diff. All good to go if those are resolved too.

AshishKumar4: Forgot about those, done, pushed

PR #396, opened 2026-02-16
feat: Add backup and restore API

whoiskatrin:
## Summary

Adds backup and restore functionality for sandboxes using squashfs archives stored in R2. This allows users to create point-in-time snapshots of directories and restore them later.

**Core Features:**
- `createBackup(dir, options)` - Creates a compressed squashfs archive of a directory and uploads to R2
- `restoreBackup(id, options)` - Downloads and extracts a backup to restore files
- TTL-based expiration with configurable time-to-live (default: 1 hour, max: 24 hours)
- Chunked uploads/downloads for large archives (50MB chunks, up to 500MB total)

**Implementation:**
- **packages/sandbox/src/sandbox.ts** - Main backup methods with chunked upload/download support using R2 multipart API
- **packages/sandbox/src/clients/backup-client.ts** - Type-safe HTTP client for backup operations
- **packages/sandbox-container/src/services/backup-service.ts** - Container-side service using mksquashfs/unsquashfs
- **packages/sandbox-container/src/handlers/backup-handler.ts** - Request validation and routing with path traversal protection
- **packages/shared/src/errors/** - New error codes: `BACKUP_NOT_FOUND`, `BACKUP_EXPIRED`, `INVALID_BACKUP_CONFIG`, `BACKUP_CREATE_FAILED`, `BACKUP_RESTORE_FAILED`

**Security:**
- Defense-in-depth validation at both DO and container layers
- Path traversal prevention (no `..` allowed, must be absolute paths)
- Archive storage restricted to `/var/backups/` in container
- 60-second buffer before expiration to prevent restore race conditions

## Test Plan

- [x] TypeScript type checking passes
- [x] E2E tests pass in CI (uses real R2 `sandbox-e2e-test` bucket)

### DIFF
--- a/.github/workflows/pullrequest.yml
+++ b/.github/workflows/pullrequest.yml
@@ -261,10 +261,14 @@ jobs:
             AWS_ACCESS_KEY_ID
             AWS_SECRET_ACCESS_KEY
             CLOUDFLARE_ACCOUNT_ID
+            R2_ACCESS_KEY_ID
+            R2_SECRET_ACCESS_KEY
         env:
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           CLOUDFLARE_ACCOUNT_ID: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
+          R2_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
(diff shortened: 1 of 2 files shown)


packages/shared/src/errors/contexts.ts
```diff
@@ -169,6 +169,36 @@ export interface InvalidMountConfigContext {
   reason?: string;
 }
 
+/**
+ * Backup error contexts
+ */
+export interface BackupCreateContext {
+  dir: string;
+  backupId?: string;
```
ghostwriternr: minor devx: In all places where we use `backupId`, why not just `id`?

Issue #398, opened 2026-02-18
ws-transport sets unconfigurable 120000ms timeout, breaking long running `execStream`


rikbrown:
https://github.com/cloudflare/sandbox-sdk/blob/9908a8bdf790f53c1a475252337a738c61aa580b/packages/sandbox/src/clients/transport/ws-transport.ts#L313

The timeout for the WS transport is set at 2 minutes.
This means if you run a process via `execStream` in the Sandbox, and it takes over 2 minutes (streaming from an agent, for example), the stream dies with an unhandled error `timeout after 120000ms`.

This also appears to be an issue when I start a process as a background process, and then connect to its log stream, presumably same code path.

Suggestion: allow this to be configurable in wrangler file and document that the default is only 2mins.
alternatively, if I'm missing something let me know!

whoiskatrin:
@rikbrown check if this makes your life better https://github.com/cloudflare/sandbox-sdk/pull/400, you can test using image and package in this PR

rikbrown:
Thank you! I will take a test drive this evening, just been randomized onto some less fun work for a bit.

rikbrown:
Testing this now

rikbrown:
Confirm this seems to work, thank you for the fast response!

PR #400, opened 2026-02-18
Fix WS stream timeout killing long-running streams

whoiskatrin:
## Summary

Convert WebSocket streaming timeout from a fixed 2-minute total-duration timeout to an inactivity-based idle timeout (default: 5 minutes) that resets on every chunk, so streams stay alive as long as data is flowing.

Closes https://github.com/cloudflare/sandbox-sdk/issues/398

---

<a href="https://app.devin.ai/review/cloudflare/sandbox-sdk/pull/400" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/packages/sandbox/src/clients/transport/ws-transport.ts
+++ b/packages/sandbox/src/clients/transport/ws-transport.ts
@@ -403,7 +403,11 @@ export class WebSocketTransport extends BaseTransport {
             `Stream idle timeout after ${idleTimeoutMs}ms: ${method} ${path}`
           );
           if (firstMessageReceived) {
-            streamController?.error(error);
+            try {
+              streamController?.error(error);
+            } catch {
+              // Stream controller may already be closed/errored
+            }
           } else {
             rejectStream(error);
           }
@@ -462,11 +466,15 @@ export class WebSocketTransport extends BaseTransport {
           } else {
             // Stream was already returned, now closing
             if (response.status >= 400) {
-              streamController?.error(
-                new Error(
-                  `Stream error: ${response.status} - ${JSON.stringify(response.body)}`
-                )
-              );
+              try {
+                streamController?.error(
+                  new Error(
+                    `Stream error: ${response.status} - ${JSON.stringify(response.body)}`
+                  )
+                );
+              } catch {
+                // Stream controller may already be closed/errored
+              }
             } else {
               streamController?.close();
             }
--- a/packages/sandbox/src/sandbox.ts
+++ b/packages/sandbox/src/sandbox.ts
@@ -114,10 +114,6 @@ export function getSandbox<T extends Sandbox<any>>(
     stub.setContainerTimeouts(options.containerTimeouts);
   }
 
-  if (options?.transportTimeouts) {
-    stub.setTransportTimeouts(options.transportTimeouts);
-  }
-
   const defaultSessionId = `sandbox-${effectiveId}`;
 
   // IMPORTANT: Any method that returns ExecutionSession must be listed here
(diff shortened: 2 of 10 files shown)


whoiskatrin:
/bonk fix the conflicts and after review this PR, fix small issues if needed

packages/sandbox/src/sandbox.ts
```diff
@@ -466,6 +499,21 @@ export class Sandbox<Env = unknown> extends Container<Env> implements ISandbox {
...
+    timeouts: NonNullable<SandboxOptions['transportTimeouts']>
+  ): Promise<void> {
+    this.transportTimeouts = { ...this.transportTimeouts, ...timeouts };
+
+    // Recreate client so the new timeouts take effect
+    this.client.disconnect();
+    this.client = this.createSandboxClient();
+
+    this.logger.debug('Transport timeouts updated', this.transportTimeouts);
+  }
```
ghostwriternr: Hmm this disconnects the WebSocket and recreates the entire client, which kills all in-flight requests and active streams - is this intentional??

This is also being called on every `getSandbox()` call with `transportTimeouts` set and there's no guard to skip when values haven't changed, so any concurrent streams get killed on every call. And even when values are changing, disconnect is maybe unnecessary. Can't we mutate the config in-place and have new values take effect naturally on the next timer reset?

packages/sandbox/src/clients/transport/ws-transport.ts
```diff
@@ -584,6 +606,11 @@ export class WebSocketTransport extends BaseTransport {
       return;
     }
 
+    // Reset the idle timeout since we received activity
+    if (pending.isStreaming) {
+      this.resetStreamIdleTimeout(chunk.id, pending);
+    }
+
     // Call onFirstChunk if this is the first chunk (triggers stream return)
     if (pending.onFirstChunk) {
```
ghostwriternr: When the very first chunk arrives, the idle timeout is reset before the stream is fully set up here. And if no more chunks arrive after that, the timeout fires but doesn't know how to reject the stream and it just silently cleans up, and the caller will hang forever waiting for a promise that never resolves or rejects.

Can we swap the order here so the stream is set up before the timeout is reset?

packages/sandbox/src/clients/transport/ws-transport.ts
```diff
@@ -372,25 +376,30 @@ export class WebSocketTransport extends BaseTransport {
       body
     };
 
+    const idleTimeoutMs = this.config.streamIdleTimeoutMs ?? 300_000;
```
ghostwriternr: nit: magic numbers should likely be a constant that's re-used everywhere

packages/sandbox/src/sandbox.ts
```diff
@@ -466,6 +499,21 @@ export class Sandbox<Env = unknown> extends Container<Env> implements ISandbox {
     await this.ctx.storage.put('keepAliveEnabled', keepAlive);
   }
 
+  /**
+   * RPC method to configure transport-level timeouts (WebSocket mode)
+   */
+  async setTransportTimeouts(
```
ghostwriternr: nit: we should maybe add validation here, to avoid zero or negative values, that would cause the idle timer to fire immediately on every chunk

ghostwriternr (changes requested):

packages/sandbox/src/clients/transport/ws-transport.ts
```diff
@@ -631,6 +677,31 @@ export class WebSocketTransport extends BaseTransport {
...
+        try {
+          pending.streamController.error(
+            new Error(`Stream idle timeout after ${idleTimeoutMs}ms`)
+          );
+        } catch {
+          // Stream may already be closed/errored
+        }
+      }
+    }, idleTimeoutMs);
+  }
```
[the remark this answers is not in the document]
ghostwriternr: `streamController` is always set before `resetStreamIdleTimeout` is called, `onFirstChunk` assigns it synchronously before the idle timer is reset, and it's never cleared afterward. So don't see the point of this.

PR #402, opened 2026-02-19
Use native file read operations

scuffi:
# Summary
This PR replaces the per-chunk file access with native file access through Bun. Instead of spawning one process per 65 KB chunk, we now read the file in a single Bun native file operation and stream it directly. Every 65 KB of file data requires spawning a new dd process (and base64 for binary). For a 10 MB file this goes from 163 shell execs → 3 shell execs (metadata) + 1 Bun file read. 

I've also added a maximum file size check before we attempt to read the file for RPC transfers. RPC transfers have a maximum transport size of 32MiB, so failing fast here stops us from unnecessarily loading & attempting to transfer the file out of the container. (You can see in the benchmark **After** sections there's a new error for files larger than >32MiB)

## Implementation
**packages/sandbox-container/src/services/file-service.ts** -> Change `read` and `readFileStreamOperation` to use `Bun.file()` access instead

## Benchmark
<details>

<summary>HTTP Transport</summary>

## Before
| File | Read method | Transport | Size | Time | Speed | OK | Error |
| :--- | :---------- | :-------- | ---: | ---: | -----: | :-: | :---- |
| arbitrary_1mb | readFileStream | http | 1.33 MB | 2.24 s | 609 KB/s | ✓ |  |
| arbitrary_1mb | readFile | http | 1.00 MB | 245 ms | 4.08 MB/s | ✓ |  |
| arbitrary_10mb | readFileStream | http | 13.34 MB | 9.08 s | 1.47 MB/s | ✓ |  |
| arbitrary_10mb | readFile | http | 10.00 MB | 504 ms | 19.84 MB/s | ✓ |  |
| arbitrary_32mb | readFileStream | http | 21.52 KB | 192 ms | 112 KB/s | ✓ |  |
| arbitrary_32mb | readFile | http | 16.00 KB | 186 ms | 86.02 KB/s | ✓ |  |
| arbitrary_50mb | readFileStream | http | 66.69 MB | 44.28 s | 1.51 MB/s | ✓ |  |
| arbitrary_50mb | readFile | http | — | — | — | ✗ | Serialized RPC arguments or return values are limited to 32… |
| arbitrary_100mb | readFileStream | http | 133 MB | 90.58 s | 1.47 MB/s | ✓ |  |
| arbitrary_100mb | readFile | http | — | — | — | ✗ | Serialized RPC arguments or return values are limited to 32… |
| arbitrary_500mb | readFileStream | http | 667 MB | 467.25 s | 1.43 MB/s | ✓ |  |
| arbitrary_500mb | readFile | http | — | — | — | ✗ | Cannot read properties of undefined (reading 'content') |

## After
| File | Read method | Transport | Size | Time | Speed | OK | Error |
| :--- | :---------- | :-------- | ---: | ---: | -----: | :-: | :---- |
| arbitrary_1mb | readFileStream | http | 1.33 MB | 104 ms | 12.83 MB/s | ✓ |  |
| arbitrary_1mb | readFile | http | 1.00 MB | 71 ms | 14.08 MB/s | ✓ |  |
| arbitrary_10mb | readFileStream | http | 13.34 MB | 148 ms | 90.13 MB/s | ✓ |  |
| arbitrary_10mb | readFile | http | 10.00 MB | 363 ms | 27.55 MB/s | ✓ |  |
| arbitrary_32mb | readFileStream | http | 21.52 KB | 54 ms | 399 KB/s | ✓ |  |
| arbitrary_32mb | readFile | http | 16.00 KB | 53 ms | 302 KB/s | ✓ |  |
| arbitrary_50mb | readFileStream | http | 66.69 MB | 445 ms | 150 MB/s | ✓ |  |
| arbitrary_50mb | readFile | http | — | — | — | ✗ | SandboxError: File too large. Size 52428800 bytes exceeds t… |
| arbitrary_100mb | readFileStream | http | 133 MB | 816 ms | 163 MB/s | ✓ |  |
| arbitrary_100mb | readFile | http | — | — | — | ✗ | SandboxError: File too large. Size 104857600 bytes exceeds … |
| arbitrary_500mb | readFileStream | http | 667 MB | 4.44 s | 150 MB/s | ✓ |  |
| arbitrary_500mb | readFile | http | — | — | — | ✗ | SandboxError: File too large. Size 524288000 bytes exceeds … |

</details>

<details>

<summary>Websocket Transport</summary>

## Before

| File | Read method | Transport | Size | Time | Speed | OK | Error |
| :--- | :---------- | :-------- | ---: | ---: | -----: | :-: | :---- |
| arbitrary_1mb | readFileStream | websocket | 1.33 MB | 1.13 s | 1.18 MB/s | ✓ |  |
| arbitrary_1mb | readFile | websocket | 1.00 MB | 222 ms | 4.50 MB/s | ✓ |  |
| arbitrary_10mb | readFileStream | websocket | 13.34 MB | 9.11 s | 1.46 MB/s | ✓ |  |
| arbitrary_10mb | readFile | websocket | 10.00 MB | 517 ms | 19.34 MB/s | ✓ |  |
| arbitrary_32mb | readFileStream | websocket | 21.52 KB | 200 ms | 108 KB/s | ✓ |  |
| arbitrary_32mb | readFile | websocket | 16.00 KB | 187 ms | 85.56 KB/s | ✓ |  |
| arbitrary_50mb | readFileStream | websocket | 66.69 MB | 44.48 s | 1.50 MB/s | ✓ |  |
| arbitrary_50mb | readFile | websocket | — | — | — | ✗ | Request timeout after 120000ms: POST /api/read |
| arbitrary_100mb | readFileStream | websocket | — | — | — | ✗ | Network connection lost. |
| arbitrary_100mb | readFile | websocket | — | — | — | ✗ | WebSocket not connected |
| arbitrary_500mb | readFileStream | websocket | — | — | — | ✗ | Network connection lost. |
| arbitrary_500mb | readFile | websocket | — | — | — | ✗ | WebSocket not connected |

## After

| File | Read method | Transport | Size | Time | Speed | OK | Error |
| :--- | :---------- | :-------- | ---: | ---: | -----: | :-: | :---- |
| arbitrary_1mb | readFileStream | websocket | 1.33 MB | 5.87 s | 233 KB/s | ✓ |  |
| arbitrary_1mb | readFile | websocket | 1.00 MB | 124 ms | 8.06 MB/s | ✓ |  |
| arbitrary_10mb | readFileStream | websocket | 13.34 MB | 180 ms | 74.10 MB/s | ✓ |  |
| arbitrary_10mb | readFile | websocket | 10.00 MB | 385 ms | 25.97 MB/s | ✓ |  |
| arbitrary_32mb | readFileStream | websocket | 21.52 KB | 55 ms | 391 KB/s | ✓ |  |
| arbitrary_32mb | readFile | websocket | 16.00 KB | 54 ms | 296 KB/s | ✓ |  |
| arbitrary_50mb | readFileStream | websocket | 66.69 MB | 483 ms | 138 MB/s | ✓ |  |
| arbitrary_50mb | readFile | websocket | — | — | — | ✗ | SandboxError: File too large. Size 52428800 bytes exceeds t… |
| arbitrary_100mb | readFileStream | websocket | 133 MB | 860 ms | 155 MB/s | ✓ |  |
| arbitrary_100mb | readFile | websocket | — | — | — | ✗ | SandboxError: File too large. Size 104857600 bytes exceeds … |
| arbitrary_500mb | readFileStream | websocket | 667 MB | 5.04 s | 132 MB/s | ✓ |  |
| arbitrary_500mb | readFile | websocket | — | — | — | ✗ | SandboxError: File too large. Size 524288000 bytes exceeds … |
</details>

### DIFF
--- a/packages/sandbox-container/src/services/file-service.ts
+++ b/packages/sandbox-container/src/services/file-service.ts
@@ -975,7 +975,7 @@ export class FileService implements FileSystemOperations {
       //    unknown extensions it returns 'application/octet-stream'.  In that
       //    case we run `file --mime-type` as a fallback so we can correctly
       //    classify extension-less binaries (e.g. compiled executables).
-      let mimeType = bunFile.type;
+      let mimeType = bunFile.type.split(';')[0].trim();
       if (mimeType === 'application/octet-stream') {
         const escapedPath = shellEscape(path);
         const mimeResult = await exec(`file --mime-type -b ${escapedPath}`);
@@ -985,7 +985,6 @@ export class FileService implements FileSystemOperations {
         // If the fallback fails we keep 'application/octet-stream', which
         // isBinaryMimeType() will correctly classify as binary.
       }
-      mimeType = mimeType.split(';')[0].trim(); // In case there are extra parameters like charset
 
       // 4. Classify binary vs text
       const isBinary = this.isBinaryMimeType(mimeType);


threepointone:
bun

tests/e2e/test-worker/startup-test.sh
```diff
@@ -8,7 +8,7 @@ MARKER_FILE="/tmp/startup-marker.txt"
 TIMESTAMP=$(date +%s)
 
 # Write marker file with timestamp to prove execution
-echo "startup-${TIMESTAMP}" > "${MARKER_FILE}"
+echo -n "startup-${TIMESTAMP}" > "${MARKER_FILE}"
```
scuffi, line 11: this was adding a newline after the 'startup-12345' text, so the regex lookup failed, I guess `\n` used to be removed from the end of the content when returned?

packages/sandbox-container/src/services/file-service.ts
```diff
@@ -97,77 +101,74 @@ export class FileService implements FileSystemOperations {
...
-              } satisfies FileNotFoundContext
-            };
-          }
+          let targetPath = path;
 
-          // Get file size using stat
-          const statCommand = `stat -c '%s' ${escapedPath} 2>/dev/null`;
-          const statResult = await exec(statCommand);
+          if (!path.startsWith('/')) {
+            const pwdResult = await exec('pwd');
```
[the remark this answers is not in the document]
whoiskatrin: Before this change, relative paths were passed to shell commands in-session, which already allowed ../, imo this is a false positive

packages/sandbox-container/src/services/file-service.ts
```diff
@@ -1001,16 +973,11 @@ export class FileService implements FileSystemOperations {
...
-      // 2. Check if file exists using session-aware check
-      const existsResult = await this.exists(path, sessionId);
-      if (!existsResult.success) {
-        return {
-          success: false,
-          error: existsResult.error
-        };
-      }
+      // 2. Use Bun.file() for existence and stat.
+      const bunFile = Bun.file(path);
```
whoiskatrin, line 955: Relative paths here are resolved against process cwd, not the session cwd. Can we resolve path via session pwd first (same as read()) and call Bun.file(resolvedPath) to keep session behaviour consistent?

packages/sandbox-container/src/services/file-service.ts
```diff
@@ -97,77 +101,74 @@ export class FileService implements FileSystemOperations {
...
-          const statResult = await exec(statCommand);
+          if (!path.startsWith('/')) {
+            const pwdResult = await exec('pwd');
+            if (pwdResult.exitCode !== 0) {
+              throw {
+                code: ErrorCode.FILESYSTEM_ERROR,
+                message: `Failed to resolve working directory for '${path}'`,
+                details: {
+                  path,
+                  operation: Operation.FILE_WRITE,
```
whoiskatrin: it should be FILE_READ?

packages/sandbox-container/src/services/file-service.ts
scuffi: yeah, the `path` here is passed in from [here](https://github.com/cloudflare/sandbox-sdk/blob/6db285dcb67ee6b76af360c0b837ed44033f210c/packages/sandbox-container/src/services/file-service.ts#L1412) which uses the session pwd to create the path. I wanted to keep this function independent from that logic so it can be more reused, I can move that session pwd logic into this function body if you think that'll be better though?

whoiskatrin: I’d move pwd resolution into getFileMetadata(). Current behaviour relies on an implicit caller contract, which is easy to break down the road

scuffi: updated

packages/sandbox-container/src/services/file-service.ts
```diff
@@ -21,6 +22,9 @@ export interface SecurityService {
   validatePath(path: string): { isValid: boolean; errors: string[] };
 }
 
+// Maximum file size for RPC transfers is 32 MiB to prevent performance issues. For larger files, clients should use streaming APIs.
+const MAX_RPC_FILE_SIZE = 32 * 1_048_576; // 32 MiB
```
ghostwriternr, line 27: What happens today if users try to read a file larger than this limit with the current implementation? Does it work or not work? If it does work, would this become a breaking change?

packages/sandbox-container/src/services/file-service.ts
```diff
@@ -1473,159 +1394,171 @@ export class FileService implements FileSystemOperations {
...
+  if (isBinary) {
+    // Encode bytes to base64 without line breaks.
+    // btoa only accepts latin1; convert via charCodeAt.
+    let binary = '';
+    for (let i = 0; i < slice.length; i++) {
+      binary += String.fromCharCode(slice[i]);
+    }
+    data = btoa(binary);
+  } else {
+    data = new TextDecoder().decode(slice);
```
ghostwriternr: Does this logic handle multi-byte characters that land at chunk boundaries properly? What happens if they get split across different chunks?

tests/e2e/streaming-operations-workflow.test.ts
```diff
@@ -182,7 +182,7 @@ describe('Streaming Operations Edge Cases', () => {
     const complete = events.find((e) => e.type === 'complete');
 
     expect(metadata).toBeDefined();
-    expect(metadata.mimeType).toBe('text/plain');
+    expect(metadata.mimeType).toBe('text/plain;charset=utf-8');
```
ghostwriternr: Very mildly worried, but not significantly about this one: is the inclusion of the charset going to break any customers that use string-matching to detect mime types from our API? Should we keep it consistent to older behaviour or mark this as a breaking change on our changeset?

packages/sandbox-container/src/services/file-service.ts
```diff
@@ -1473,159 +1394,171 @@ export class FileService implements FileSystemOperations {
...
-          );
-          controller.close();
-        } catch (error) {
-          const errorMessage =
-            error instanceof Error ? error.message : 'Unknown error';
-          this.logger.error(
-            'File streaming failed',
-            error instanceof Error ? error : undefined,
-            { path }
+        return fileStream.pipeThrough(sseTransform).pipeThrough(
```
ghostwriternr: Any reason for 2 layers of piping here?

packages/sandbox-container/src/services/file-service.ts
```diff
@@ -1473,159 +1394,171 @@ export class FileService implements FileSystemOperations {
...
+      absolutePath = `${cwd}/${path}`;
+    }
 
-            const execResult = await this.sessionManager.executeInSession(
-              sessionId,
-              command
+    const CHUNK_SIZE = 65535;
+
+    return await this.sessionManager
+      .withSession(sessionId, async () => {
```
ghostwriternr: Nit: considering we're not using `exec` that the `withSession` wrapper provides here, why even use the wrapper? Looks like the session-related work is complete even before we get to this point, so we could may be get rid of the wrapping.

packages/sandbox-container/src/services/file-service.ts
ghostwriternr: And if the [wrapping comment](https://github.com/cloudflare/sandbox-sdk/pull/402#discussion_r2845836996) makes sense, I think we can definitely eliminate this unwrapping here.

packages/sandbox-container/src/services/file-service.ts
scuffi: yes good spot, I previously was using the exec in here but moved it out, I'll remove this wrapper

packages/sandbox-container/src/services/file-service.ts
scuffi: We currently get a 'similar' error, however, the changes make us fail fast, so we don't even attempt to read the file if we know it'll trigger this RPC error. I just made it explicit as it was coming through as an unknown error so gave it a more accurate status code + message. See the benchmarks for the before/after error messages

packages/sandbox-container/src/services/file-service.ts
scuffi: The only reason for the second pipe is so we can return a normalised stream, as returning the Bun stream directly has some issues. I think we could get away with it, but I also think it's safer to return a standardised stream, wdyt?

tests/e2e/streaming-operations-workflow.test.ts
scuffi: I can clean this up easily for consistency & not introduce a breaking change?

packages/sandbox-container/src/services/file-service.ts
scuffi: I'll put more concretely in the changeset that we now explicitly do not support files >32MiB being transferred through read, and to use streaming APIs instead due to RPC constraints

packages/sandbox-container/src/services/file-service.ts
scuffi: Thanks for calling this out! I'll change it to use a stateful text decoder instead so we don't accidentally corrupt text files

packages/sandbox-container/src/services/file-service.ts
ghostwriternr: For my own understanding - what issues occur if we return the bun stream as-is?

packages/sandbox-container/src/services/file-service.ts
```diff
@@ -97,77 +101,74 @@ export class FileService implements FileSystemOperations {
...
-              code: ErrorCode.FILESYSTEM_ERROR,
-              message: `Failed to get file size for '${path}'`,
-              details: {
-                path,
-                operation: Operation.FILE_READ,
-                stderr: statResult.stderr
-              } satisfies FileSystemContext
-            };
+            const cwd = pwdResult.stdout.trim();
+            targetPath = `${cwd}/${path}`;
```
ghostwriternr: Minor: Similar to #404 , we should be using `resolve` so even paths like `./foo` and `../bar` would work

packages/sandbox-container/src/services/file-service.ts
```diff
@@ -1473,159 +1395,185 @@ export class FileService implements FileSystemOperations {
...
-                type: 'error',
-                error: `Failed to read chunk at offset ${bytesRead}: ${execResult.data.stderr}`
-              };
-              controller.enqueue(
-                encoder.encode(`data: ${JSON.stringify(errorEvent)}\n\n`)
-              );
-              controller.close();
-              return;
-            }
+    const fileStream = Bun.file(absolutePath).stream();
```
ghostwriternr: Just thinking out loud: Now that `withSession` is removed, the entire file stream runs outside any session lock. The pwd resolution happened earlier via `executeInSession`, but between that and the actual read here, another session operation _could_ technically `cd` elsewhere or modify/delete the file. Avoiding such race conditions so that users get a consistent bash-like behaviour that only allows one foreground command at a time is the reason for having `withSession` — and #404 intentionally keeps its work inside `withSession` for this reason, which I like.

I think maybe we should keep it too?

packages/sandbox-container/src/services/file-service.ts
```diff
@@ -177,52 +178,26 @@ export class FileService implements FileSystemOperations {
...
-                  operation: Operation.FILE_READ,
-                  exitCode: base64Result.exitCode,
-                  stderr: base64Result.stderr
-                } satisfies FileSystemContext
-              };
+            const buffer = await bunFile.arrayBuffer();
+            const bytes = new Uint8Array(buffer);
+            // btoa only accepts latin1; convert byte-by-byte.
+            let binary = '';
+            for (let i = 0; i < bytes.length; i++) {
```
ghostwriternr: I noticed this loop in 2 places in your change - can we not just use `content = Buffer.from(buffer).toString('base64')`, which should functionally be the same but likely a lot more performant considering this loop byte-by-byte creates new copies of the string - which could be a _lot_ when reading larger files.

packages/sandbox-container/src/services/file-service.ts
scuffi: I'll need to rewrite the `getFileMetadata` method to take an 'exec' or 'pwd' as a property, as if we undid the change we would still have the issue where the session lock could be taken in between these calls. I'll rewrite the metadata method to not use the sessionManager itself

packages/sandbox-container/src/services/file-service.ts
scuffi: I was wrong - not sure where I read that but went through some docs with Gemini and I think I was thinking about something else. I've removed the second pipe 👍

packages/sandbox-container/src/services/file-service.ts
```diff
@@ -1473,159 +1342,203 @@ export class FileService implements FileSystemOperations {
...
+        if (!metadataResult.success) {
+          return new ReadableStream({
+            start(controller) {
               const errorEvent = {
                 type: 'error',
-                error: `Failed to read chunk at offset ${bytesRead}: Command execution failed`
+                error: metadataResult.error.message
               };
               controller.enqueue(
                 encoder.encode(`data: ${JSON.stringify(errorEvent)}\n\n`)
```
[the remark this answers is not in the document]
scuffi: we validate the path in the `readFileStreamOperation` method itself, but we also validate the absolutePath in the call to `getFileMetadata` after the absolute path is resolved, so we are validating both paths here

packages/sandbox-container/src/services/file-service.ts
```diff
@@ -1024,101 +969,25 @@ export class FileService implements FileSystemOperations {
...
+      if (mimeType === 'application/octet-stream') {
+        const escapedPath = shellEscape(path);
+        const mimeResult = await exec(`file --mime-type -b ${escapedPath}`);
+        if (mimeResult.exitCode === 0) {
+          mimeType = mimeResult.stdout.trim();
+        }
+        // If the fallback fails we keep 'application/octet-stream', which
+        // isBinaryMimeType() will correctly classify as binary.
       }
+      mimeType = mimeType.split(';')[0].trim(); // In case there are extra parameters like charset
```
ghostwriternr: Shouldn't this transformation be before the `if` block above - which is evaluating only for the prefix too?

ghostwriternr: Also, if this indeed should have been earlier, why did our tests not catch it?

PR #404, opened 2026-02-20
Use native file writes in FileService

whoiskatrin:
## Summary
This PR applies the same optimization idea as #402 to the write path.

`FileService.write()` now uses native `Bun.write()` instead of shell pipelines (`printf '%s' ... | base64 -d > ...`).

## Why
The previous write path spawned shell commands for every write. This added avoidable process overhead and complexity around shell-safe encoding/quoting.

Using native writes improves:
- write latency and throughput
- reliability for text/binary payloads
- security posture (less command construction)

## What changed
- **`packages/sandbox-container/src/services/file-service.ts`**
  - Replaced shell-based write implementation with `Bun.write()`
  - Preserved encoding behavior (`utf-8`, `utf8` alias, `base64`)
  - Preserved relative-path behavior by resolving against session `pwd`
  - Kept base64 input validation for `encoding: 'base64'`

- **`packages/sandbox-container/tests/services/file-service.test.ts`**
  - Updated assertions for native write behavior
  - Added relative-path resolution coverage
  - Added `vi.restoreAllMocks()` in `beforeEach` to avoid cross-suite mock leakage

- **`packages/sandbox-container/scripts/benchmark-write.ts`** (new)
  - Benchmarks legacy write path vs native write path
  - Supports `--tier=core|stress|all`, `--runs`, `--warmup`
  - Prints markdown summary for PRs

- **`packages/sandbox-container/package.json`**
  - Added `bench:write` script

- **`packages/sandbox-container/tests/services/process-store.test.ts`**
  - Isolated process-store test directory per test run (stability fix discovered while validating full suite)

## Benchmark
Command:
`npm run bench:write -w @repo/sandbox-container -- --tier=core --runs=3 --warmup=1`

### Before vs After (10MB / 100MB)
| Case | Size | Encoding | Before avg (ms) | After avg (ms) | Speedup | Before MB/s | After MB/s |
| :--- | ---: | :--- | ---: | ---: | ---: | ---: | ---: |
| utf8-10MB-abs | 10MB | utf-8 | 3550.39 | 2.16 | 1641.75x | 2.82 | 4624.16 |
| utf8-100MB-abs | 100MB | utf-8 | 35398.47 | 20.04 | 1766.08x | 2.82 | 4989.13 |
| base64-10MB-abs | 10MB | base64 | 3642.76 | 26.53 | 137.28x | 2.75 | 376.86 |
| base64-100MB-abs | 100MB | base64 | 36426.27 | 264.94 | 137.49x | 2.75 | 377.44 |

## Validation
- `npm run check` ✅
- Focused file/path tests pass:
  - `bun test tests/services/file-service.test.ts tests/handlers/file-handler.test.ts`

### DIFF
--- a/packages/sandbox-container/src/services/file-service.ts
+++ b/packages/sandbox-container/src/services/file-service.ts
@@ -1,3 +1,4 @@
+import { resolve } from 'node:path';
 import type { FileInfo, ListFilesOptions, Logger } from '@repo/shared';
 import { shellEscape } from '@repo/shared';
 import type {
@@ -342,7 +343,7 @@ export class FileService implements FileSystemOperations {
             }
 
             const cwd = pwdResult.stdout.trim();
-            targetPath = `${cwd}/${path}`;
+            targetPath = resolve(cwd, path);
           }
 
           try {
--- a/packages/sandbox-container/tests/handlers/file-handler.test.ts
+++ b/packages/sandbox-container/tests/handlers/file-handler.test.ts
@@ -164,7 +164,8 @@ describe('FileHandler', () => {
       const writeFileData = {
         path: '/tmp/output.txt',
         content: 'Hello, File!',
-        encoding: 'utf-8'
+        encoding: 'utf-8',
+        sessionId: 'session-123'
       };
 
       (mockFileService.writeFile as any).mockResolvedValue({
@@ -191,7 +192,37 @@ describe('FileHandler', () => {
         'Hello, File!',
         {
           encoding: 'utf-8'
-        }
+        },
+        'session-123'
+      );
+    });
+
+    it('should pass undefined sessionId when not provided', async () => {
+      const writeFileData = {
+        path: '/tmp/output.txt',
+        content: 'Hello, File!'
+      };
+
+      (mockFileService.writeFile as any).mockResolvedValue({
+        success: true
+      });
+
+      const request = new Request('http://localhost:3000/api/write', {
+        method: 'POST',
+        headers: { 'Content-Type': 'application/json' },
+        body: JSON.stringify(writeFileData)
+      });
+
+      const response = await fileHandler.handle(request, mockContext);
+
+      expect(response.status).toBe(200);
+      expect(mockFileService.writeFile).toHaveBeenCalledWith(
+        '/tmp/output.txt',
+        'Hello, File!',
+        {
+          encoding: undefined
+        },
+        undefined
       );
     });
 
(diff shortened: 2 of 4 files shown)


whoiskatrin:
We intentionally resolve relative paths inside withSession() so path resolution and write execute under the same session lock. This preserves cwd consistency when concurrent commands in the same session may run cd. Resolving pwd before withSession() could race and write to a stale directory.

packages/sandbox-container/src/services/file-service.ts
```diff
@@ -322,42 +319,56 @@ export class FileService implements FileSystemOperations {
...
+                  operation: Operation.FILE_WRITE,
+                  exitCode: pwdResult.exitCode,
+                  stderr: pwdResult.stderr
+                } satisfies FileSystemContext
+              };
+            }
 
-      const result = execResult.data;
+            const cwd = pwdResult.stdout.trim();
+            targetPath = `${cwd}/${path}`;
```
ghostwriternr: minor: should maybe use `path.join` so even stuff like `./file.txt` would normalize properly

ghostwriternr (changes requested):
I'd not have noticed this if not for a [single-line change](https://github.com/cloudflare/sandbox-sdk/pull/402/changes#diff-44b4f6a1cf9c201a75d026f425656506e3e9f6e4771191a148813c80d6007d94) from #402 that's missing here in `file-handler.ts`:
```ts
- const result = await this.fileService.writeFile(body.path, body.content, {
-   encoding: body.encoding
- });
+ const result = await this.fileService.writeFile(body.path, body.content, {
+   encoding: body.encoding
+ }, body.sessionId); // this line
```

Without this change, code like this would seem to work:
```ts
const session = await sandbox.createSession({ id: 'my-session', cwd: '/workspace' });
await session.exec('mkdir -p project && cd project');
await session.writeFile('file.txt', 'hello');
```
But the file actually lands in `/workspace`, not `/workspace/project` so:
```ts
await session.exec('cat /workspace/myproject/hello.txt);
```
would fail — also indicating our tests are insufficient around this and should be updated. Can we ensure `sessionId` is passed for writeFile too from file-handler so `pwd` actually gets the right directory in such cases? And updated tests

packages/sandbox-container/src/services/file-service.ts
whoiskatrin: Fixed

packages/sandbox-container/tests/handlers/file-handler.test.ts
```diff
@@ -191,7 +192,37 @@ describe('FileHandler', () => {
         'Hello, File!',
         {
           encoding: 'utf-8'
-        }
+        },
+        'session-123'
+      );
```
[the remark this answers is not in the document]
ghostwriternr: +1 on this but not blocking

ghostwriternr (approved):

PR #422, opened 2026-02-25
Add desktop environment container runtime

ghostwriternr:
## Summary

Sandboxes can now run full desktop environments (Xvfb + XFCE4 + x11vnc + websockify) with programmatic screenshot capture. This is the container-side runtime — a Go wrapper around robotgo for native screen capture, exposed to Bun via koffi FFI, with a service/handler layer wired into the existing DI container.

## Design decisions

**Go FFI for screenshots rather than a CLI tool like `scrot`**: robotgo gives direct access to the X11 framebuffer without spawning a process per capture. The Go wrapper compiles as a c-shared library and is called from a Bun worker thread via koffi, keeping the main event loop unblocked.

**`os.Setenv("DISPLAY", ":99")` in Go `init()`**: In c-shared mode, Go's `os.Getenv` doesn't reliably inherit the container's environment variables. The xgb library (used internally by robotgo) calls `os.Getenv("DISPLAY")` at connection time, so the display must be set from Go's own environment before any capture call.

**`koffi.disposable('HeapStr', 'str')` for C string returns**: koffi's default `'str'` return type copies the C string to a JS string but never frees the original C allocation. The disposable wrapper calls `free()` after conversion, preventing a memory leak on every screenshot.

**robotgo `Capture()` returns Go-GC-managed `*image.RGBA`**: Unlike the lower-level C bitmap API, this doesn't require manual `FreeBitmap` calls. The Go garbage collector handles cleanup.

**Resolution validated and clamped**: Width/height must be even numbers (X11 requirement), minimum 640×480, maximum 3840×2160. The desktop manager validates at start time rather than letting Xvfb fail with a cryptic error.

---

This is **part 1 of 4** in the desktop environment stack:
- <kbd>&nbsp;4&nbsp;</kbd> #425 — Add desktop viewer example
- <kbd>&nbsp;3&nbsp;</kbd> #424 — Add desktop environment tests
- <kbd>&nbsp;2&nbsp;</kbd> #423 — Add desktop environment SDK client
- <kbd>&nbsp;1&nbsp;</kbd> #422 — Add desktop environment container runtime 👈
- `main`


packages/sandbox-container/src/workers/desktop-worker.ts
```diff
@@ -0,0 +1,224 @@
...
+  }
+}
+
+self.onmessage = (event: MessageEvent) => {
+  const { id, op, ...args } = event.data;
+
+  try {
+    if (!lib && !loadLibrary()) {
+      self.postMessage({ id, error: 'Desktop library not available' });
+      return;
```
scuffi, line 144: nit: could throw here to be more consistent with the rest of the method body?

packages/sandbox-container/src/handlers/desktop-handler.ts
```diff
@@ -0,0 +1,265 @@
...
+
+  async handle(request: Request, context: RequestContext): Promise<Response> {
+    const url = new URL(request.url);
+    const pathname = url.pathname;
+
+    switch (pathname) {
+      case '/api/desktop/start':
+        return this.handleStart(request, context);
+      case '/api/desktop/stop':
+        return this.handleStop(request, context);
```
scuffi, line 35: just an idea for future, would we want a `/desktop/restart` handler too? Below we advise people to restart with `sandbox.desktop.stop()` and `sandbox.desktop.start()`, which we could handle in a `sandbox.desktop.restart()` call along with some failsafes, and more error handling for specifically restarting a desktop?

PR #430, opened 2026-02-26
Split mimeType in file reads

scuffi:
# Summary

Copies the mime type splitting logic from `getFileMetadata` into the `readFile` method for parity in their metadata responses.

# Testing

- Updates the mine type detection tests to account for charsets in text/plain mime types


ghostwriternr:
changeset pls

PR #434, opened 2026-02-27
Implement an example project using vite to run both host and sandboxed applications

aron-cf:
This PR introduces an example project demonstrating a Vite React application embedded in a sandbox hosted by a Vite React application. A "counter" script changes the sandbox App.jsx file to demonstrate hot module reloading (HMR).

https://github.com/user-attachments/assets/f1d8c4ce-9235-4b0f-89d9-865633387db4

## Setup

Start the development server:

```bash
cd examples/vite-sandbox
npm start
```

## Usage

This is a non-interactive demo. The counters in both host and iframed sandbox will increment/decrement once per second to demonstrate that the hot module reloading is working over websockets between browser and sandbox.

## Deploy

```bash
npm run deploy
```

## Implementation Notes

Hosting two Vite servers on the same port along with the Cloudflare wrangler server has the potential for unexpected behavior.

We refer to the current directory as the "host" server and the one loaded in the sandbox as the "sandbox" server. The Cloudflare services (workers, assets, storage etc.) are referred to as wrangler. Configuration for the host Vite server is in the root vite.config.js, the Cloudflare config is in wrangler.jsonc and the sandbox Vite config is in sandbox-app/vite.config.js.

This repository has been setup in a way to reduce the confusion.

 1. We assume static assets will be served by Cloudflare. The host Vite server has `appType` set to `"custom"` to disable Vite handling HTML.
 2. The hot module reloading server is configured under `server.hmr` has been set to run on a different port to the Vite dev server. This reduces the chance of conflicts between the host and sandbox HMR websockets.
 3. We pass the host port via the `VITE_CLIENT_PORT` environment variable so that the HMR server is configured correctly.

### DIFF
--- a/examples/vite-sandbox/README.md
+++ b/examples/vite-sandbox/README.md
@@ -55,7 +55,7 @@ Depending on your Vite configuration and application setup the above setup may s
     "assets": {
         "binding": "Assets",
         "not_found_handling": "none",
-        "run_worker_first": ["/*", "!/assets]
+        "run_worker_first": ["/*", "!/assets"]
     },
     ```
 
--- a/examples/vite-sandbox/sandbox-app/vite.config.js
+++ b/examples/vite-sandbox/sandbox-app/vite.config.js
@@ -25,7 +25,7 @@ export default defineConfig({
     host: process.env.VITE_HOST ?? '0.0.0.0',
     port: process.env.VITE_PORT ?? 5173,
     hmr: {
-      clientPort: process.env.VITE_HMR_CLIENT_PORT ?? 3000
+      clientPort: process.env.VITE_HMR_CLIENT_PORT
     }
   }
 });
--- a/examples/vite-sandbox/src/worker.js
+++ b/examples/vite-sandbox/src/worker.js
@@ -38,7 +38,8 @@ async function handleAPISandboxRoute(url, env) {
       env: {
         VITE_BASE: VITE_BASE,
         VITE_PORT: `${VITE_PORT}`,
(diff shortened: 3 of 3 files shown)


aron-cf:
I'm still debugging why this doesn't work. With a default setup we run into issues with the host dev server running with the @cloudflare/vite plugin intercepting the sandbox request for the `@vite/client` file.

e.g. for a dev server running on localhost:3000 with a sandbox server running on port 5173 you'd get a request to:

    http://5173-vite-sandbox-<token>.localhost:3000/@vite/client

This would return the configuration for the host dev server, and the subsequent websocket would be:

    ws://5173-vite-sandbox-<token>.localhost:3000/?token=fr4WLvivldEq
    
This would again connect to the host websocket rather than the sandbox one.

This can be worked around by using the `base` setting in the sandbox `vite.config.js` file to something like `/sandbox`. This then separates the two and the initial request becomes:

    http://5173-vite-sandbox-<token>.localhost:3000/sandbox/@vite/client

This has the drawback that the sandbox server _must_ be served from a prefixed path, but for now that's probably fine as a constraint. However for some reason this _still_ doesn't work :/ The websocket connection no longer connects to anything. So needs more investigation.

In the container I can use `websocat` to connect to the stream just fine:

    curl -L https://github.com/vi/websocat/releases/download/v1.14.1/websocat.aarch64-unknown-linux-musl > websocat
    chmod +x websocat
     ./websocat "ws://0.0.0.0:5173/sandbox/?token=fr4WLvivldEq" --protocol vite-hmr

PR #436, opened 2026-03-02
Add backup exclude options and defaults

whoiskatrin:
## Summary
This PR adds end-to-end support for excluding files from backup archives.

`createBackup()` now supports:
- `exclude: string[]` for custom mksquashfs exclude patterns
- `excludeDefaults: boolean` to include a default ignore set

Default excludes are:
`node_modules`, `.git`, `dist`, `build`, `.next`, `.turbo`, `.cache`

## Why
Backups currently include dependency and build directories by default,
which can make archives much larger than needed and slow down backup and
restore workflows.

## API and behavior
`BackupOptions.exclude` is now wired through SDK -> DO -> container backup
service. Patterns are passed to `mksquashfs -ef` (one pattern per line).

`excludeDefaults` defaults to `false` for backward-compatible behavior.
Validation was added for `exclude` and `excludeDefaults` inputs.

## Tests
Added two backup workflow E2E tests:
- restore excludes paths passed in `exclude`
- restore excludes default paths when `excludeDefaults: true`

Also ran local type checks through the pre-push hook.

## Reviewer notes
This is intentionally minimal and keeps the existing backup flow. The
only functional flow change is passing an optional exclude file to
`mksquashfs`.

### DIFF
--- a/.changeset/backup-exclude-patterns.md
+++ b/.changeset/backup-exclude-patterns.md
@@ -1,0 +1,7 @@
+---
+'@cloudflare/sandbox': patch
+---
+
+Add backup exclude support so `createBackup()` can skip files and directories.
+Use `exclude` for custom patterns and `excludeDefaults: true` to skip common
+build and dependency paths like `node_modules`, `.git`, and `dist`.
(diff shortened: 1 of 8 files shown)


whoiskatrin:
Closing in favor of #437, which uses branch  in this repository (no fork).

PR #447, opened 2026-03-03
Overhaul CI pipeline and eliminate E2E test flakiness

ghostwriternr:
CI was both slow and unreliable. Every PR ran the full pipeline regardless of what changed (~18-25 min wall clock), Docker images were rebuilt from scratch in both PR and release paths, and quality checks ran sequentially. E2E tests compounded this — 17 of 23 test files shared a single sandbox, leaking filesystem/port/process state between tests. The OpenCode readiness probe hit a no-op endpoint. File watch cleanup had race conditions. The SSE stream handler had a listener registration ordering bug. `retry: 1` papered over all of it.

This PR replaces the monolithic CI with a compositional architecture and fixes the structural causes of E2E flakiness, so that a green pipeline actually means something.

### Pipeline architecture

Three reusable workflows (`reusable-build`, `reusable-quality`, `reusable-e2e`) composed by both `pr.yml` and `release.yml`. Quality checks (lint+types, SDK tests, container tests) and E2E suites (HTTP, WebSocket, browser) run in parallel. `dorny/paths-filter` at the entry point skips entire workflow trees when changes don't affect them — a docs-only PR runs nothing; an SDK-only change skips Docker entirely.

**Docker Buildx Bake with a single HCL config.** 4 image variants built in one invocation with shared base layers and a single registry cache round-trip, replacing 4 separate `docker build` commands with independent cache pulls.

**`crane copy` for releases.** The old release rebuilt Docker images from scratch despite identical images already existing from E2E. Now we retag the tested image — no rebuild, no drift between what was tested and what ships.

**Unified cleanup with explicit container teardown.** Worker deletion does not delete containers. `cleanup.yml` now handles both PR-close and daily-sweep triggers, with explicit container deletion and per-step timeouts.

### E2E test isolation

Every test file now creates and destroys its own sandbox via `createTestSandbox()` / `cleanupTestSandbox()`, with correct `X-Sandbox-Type` header routing for typed variants (python, opencode, desktop, musl, standalone). `retry: 1` is removed — tests pass on first attempt or they fail.

### SDK and container fixes caught by the audit

**OpenCode readiness probe.** `/global/health` unconditionally returns `{ healthy: true }` regardless of init state. Changed to probe `/path`, which only succeeds after the binary is actually ready. 3 call sites in `opencode.ts`.

**Watch process lifecycle.** `watchDirectory()` had a race where the 4-hop cancel chain (SDK → DO → container → shell) was slower than a new watch starting. Added centralized `stopWatch()` with SIGTERM→SIGKILL escalation and stale watcher cleanup before spawning.

**SSE stream listener ordering.** The container's `handleStream` registered output listeners *after* replaying buffered logs, creating a window where output emitted between the buffer snapshot and listener registration was lost. This caused intermittent `waitForLog` timeouts on HTTP transport while WebSocket was unaffected (different streaming path). Fixed by registering listeners before replay. Also removed the now-dead `streamProcessLogs` service method the handler no longer calls.

**Test worker process lookup.** The `waitForLog`, `waitForPort`, and `waitForExit` routes in the E2E test worker used `listProcesses().find()` to locate a process by ID. Changed to `getProcess(id)` — a direct lookup that matches how real SDK users would use the API.

**Git clone test repo.** Replaced `github/gitignore` (~15MB) with `octocat/Spoon-Knife` (~1KB). The test validates clone mechanics, not repo size.

### What was removed

`pullrequest.yml`, `cleanup-stale.yml`, `pkg-pr-new.yml`, `sync-docs.yml`, `scripts/cleanup-test-deployment.sh` — replaced or dead code.

### Post-merge

E2E worker naming changes from `sandbox-e2e-test-worker-main` to `sandbox-e2e-test-worker-main-http` + `sandbox-e2e-test-worker-main-websocket`. The old worker needs manual cleanup.

---

<a href="https://app.devin.ai/review/cloudflare/sandbox-sdk/pull/447" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -97,11 +97,9 @@ npm run docker:rebuild     # Rebuild container image locally (includes clean bui
 
 **Note:** Docker images are automatically built and published by CI (`release.yml`):
 
-- Beta images on every main commit
-- Stable images when "Version Packages" PR is merged
-- Multi-arch builds (amd64, arm64) handled by CI
-
-**Critical:** Docker image version MUST match npm package version. This is enforced via `ARG SANDBOX_VERSION` in Dockerfile.
+- Images are built on every push to main (for E2E testing against CF registry)
+- Published to Docker Hub only when the "Version Packages" PR is merged (stable releases)
+  **Critical:** Docker image version MUST match npm package version. This is enforced via `ARG SANDBOX_VERSION` in Dockerfile.
 
 ### Development Server
 
@@ -161,8 +159,6 @@ npm run dev                # Start wrangler dev server (builds Docker on first r
 8. **Automated release** (no manual intervention):
    - Changesets action creates a "Version Packages" PR when changesets exist
    - Merging that PR triggers automated npm + Docker Hub publishing
-   - Beta releases published on every main commit
-   - Stable releases published when changesets are merged
 
 ## Testing Architecture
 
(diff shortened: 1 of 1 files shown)


whoiskatrin:
Love this!

PR #449, opened 2026-03-03
Fix transport retry budget exceeding before container startup completes

whoiskatrin:
## Summary

- Make the transport layer's 503 retry budget aware of configured container startup timeouts, so the client doesn't give up before the container finishes starting
- Previously hard-coded at 120s, the retry budget now automatically scales to `instanceGetTimeoutMS + portReadyTimeoutMS + 30s` (with a 120s floor)

## Problem

The SDK has two independent timeout layers that could fight each other:

1. **Container startup timeouts** (`instanceGetTimeoutMS` + `portReadyTimeoutMS`) — configurable up to 5min + 10min = 15min total
2. **Transport retry budget** (`BaseTransport.fetch()`) — hard-coded at 120s

When customers configure longer startup timeouts for large images (e.g., 6+ GB), the transport layer gives up at 120s even though the DO is still waiting for the container to start. The request fails with a 503, and the customer never benefits from the timeout they configured.

```
Customer sets:  instanceGetTimeoutMS = 300_000 (5 min)
                portReadyTimeoutMS  = 600_000 (10 min)

Transport gives up at 120s → customer sees failure
Container would have started at ~180s → never reached
```

## Changes

| File | Change |
|---|---|
| `transport/types.ts` | Add `retryTimeoutMs` to `TransportConfig` |
| `transport/base-transport.ts` | Use configurable retry budget (default 120s, backwards compatible) |
| `clients/types.ts` | Add `retryTimeoutMs` to `HttpClientOptions` |
| `clients/sandbox-client.ts` | Thread `retryTimeoutMs` to WebSocket transport |
| `clients/base-client.ts` | Thread `retryTimeoutMs` to transport creation |
| `sandbox.ts` | Compute retry budget from startup timeouts; recreate client when timeouts change |

## Reviewer notes

- The retry budget formula is `max(120_000, instanceGetTimeout + portReadyTimeout + 30_000)`. The 30s margin covers the maximum single backoff delay (capped at 30s in `BaseTransport`). The 120s floor preserves current behavior when timeouts are at defaults (30s + 90s = 120s → floor kicks in → 120s, same as before).
- `createSandboxClient()` is called from three places now: constructor (line 385), `blockConcurrencyWhile` when stored timeouts are loaded, and `setContainerTimeouts()`. The client + interpreter are recreated because `retryTimeoutMs` is set at transport construction time — there's no way to mutate it after the fact without adding mutable state to the transport.
- Recreating `SandboxClient` is cheap (just object allocation, no I/O or connections). The WebSocket transport, if used, would need to reconnect, but `setContainerTimeouts` is only called once per sandbox via `getSandbox()`.
- Both HTTP and WebSocket transports inherit from `BaseTransport`, so both benefit from this fix.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10856,6 +10856,7 @@
       "os": [
         "android"
       ],
+      "peer": true,
       "engines": {
         "node": ">= 12.0.0"
       },
@@ -10876,6 +10877,7 @@
       "os": [
         "darwin"
       ],
+      "peer": true,
       "engines": {
         "node": ">= 12.0.0"
       },
@@ -10896,6 +10898,7 @@
       "os": [
         "darwin"
       ],
+      "peer": true,
       "engines": {
         "node": ">= 12.0.0"
       },
@@ -10916,6 +10919,7 @@
       "os": [
         "freebsd"
       ],
+      "peer": true,
       "engines": {
         "node": ">= 12.0.0"
       },
@@ -10936,6 +10940,7 @@
       "os": [
         "linux"
       ],
+      "peer": true,
       "engines": {
         "node": ">= 12.0.0"
       },
@@ -10956,6 +10961,7 @@
       "os": [
         "linux"
       ],
+      "peer": true,
       "engines": {
         "node": ">= 12.0.0"
       },
@@ -10976,6 +10982,7 @@
       "os": [
         "linux"
       ],
+      "peer": true,
       "engines": {
         "node": ">= 12.0.0"
       },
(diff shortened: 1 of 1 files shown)


ghostwriternr:
/bonk `this.client` and `this.codeInterpreter` are recreated whenever container timeouts change, so `retryTimeoutMs` on the transport can be updated. While this works for now, it's fragile because any future component that snapshots `this.client` at construction time will be stuck with a stale reference unless the new code also goes via all the recreation points. Can we perhaps have a setter on the transport instead, so the retry timeout can be updated without recreating the client.

PR #450, opened 2026-03-04
Handle exec timeouts correctly

scuffi:
# Summary

We now propagate timeouts for `sandbox.exec` & `session.exec` calls correctly, they will now correctly raise a `CommandError` if they time out. By default, commands will only time out if you explicitly pass the `timeout` property into your call. 

We now offer the ability to define a session timeout (applies to all commands in that session):

```js
const session = await sandbox.createSession({ commandTimeoutMs: 5000 }) // 5s
await session.exec('sleep 10') // Will timeout after 5s
```
...and individual command timeouts:
```js
const session = await sandbox.createSession({ commandTimeoutMs: 5000 }) // 5s
await session.exec('sleep 10', { timeoutMs: 3000 } ) // Will timeout after 3s
```

# Fix

## Per-command timeout propagation

- **`session.ts`**: Added `timeoutMs` directly into the timeout handler in `exec()`

## Per-session timeout propagation

- **`packages/shared/src/types.ts`**: Added `commandTimeoutMs` to `SessionOptions`
- **`packages/sandbox/src/sandbox.ts`**: `createSession()` now forwards `options.commandTimeoutMs` to the container
- **`packages/sandbox/src/clients/utility-client.ts`**: Added `commandTimeoutMs` to `CreateSessionRequest`
- **`packages/sandbox-container/src/handlers/session-handler.ts`**: `handleCreate()` now parses `commandTimeoutMs` from the request body correctly passes it to the session manager
- **`packages/sandbox-container/src/services/session-manager.ts`**: Now correctly passes `timeoutMs` to `session.exec()`

# Testing

Added `tests/e2e/command-timeout.test.ts` with 5 test cases:

- `sandbox.exec` respects per-command `timeout`
- `session.exec` respects per-command `timeout`
- `session.exec` respects session-level `commandTimeoutMs`
- Per-command `timeout` takes precedence over session-level `commandTimeoutMs`
- Timed-out command cleanup: session deletion kills the underlying process

### DIFF
--- a/packages/sandbox-container/src/services/session-manager.ts
+++ b/packages/sandbox-container/src/services/session-manager.ts
@@ -15,6 +15,7 @@ import type {
 } from '@repo/shared/errors';
 import { ErrorCode } from '@repo/shared/errors';
 import { Mutex } from 'async-mutex';
+import { time } from 'console';
 import { CONFIG } from '../config';
 import {
   type ServiceError,
@@ -316,11 +317,9 @@ export class SessionManager {
 
         const result = await session.exec(
           command,
-          timeoutMs !== undefined
+          cwd || env || timeoutMs !== undefined
             ? { cwd, env, timeoutMs }
-            : cwd || env
-              ? { cwd, env }
-              : undefined
+            : undefined
         );
 
         return {


packages/sandbox/Dockerfile
```diff
@@ -218,6 +218,14 @@ ENTRYPOINT ["/container-server/sandbox"]
 # ============================================================================
 FROM golang:1.24-bookworm AS go-builder
 
+RUN mkdir -p /usr/local/share/ca-certificates
+RUN --mount=type=secret,id=wrangler_ca \
+    if [ -f /run/secrets/wrangler_ca ] && [ -s /run/secrets/wrangler_ca ]; then \
+        cp /run/secrets/wrangler_ca /usr/local/share/ca-certificates/wrangler-dev-ca.crt && \
+        apt-get update && apt-get install -y --no-install-recommends ca-certificates && \
+        update-ca-certificates; \
+    fi
```
scuffi, line 227: Unrelated fix to support Warp builds for new Dockerfile.desktop

.changeset/propagate-command-timeouts.md
```diff
@@ -0,0 +1,7 @@
+---
+'@cloudflare/sandbox': patch
+---
+
+Support per-command and per-session timeouts for exec
+
+Timeouts now propagate correctly through the full stack. Per-command `timeout` on `exec()` takes priority over session-level `commandTimeoutMs` set via `createSession()`, which takes priority over the container-level `COMMAND_TIMEOUT_MS` environment variable.
```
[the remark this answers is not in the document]
scuffi: yes

tests/e2e/command-timeout.test.ts
```diff
@@ -0,0 +1,228 @@
...
+
+    expect(execResponse.status).toBe(500);
+    const execData = (await execResponse.json()) as ErrorResponse;
+    expect(execData.error).toMatch(/timeout/i);
+
+    // Should have timed out in ~1s, not ~30s
+    expect(elapsed).toBeLessThan(10000);
+  }, 30000);
+
+  test('timed-out command continues running; session can be deleted while command runs', async () => {
```
deathbyknowledge, line 148: Do we prefer keeping commands running on timeout rather than killing pending processes?

packages/sandbox-container/src/services/session-manager.ts
```diff
@@ -316,7 +316,11 @@ export class SessionManager {
 
         const result = await session.exec(
           command,
-          cwd || env ? { cwd, env } : undefined
+          timeoutMs !== undefined
```
deathbyknowledge: nit: I think you can skip this conditional chaining since you're already handling undefined inside `session.exec`

tests/e2e/command-timeout.test.ts
scuffi: yeah I think this was a design choice, cc: @whoiskatrin

PR #460, opened 2026-03-06
add activity tracking to prevent container shutdown during operations

whoiskatrin:
## Problem

The \`@cloudflare/containers\` base class calls \`renewActivityTimeout()\` **once** before \`containerFetch()\`. For long-running operations (e.g. a matplotlib chart-generation command taking >6s), the \`sleepAfter\` alarm fires and calls \`onActivityExpired()\` → \`stop()\` → SIGTERM — killing the container mid-operation.

**Production symptom**: \`ReadableStream received over RPC disconnected prematurely\` and \`SessionDestroyedError\` during otherwise valid \`session.exec()\` calls.

This is an SDK-level fallback fix. The upstream \`@cloudflare/containers\` issue (alarm not being reset during active operations) remains open separately.

## Solution

An **in-memory \`activeOperations\` counter** on the \`Sandbox\` class that guards \`onActivityExpired()\` — refusing to shut down while any operation is in progress.

### Three primitives added

| Primitive | Purpose | Used for |
|---|---|---|
| \`withActivityTracking(fn)\` | Wraps async operations — increments counter before, decrements after (including on error) | 32 blocking public methods (exec, writeFile, readFile, setEnvVars, killAllProcesses, session wrappers, etc.) |
| \`trackStream(stream)\` | Returns a TransformStream wrapper that holds the counter open until the stream completes, cancels, or errors | 6 stream-returning methods (execStream, readFileStream, streamProcessLogs, watch, runCodeStream, execStreamWithSession) |
| \`onActivityExpired()\` guard | Checks \`activeOperations > 0\` and returns early (deferring shutdown) instead of calling \`super.onActivityExpired()\` | Automatic — fires on every alarm tick |

### Key design decisions

1. **Ephemeral state only** — \`activeOperations\`, \`operationStartTimes\`, and \`operationIdCounter\` are in-memory. If the DO is evicted and re-instantiated, counter resets to 0. This is correct because eviction kills all in-flight operations, so there's nothing to guard.

2. **30-minute safety valve** — Each operation is timestamped. If ALL tracked operations are older than 30 minutes, the guard yields and allows shutdown. This prevents leaked counters (e.g. from a stream that never closes) from keeping containers alive forever.

3. **\`keepAlive\` remains highest priority** — The \`keepAliveEnabled\` check comes first in \`onActivityExpired()\`, unchanged.

4. **No manual \`renewActivityTimeout()\` call** — After \`onActivityExpired()\` returns, the base class [automatically calls \`renewActivityTimeout()\`](https://github.com/cloudflare/containers/blob/main/packages/containers/src/container.ts) regardless. Our guard just needs to \`return\` without calling \`super.onActivityExpired()\`.

5. **\`startProcessCallbackStream\` is self-tracking** — This method is fire-and-forget (called with \`.catch(() => {})\`), so it wraps itself with \`withActivityTracking\` rather than relying on the caller.

6. **Double-counting is accepted** — \`waitForLogPattern\` internally calls \`streamProcessLogs\` (a tracked stream method). Both increment the counter. This is harmless — the counter just reaches 2 instead of 1, and both decrements happen correctly.

7. **\`execWithSession\` is the root wrap point for session paths** — \`session.exec()\` and all backup/restore internal calls route through the private \`execWithSession()\` method. Wrapping \`execWithSession\` itself (rather than each caller) ensures all these paths are covered without duplication.

## Five problem patterns covered

| # | Pattern | Example methods | Fix |
|---|---|---|---|
| 1 | Blocking \`containerFetch\` | \`exec\`, \`writeFile\`, \`readFile\`, \`setEnvVars\`, \`killAllProcesses\` | \`withActivityTracking()\` |
| 2 | In-DO streaming loops | \`startProcessCallbackStream\`, \`waitForPortReady\`, \`waitForLogPattern\`, \`waitForProcessExit\` | \`withActivityTracking()\` |
| 3 | Streams returned via RPC | \`execStream\`, \`readFileStream\`, \`streamProcessLogs\`, \`watch\`, \`runCodeStream\` | \`trackStream()\` |
| 4 | WebSocket transport | Data after initial WS connection bypasses \`containerFetch\` entirely | Covered by patterns 1-3 wrapping the SDK methods |
| 5 | Session wrapper & internal paths | \`session.exec()\`, \`session.listFiles()\`, \`session.setEnvVars()\`, \`session.runCode()\`, backup/restore | \`execWithSession\` wrapped at root; session wrapper methods delegate to tracked public methods |

## Files changed

| File | Change |
|---|---|
| \`packages/sandbox/src/sandbox.ts\` | Added ~85 lines of infrastructure (properties, \`onActivityExpired\` guard, \`withActivityTracking\`, \`trackStream\`, \`setEnvVarsWithSession\`). Wrapped 32 methods with \`withActivityTracking\` and 6 with \`trackStream\` throughout the file. |
| \`packages/sandbox/tests/activity-guard.test.ts\` | **New file** — 29 unit tests covering guard behavior, counter behavior, stream tracking, pattern wiring, session wrapper paths, and \`killAllProcesses\`. |

### Recommended review order

1. **Start with the infrastructure** (bottom of \`sandbox.ts\`):
   - \`onActivityExpired()\` override — the guard logic
   - \`withActivityTracking()\` — the async wrapper
   - \`trackStream()\` — the TransformStream wrapper
   - \`setEnvVarsWithSession()\` — extracted helper for session \`setEnvVars\`
   - Properties near the top of the class

2. **Spot-check a few wrapped methods** — the pattern is identical everywhere:
   - Simple wrap: \`exec()\` — \`return this.withActivityTracking(async () => { ... })\`
   - Stream wrap: \`execStream()\` — \`const stream = await ...; return this.trackStream(stream)\`
   - Root wrap: \`execWithSession()\` — wraps the private method body, covering \`session.exec()\` and all backup/restore paths
   - Self-tracking: \`startProcessCallbackStream()\` — wraps itself since it's fire-and-forget

3. **Read the tests** (\`activity-guard.test.ts\`) — they document the expected behavior clearly

### What to look for

- ✅ Every public method that touches the container is wrapped (32 with \`withActivityTracking\`, 6 with \`trackStream\`)
- ✅ Session wrapper paths covered: \`session.exec()\`, \`session.listFiles()\`, \`session.setEnvVars()\`, code interpreter methods
- ✅ Counter always decrements — even on error paths (\`finally\` block in \`withActivityTracking\`, triple-path release in \`trackStream\`)
- ✅ No \`as any\`, \`@ts-ignore\`, or type suppressions
- ✅ No changes outside \`packages/sandbox/\`
- ✅ No persisted state — everything is in-memory

### What NOT to worry about

- The diff looks large (+1272/-570) but most of it is **indentation changes** from wrapping method bodies inside \`return this.withActivityTracking(async () => { ... })\`. The actual new logic is ~85 lines.
- Double-counting in \`waitForLogPattern\` → \`streamProcessLogs\` is intentional and harmless.

## Verification

- TypeScript: \`npm run typecheck\` ✅ (all packages clean, pre-existing \`sandbox-desktop-example\` missing-dep error unrelated to this PR)
- Tests: \`npm test\` ✅ (518 passed = 489 original + 29 new activity-guard tests)
- Build: \`npm run build\` ✅
- No regressions in existing test suite

---

<a href="https://app.devin.ai/review/cloudflare/sandbox-sdk/pull/460" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/packages/sandbox/src/sandbox.ts
+++ b/packages/sandbox/src/sandbox.ts
@@ -1372,9 +1372,11 @@ export class Sandbox<Env = unknown> extends Container<Env> implements ISandbox {
 
     if (this.activeOperations > 0) {
       const now = Date.now();
-      const allExpired = [...this.operationStartTimes.values()].every(
-        (startTime) => now - startTime > Sandbox.MAX_OPERATION_DURATION_MS
-      );
+      const allExpired =
+        this.operationStartTimes.size > 0 &&
+        [...this.operationStartTimes.values()].every(
+          (startTime) => now - startTime > Sandbox.MAX_OPERATION_DURATION_MS
+        );
 
       if (allExpired) {
         this.logger.warn(


ghostwriternr:
As discussed, closing this PR in favour of getting this fixed properly at the runtime level, considering this approach (and anything similar we come up with) is going to be fragile in its implementation, as well as introduce a great amount of tech debt and rigour needed to maintain the ever-growing codebase.

I've filed an issue at https://github.com/cloudflare/containers/issues/162 for now, and tracking with the team internally to chase a lower-level fix.

PR #468, opened 2026-03-06
Support local R2 bucket mounting

scuffi:
## Problem

Bucket mounting required a live S3-compatible endpoint, making it unusable during local development with `wrangler dev` where R2 buckets are available as DO bindings, not HTTP endpoints.

## Solution

Added `localBucket: boolean` into `options` passed to `mountBucket()`, the Sandbox DO resolves the R2 binding from its own `env` and starts a `LocalMountSyncManager`, a bidirectional sync loop between the R2 bucket and the container directory instead of launching `s3fs` FUSE

### How it works

| Direction | Mechanism |
|---|---|
| R2 -> Container | Polls `bucket.list()` every 1s, diffs etags against an in-memory snapshot, transfers only changed objects |
| Container -> R2 | Subscribes to the existing `watch()` SSE stream to detect writes/deletes, pushes changes to R2 via binding |

## Files changed

| File | Change |
|---|---|
| `packages/sandbox/src/local-mount-sync.ts`| `LocalMountSyncManager` class, handles all the sync logic |
| `packages/sandbox/tests/local-mount-sync.test.ts`| New unit tests |
| `packages/sandbox/src/sandbox.ts` | `mountBucket()` / `unmountBucket()` / `destroy()` / `onStop()` branch on `mountType` |
| `packages/sandbox/src/storage-mount/types.ts` | `MountInfo` -> now a union `FuseMountInfo \| LocalSyncMountInfo` |
| `packages/shared/src/types.ts` | `endpoint` made optional; `localBucket?: boolean` added to `MountBucketOptions` |

# Usage

To use this local setup, create sandbox project that has an R2 bucket configured in the `wrangler.jsonc` such as:
```json
"r2_buckets": [
  {
    "binding": "MY_BUCKET",
    "bucket_name": "my-test-bucket"
  }
]
```
then, mount your bucket like you would expect, but pass the `localBucket: true` prop into the `options` field as so:
```js
await sandbox.mountBucket('MY_BUCKET', '/workspace/mnt/r2', {
        localBucket: true,
      });
```

Now, you can interact with your container & bucket like you would in production. 

_Note: A better, more detailed guide will be added to the documentation after this PR is merged_

---

<a href="https://app.devin.ai/review/cloudflare/sandbox-sdk/pull/468" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/packages/sandbox/src/sandbox.ts
+++ b/packages/sandbox/src/sandbox.ts
@@ -672,12 +672,16 @@ export class Sandbox<Env = unknown> extends Container<Env> implements ISandbox {
   ): Promise<void> {
     this.logger.info(`Mounting bucket ${bucket} to ${mountPath}`);
 
-    if (options.localBucket) {
+    if ('localBucket' in options && options.localBucket) {
       await this.mountBucketLocal(bucket, mountPath, options);
       return;
     }
 
-    await this.mountBucketFuse(bucket, mountPath, options);
+    await this.mountBucketFuse(
+      bucket,
+      mountPath,
+      options as RemoteMountBucketOptions
+    );
   }
 
   /**


whoiskatrin:
I love this! Let's get a really nice and clear guide in the docs for it as well

packages/sandbox/src/sandbox.ts
```diff
@@ -884,7 +968,7 @@ export class Sandbox<Env = unknown> extends Container<Env> implements ISandbox {
     }
 
     // Add endpoint URL
-    s3fsArgs.push(`url=${options.endpoint}`);
+    s3fsArgs.push(`url=${options.endpoint!}`);
```
aron-cf: If this has been validated elsewhere and we know this will always be passed to this method, let's fix the method interface, rather than forcing the types.

```
options: Omit<MountBucketOptions, "endpoint"> & {endpoint: string},
```

packages/sandbox/src/local-mount-sync.ts
```diff
@@ -0,0 +1,437 @@
...
+    for (const [key] of this.snapshot) {
+      if (!newSnapshot.has(key)) {
+        const containerPath = this.r2KeyToContainerPath(key);
+        this.suppressEcho(containerPath);
+
+        try {
+          await this.client.files.deleteFile(containerPath, this.sessionId);
+          this.logger.debug('R2 -> Container: deleted file', { key });
+        } catch {
+          // File may already be gone
```
aron-cf: Is there anything more specific we can check here to identify if it's gone vs something broke?

packages/sandbox/src/local-mount-sync.ts
```diff
@@ -0,0 +1,437 @@
...
+      await Promise.all(
+        batch.map(async ({ key, action }) => {
+          const containerPath = this.r2KeyToContainerPath(key);
+          await this.ensureParentDir(containerPath);
+          this.suppressEcho(containerPath);
+          await this.transferR2ObjectToContainer(key, containerPath);
+          this.logger.debug('R2 -> Container: synced object', { key, action });
+        })
+      );
+    }
```
aron-cf, line 211: This block could probably do with some error handling. Currently if any of these actions fail the entire sync will be left in a partial state.

packages/shared/src/types.ts
```diff
@@ -1084,9 +1084,9 @@ export interface MountBucketOptions {
    * - AWS S3: 'https://s3.us-west-2.amazonaws.com'
    * - GCS: 'https://storage.googleapis.com'
    *
-   * Required field
+   * Required when `localBucket` is false
    */
-  endpoint: string;
+  endpoint?: string;
```
aron-cf: This is probably worth implementing as a typed union, that way you ensure the different configs are correct.

```ts
LocalMountBucketOptions = { localBucket: true, endpoint: never, ... }
FuseMountBucketOptions = { localBucket: false, endpoint: string, ... }
MountBucketOptions = LocalMountBucketOptions | FuseMountBucketOptions
```

packages/sandbox/src/local-mount-sync.ts
scuffi: I'll add an error log here, however, if something goes wrong here from the Sandbox, we don't have any options on the client side to handle this? As this is also for local development the user will have access to the container to debug so not sure if there's much more we can/should do over an error log?

PR #469, opened 2026-03-06
Add shell option to PTY

maschwenk:
## Summary
- Adds a `shell` option to `PtyOptions` so callers can specify which shell to spawn (e.g. `zsh`, `sh`, `fish`) instead of always using `bash`
- Threads the option through the client proxy (query param), WS upgrade handler, and `Pty.initialize()`
- Defaults to `bash` when not specified

### DIFF
--- a/.changeset/many-deers-dance.md
+++ b/.changeset/many-deers-dance.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/sandbox": patch
+---
+
+Add `shell` option to `PtyOptions` to allow specifying which shell to spawn (e.g. `zsh`, `sh`, `fish`). Defaults to `bash` when not specified.


packages/sandbox-container/src/pty.ts
```diff
@@ -44,7 +44,9 @@ export class Pty {
       }
     });
 
-    this.process = Bun.spawn(['bash'], {
+    const shell = options.shell ?? 'bash';
+
+    this.process = Bun.spawn([shell], {
```
maschwenk, line 49: Another option should be reading from `this.env.SHELL` or `process.env.SHELL`? idk

ghostwriternr (dismissed):
nice and clean : )

PR #470, opened 2026-03-06
Pass session env vars to PTY on creation

maschwenk:
## Summary
- Fixes `setEnvVars()` not being reflected in terminal sessions
- `getPty()` now captures the session shell's environment via `env -0` before spawning the PTY process
- Env vars like `ANTHROPIC_API_KEY`, `SHELL`, `PATH` set through `setEnvVars()` are now inherited by the terminal

## Context
Currently, calling `sandbox.setEnvVars({ ANTHROPIC_API_KEY: '...' })` before `sandbox.terminal(request)` has no effect on the PTY — the Pty constructor receives no `env` and the spawned process only inherits `process.env` from the container.

---

<a href="https://app.devin.ai/review/cloudflare/sandbox-sdk/pull/470" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/packages/sandbox-container/src/services/session-manager.ts
+++ b/packages/sandbox-container/src/services/session-manager.ts
@@ -755,9 +755,11 @@ export class SessionManager {
         return { success: true, data: session.pty };
       }
 
-      // Capture the session shell's current environment so the PTY
-      // inherits env vars set via setEnvVars().
-      let sessionEnv: Record<string, string> = {};
+      // Capture the session shell's current environment and working
+      // directory so the PTY inherits env vars set via setEnvVars()
+      // and reflects any directory changes made in the session.
+      const sessionEnv: Record<string, string> = {};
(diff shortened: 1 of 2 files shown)


packages/sandbox-container/src/services/session-manager.ts
```diff
@@ -755,8 +755,22 @@ export class SessionManager {
         return { success: true, data: session.pty };
       }
 
+      // Capture the session shell's current environment so the PTY
+      // inherits env vars set via setEnvVars().
+      const envResult = await session.exec('env -0');
```
maschwenk: this seems incredibly sus; open to other ideas

maschwenk: ~hang on this might not even be an issue. now I'm not able to reproduce. confusing~ nah this is definitely an issue

PR #474, opened 2026-03-09
feat(backup): add gitignore and excludes options to createBackup

whoiskatrin:
## Summary

Add `gitignore` and `excludes` options to `createBackup()` for controlling which files are included in backup archives.

### Changes
- Add `gitignore` option (default: `false`) to `BackupOptions` — excludes gitignored files when the directory is inside a git repo
- Add `excludes: string[]` option for explicit glob patterns to exclude from backups
- Soft failure when `gitignore: true` but git is not installed — logs a warning and proceeds without git-based exclusions
- Use `git ls-files` to resolve ignored paths before creating archives
- Fix handling of special characters in gitignored filenames (spaces, unicode)

### Testing
- Unit tests verify git detection, soft failure, user-provided excludes, special character handling, and mksquashfs integration
- E2E tests cover default behavior, opt-out flag, nested gitignores, and non-git directories

### API Contract
- Default behavior: all files are included (no breaking change)
- `gitignore: true`: gitignored files are excluded; gracefully degrades if git is unavailable
- `excludes: ["node_modules", "*.log"]`: explicit glob patterns passed to mksquashfs

### DIFF
--- a/packages/sandbox-container/src/services/backup-service.ts
+++ b/packages/sandbox-container/src/services/backup-service.ts
@@ -51,7 +51,7 @@ function isSafeAbsolutePath(path: string): boolean {
  */
 function validateBackupPaths(dir: string, archivePath: string): string | null {
   if (!isSafeAbsolutePath(dir)) {
-    return `Backup directory must be an absolute path under /workspace or /tmp: ${dir}`;
+    return `Backup directory must be a safe absolute path (no '..' or null bytes) under an allowed prefix (${ALLOWED_PREFIXES.join(', ')}): ${dir}`;
   }
   // Allowlist check: dir must start with one of the allowed prefixes
   const isAllowed = ALLOWED_PREFIXES.some(


packages/sandbox-container/src/handlers/backup-handler.ts
```diff
@@ -115,12 +115,27 @@ export class BackupHandler extends BaseHandler<Request, Response> {
...
+    }
+
     const sessionId = body.sessionId ?? context.sessionId ?? 'default';
 
     const result = await this.backupService.createArchive(
       body.dir,
       body.archivePath,
-      sessionId
+      sessionId,
+      body.useGitignore ?? true
```
aron-cf: Do we want to default to `true`? Feels like this could trigger some unexpected behaviour. I'd imagined primarily using this for speeding up the restoration of an existing session. If I've pulled a project from github. If I've done all the project bootstrap of installing modules and building dependencies (likely included in the .gitignore file) then I'd be surprised if on restoration these files were missing.

packages/shared/src/types.ts
```diff
@@ -1026,6 +1026,13 @@ export interface BackupOptions {
   name?: string;
   /** Seconds until automatic garbage collection. Default: 259200 (3 days). No upper limit. */
   ttl?: number;
+  /**
+   * Respect git ignore rules for the backup directory when it is inside a git repository.
+   *
+   * Default: true.
+   * If the directory is not inside a git repository, no git-based exclusions are applied.
+   */
+  useGitignore?: boolean;
```
aron-cf: ~What do you think to making this an array of potential ignore files and calling it `ignore` or `ignoreFiles`?~

~e.g.~

```
ignoreFiles: [".gitignore", ".npmignore", ".dockerignore"]
```

~This would allow the user to also specific a generic `.ignore` file with application specific excludes without having to rely on git. No strong opinion here, just a thought as this would also make the feature not tied to the git implementation of the ignore file (e.g. is the expectation that we also support nested `.gitignore` files like git does?)~

EDIT: Nevermind, I've just read through the implemenation and see that we actually run git to resolve the ignores, so that should handle everything but ties the implementation to git.

packages/sandbox-container/src/services/backup-service.ts
```diff
@@ -228,6 +277,65 @@ export class BackupService {
...
+    dir: string,
+    sessionId: string,
+    opLogger: Logger
+  ): Promise<string[]> {
+    const gitAvailableResult = await this.sessionManager.executeInSession(
+      sessionId,
+      'command -v git >/dev/null 2>&1'
+    );
+    if (!gitAvailableResult.success || gitAvailableResult.data.exitCode !== 0) {
+      opLogger.warn('Git is unavailable; skipping gitignore exclusions', {
```
aron-cf: This should probably be an error or can we ensure that git is available if we're going to depend on it for SDK functionality?

aron-cf (approved):
👍 I think it would be interesting to know how many real-world projects end up adopting this.

PR #534, opened 2026-03-27
Add new OpenCode x Sandboxes demo with remote tool calls

aron-cf:
Adds an experimental example showing OpenCode running inside a Cloudflare Sandbox, with all tool calls (bash, read, write, edit, grep, glob, list) routed through an isolated CodeSandbox rather than the container's own filesystem.

The demo aims to demonstrate how workers can provide a `capnproto` RPC API very easily to services (either internally or externally) that allow easy interaction with those primatives.

```ts
class SandboxRpcApi extends RpcTarget {
  #env: Env;

  constructor(env: Env) {
    super();
    this.#env = env;
  }

  async sandbox(sessionId: string) {
    return getSandbox(this.#env.SANDBOX, `session-${sessionId}`);
  }
}
```

Then consumed in the OpenCode plugin:

```ts
const api = await getSandbox();
    const result = await api
      .sandbox(context.sessionID)
      .exec(args.command, { timeout: args.timeout });
```

Also includes a small addition to `@cloudflare/sandbox`'s OpenCode integration: exposes `process` on `OpencodeServer` and adds an `onStart` callback to `OpencodeOptions`, invoked only when the server is freshly started (not reused). The example uses this to stream OpenCode logs to the Worker's console.

## Architecture

Two containers back the Worker:

- **OpenCodeSandbox** — runs `opencode serve` with a custom plugin that opens a Cap'n Web WebSocket server on port 3001
- **CodeSandbox** — one instance per OpenCode session; all tool calls are routed here via RPC

The Durable Object connects to the plugin over Cap'n Web and exposes a `SandboxRpcApi` that maps each tool call to the correct per-session `CodeSandbox`.

## Testing

```bash
cd examples/experimental/opencode
npm run dev
opencode attach http://localhost:8787
```


aron-cf:
Closing this for now, will revisit after #538

PR #539, opened 2026-03-30
Remove HTTP/WebSocket transports in favor of capnweb

aron-cf:
> [!IMPORTANT]
> DO NOT MERGE!

Make capnweb the sole transport between the Sandbox DO and container.
Remove the old SandboxClient, all HTTP domain clients, the transport
abstraction layer, HTTP handlers, router, route setup, middleware,
and the WebSocket adapter.

The RPCSandboxClient (backed by ContainerConnection) is now the only
client implementation. The container serves capnweb RPC at /capnweb
and PTY WebSocket at /ws/pty, with a minimal /health endpoint.

SANDBOX_TRANSPORT env var is no longer read — capnweb is always used.


ghostwriternr:
(I know this is wip, but I'm VERY EXCITED for this)

PR #567, opened 2026-04-08
Support local R2 backup/restore flow

scuffi:
# Summary

Add local-development support for backup and restore using an R2 binding instead of presigned URLs and FUSE-based mounts.

This keeps the backup archive format the same as production, but uses a simpler local flow that works in `wrangler dev`:

- `createBackup({ localBucket: true })` now writes the archive to the Worker’s `BACKUP_BUCKET` R2 binding
- `restoreBackup()` can round-trip that handle and restore locally by downloading from the binding and extracting with `unsquashfs`
- backup operations still go through the existing serialisation path to avoid overlapping create/restore work on the same sandbox

# What changed

- Added `localBucket?: boolean` to backup types so local backup handles can round-trip through `DirectoryBackup`
- Added a local backup creation path in `sandbox.ts` that:
  - creates the squashfs archive in the container
  - reads it back through the existing file APIs
  - uploads it directly to the `BACKUP_BUCKET` binding
- Added a local restore path in `sandbox.ts` that:
  - reads backup metadata from the binding
  - downloads the archive through the Worker binding
  - writes it into the container
  - extracts it with `unsquashfs`
- Added unit coverage for local backup creation, restore, validation, cleanup, expiry handling, and round-trip behavior

# Why

Production backup/restore relies on presigned URLs and FUSE-based restore behavior, which is not supported in local development.

This change provides a local path that:

- works with the existing R2 binding model in local dev
- avoids requiring presigned URL credentials
- avoids depending on FUSE for restore
- preserves the same stored archive format so local backups use the same on-bucket structure as production

---

<a href="https://app.devin.ai/review/cloudflare/sandbox-sdk/pull/567" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


packages/sandbox/src/sandbox.ts
```diff
@@ -4087,10 +4091,211 @@ export class Sandbox<Env = unknown> extends Container<Env> implements ISandbox {
...
+  /**
+   * Local-dev implementation of createBackup.
+   * Uses the R2 binding directly instead of presigned URLs.
+   * Archive format is identical to production (squashfs + meta.json).
+   */
+  private async doCreateBackupLocal(
+    options: BackupOptions
+  ): Promise<DirectoryBackup> {
+    const DEFAULT_TTL_SECONDS = 259200; // 3 days
+    const MAX_NAME_LENGTH = 256;
```
aron-cf: This shoudl use the same constants as the non-local backup.

whoiskatrin (dismissed):

whoiskatrin (dismissed):

aron-cf:
@scuffi is this still in need of review?

scuffi:
> @scuffi is this still in need of review?

Yep, just blocked from CI currently

whoiskatrin (approved):

PR #570, opened 2026-04-09
Fix unmount implementation

aron-cf:
`Sandbox.unmountBucket()` reports success even when the mount is still active. The `fusermount` command fails silently because the Ubuntu container image is missing `/etc/mtab`, which `fusermount` needs to locate active mounts. On top of that, the SDK never checks the exit code from `fusermount -u` — it unconditionally marks the mount as removed and deletes it from internal tracking. The mount point directory created during `mountBucket()` is also left behind after unmount.

The Dockerfile now creates the standard `/etc/mtab` symlink to `/proc/mounts` in the Ubuntu stage. The Alpine stage already has this symlink. In `unmountBucket()`, the `fusermount -u` exit code is now checked and a `BucketUnmountError` is thrown on failure, so tracking state only updates after a real unmount. After a successful unmount the mount directory is cleaned up via `mountpoint -q` and `rmdir`, with a warning logged if removal fails. `BucketUnmountError` is exported from `@cloudflare/sandbox` so callers can catch unmount failures specifically.

To verify, deploy a worker that mounts an R2 bucket with `mountBucket()`, then calls `unmountBucket()`. Before this change, `unmountBucket()` would return successfully while the mount remained active. After this change it either completes a real unmount and removes the directory, or throws `BucketUnmountError` with the `fusermount` stderr output.

The existing bucket mounting end-to-end test previously never called `unmountBucket()`. It now unmounts after file operations and verifies both that the mount point is no longer active and that the directory has been removed.

> [!NOTE]
> This adds a new `BUCKET_UNMOUNT_ERROR` error code and `BucketUnmountError` class to the public API. Callers currently catching `BucketMountError` will also catch this since `BucketUnmountError` extends it.

---

<a href="https://app.devin.ai/review/cloudflare/sandbox-sdk/pull/570" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


whoiskatrin (dismissed):

packages/sandbox/src/sandbox.ts
```diff
@@ -1146,6 +1155,28 @@ export class Sandbox<Env = unknown> extends Container<Env> implements ISandbox {
         }
       }
 
+      // Remove empty mount directory after FUSE unmount.
+      // Local-sync directories contain synced files so rmdir is not appropriate.
+      if (mountInfo.mountType !== 'local-sync') {
+        try {
+          const cleanup = await this.execInternal(
```
whoiskatrin: We free the `activeMounts` reservation before awaiting the directory cleanup. In the DO another request can interleave here, mount the same path again, and then have its directory removed by this old `rmdir`.
I think the path should stay reserved until teardown is fully complete, or the cleanup needs to happen before `activeMounts.delete(mountPath)` ?

whoiskatrin (approved):

packages/sandbox/src/sandbox.ts
aron-cf: Can you explain what you mean by teardown is fully complete?

The `mountpoint -q <path>` checks to see if a new mount has been established before executing the `rmdir`. The `rmdir` will fail if the directory has any contents e.g. some external request wrote an actual directory to that path. So this should be safe. Unless I'm still misunderstanding something.

packages/sandbox/src/sandbox.ts
```diff
@@ -1146,6 +1155,28 @@ export class Sandbox<Env = unknown> extends Container<Env> implements ISandbox {
         }
       }
 
+      // Remove empty mount directory after FUSE unmount.
+      // Local-sync directories contain synced files so rmdir is not appropriate.
+      if (mountInfo.mountType !== 'local-sync') {
```
aron-cf: I'll move this code up to line 1147. I think this got moved down here accidentally.

packages/sandbox/src/sandbox.ts
aron-cf: Sorry, I re-read your code and the message. Yes, you're right. I've moved the code up back into the block that does the FUSE unmount.

aron-cf: I've left the remove code _after_ setting `mountInfo.mounted = false` and `activeMounts.delete()`, I think it's better that those two sync calls are set immediately after the actual unmount. The `rmdir` cleanup then can run async and it doesn't matter if it fails due to another mount.

PR #573, opened 2026-04-10
Fix git clone timeout controls

whoiskatrin:
## Summary
- raise the default `gitCheckout()` clone timeout from 2 minutes to 10 minutes so larger or slower repositories stop failing on the current hard limit
- add an explicit `cloneTimeoutMs` override that is threaded through the SDK, transport layer, request types, container handler, and git service
- validate timeout input early and preserve sub-second precision when formatting the underlying `timeout` command so the API matches what the subprocess actually does

## Why
Users on `0.8.7` and `0.8.8` can hit a hardcoded 120 second timeout in the clone path behind `gitCheckout()`, even with `depth: 1`. This change fixes the immediate product issue and makes the timeout user-controlled without broadening the API beyond the existing checkout primitive.

## Testing
- `npm test -w @cloudflare/sandbox -- git-client.test.ts sandbox.test.ts`
- `bun test tests/handlers/git-handler.test.ts tests/services/git-service.test.ts tests/managers/git-manager.test.ts`
- `npm run typecheck:e2e`
- `git push` pre-push hook: `npm run typecheck`

## Reviewer Notes
- `cloneTimeoutMs` is intentionally scoped to the `git clone` subprocess. It does not try to represent total wall-clock time including cold start or container readiness retries, which keeps the API semantics explicit.
- The transport changes are there to support per-request request timeouts, especially in WebSocket mode, so clone requests can outlive the old 120 second client timeout without changing global request behavior.

---

<a href="https://app.devin.ai/review/cloudflare/sandbox-sdk/pull/573" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/packages/sandbox-container/src/services/git-service.ts
+++ b/packages/sandbox-container/src/services/git-service.ts
@@ -1,6 +1,7 @@
 // Git Operations Service
 
 import {
+  DEFAULT_GIT_CLONE_TIMEOUT_MS,
   type Logger,
   logCanonicalEvent,
   redactCommand,
@@ -13,7 +14,7 @@ import type {
 } from '@repo/shared/errors';
 import { ErrorCode } from '@repo/shared/errors';
 import type { CloneOptions, ServiceError, ServiceResult } from '../core/types';
-import { GIT_CLONE_TIMEOUT_SECONDS, GitManager } from '../managers/git-manager';
+import { GitManager, gitCloneTimeoutSeconds } from '../managers/git-manager';
 import type { SessionManager } from './session-manager';
 
 export interface SecurityService {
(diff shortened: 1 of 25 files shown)


packages/sandbox-container/src/services/git-service.ts
```diff
@@ -91,6 +92,25 @@ export class GitService {
       // Generate target directory if not provided
       const targetDirectory =
         options.targetDir || this.manager.generateTargetDirectory(repoUrl);
+      const cloneTimeoutMs = options.timeoutMs ?? DEFAULT_GIT_CLONE_TIMEOUT_MS;
+
+      if (!Number.isInteger(cloneTimeoutMs) || cloneTimeoutMs <= 0) {
+        errorMessage = `Invalid clone timeout '${options.timeoutMs}'. Must be a positive integer number of milliseconds.`;
```
aron-cf:
```suggestion
        errorMessage = `Invalid clone timeout '${options.timeoutMs}'. Must be a positive integer representing milliseconds.`;
```

packages/sandbox-container/src/services/git-service.ts
```diff
@@ -91,6 +92,25 @@ export class GitService {
...
+        return this.returnError({
+          message: errorMessage,
+          code: ErrorCode.VALIDATION_FAILED,
+          details: {
+            validationErrors: [
+              {
+                field: 'timeoutMs',
+                message:
+                  'Clone timeout must be a positive integer number of milliseconds',
```
aron-cf:
```suggestion
                  'Clone timeout must be a positive integer representing milliseconds',
```

aron-cf (approved):
Looks good 👍

PR #592, opened 2026-04-16
Expose container placement ID on the sandbox

whoiskatrin:
## Summary

Adds `sandbox.getContainerPlacementId()` so callers can detect when the underlying container has been replaced by the platform.

The placement ID (`CLOUDFLARE_PLACEMENT_ID`) identifies a specific container placement and changes whenever the platform replaces the container. Today, detecting replacement requires writing a sentinel file into the container and reading it back — which breaks the moment the container is unhealthy, exactly when you most need to know.

This change piggybacks on the existing session-create handshake: the container reports its placement ID in the response, the Durable Object caches it in storage, and subsequent reads return the cached value without contacting the container.

```ts
await sandbox.exec('true'); // ensures the handshake has run
const placementId = await sandbox.getContainerPlacementId();
if (typeof placementId === 'string' && placementId !== lastKnownPlacementId) {
  // Container was replaced — reconcile state
}
```

### Push on handshake, not pull on demand

The natural alternative was a dedicated `GET /api/placement` endpoint. Rejected because:

- Every read would require a healthy container — defeats the primary use case (detecting replacement when things go wrong).
- It duplicates a round-trip when the SDK already does a handshake on first use.

Instead, the placement ID travels inside `CreateSessionResponse`. The DO captures it in `ensureDefaultSession()` and writes it to `ctx.storage`. Reads go to DO storage directly — no container wake, no HTTP.

### Capture on both success and already-exists paths

`ensureDefaultSession()` has two outcomes: the session gets created, or the container reports `SESSION_ALREADY_EXISTS` (common after DO restart while the container lives on). The placement ID must be captured in **both** cases, otherwise a restarted DO would permanently lose the value until the container itself restarts.

- Success path: `CreateSessionResponse.placementId`
- Already-exists path: `SessionAlreadyExistsContext.placementId`, surfaced via `SessionAlreadyExistsError.placementId`. This flows through the standard `ServiceError.details → ErrorResponse.context` pipeline in `BaseHandler.enrichServiceError()`.

### Return type: `string | null | undefined`

Three distinct states, each meaningful:

| Value       | Meaning                                                                        |
| ----------- | ------------------------------------------------------------------------------ |
| `string`    | Handshake observed; placement ID captured.                                     |
| `null`      | Handshake observed; `CLOUDFLARE_PLACEMENT_ID` env var not set (e.g. local dev).|
| `undefined` | No handshake has occurred yet on this sandbox — call `exec()` or similar first.|

Collapsing `undefined` into `null` would hide a real distinction: "the env var is absent" vs "we have not yet looked."

### Overwrite on every handshake

A fresh placement is written on each session-create. Container replacement triggers a new handshake (the old session no longer exists in the new container), so the stored value is naturally kept current without any TTL or explicit invalidation.

## Changes

**Container (`@repo/sandbox-container`)**
- `handlers/session-handler.ts`: include `placementId: process.env.CLOUDFLARE_PLACEMENT_ID ?? null` in the success response; add `placementId` to the `details` of the `SESSION_ALREADY_EXISTS` error path.

**SDK (`@cloudflare/sandbox`)**
- `clients/utility-client.ts`: extend `CreateSessionResponse` with optional `placementId`.
- `errors/classes.ts`: `SessionAlreadyExistsError` gains a `placementId` accessor.
- `sandbox.ts`:
  - `ensureDefaultSession()` calls a new private `capturePlacementId()` on both the success and already-exists branches.
  - `capturePlacementId()` writes to `ctx.storage.put('placementId', ...)`, skipping writes when the field is absent from the response (forward-compat with older containers).
  - Public `getPlacementId()` reads from `ctx.storage.get('placementId')`.

**Shared (`@repo/shared`)**
- `errors/contexts.ts`: `SessionAlreadyExistsContext` gains optional `placementId`.
- `types.ts`: `getPlacementId()` added to the `ISandbox` interface.

## Reviewer notes

Main places worth a closer look:

- **`packages/sandbox/src/sandbox.ts`** — `ensureDefaultSession()` (~L1872) and `capturePlacementId()` (~L1924). Verify both branches capture, and that `undefined` is correctly treated as "no information" rather than a write-through.
- **`packages/sandbox-container/src/handlers/session-handler.ts`** — confirm the already-exists error path carries `placementId` in `details` (not just in the success response).
- **Return type semantics** — the `string | null | undefined` tri-state is deliberate. If a reviewer prefers `string | null` with a separate "is initialized" flag, happy to discuss, but callers lose information.
- **`.changeset/placement-id-api.md`** — user-facing description; please sanity-check the wording and the example.

Places that intentionally did **not** change:

- No `/api/placement` endpoint. We considered one and rejected it (see Design above).
- No `PlacementResponse` or `PlacementResult` type exported from the SDK — the wire DTO is internal to the session-create handshake.
- No public change to `CreateSessionResponse` shape beyond the new optional field — existing callers are unaffected.

## Tests

- **SDK unit (`packages/sandbox/tests/sandbox.test.ts`)** — new `describe('placement id capture', ...)` covers: stores a string, stores `null`, skips write when the field is omitted, `getPlacementId()` returns the stored value, `getPlacementId()` returns `undefined` pre-handshake.
- **Container unit (`packages/sandbox-container/tests/handlers/session-handler.test.ts`)** — new cases cover the success response, `null` when the env var is unset, the already-exists error path surfacing `placementId` in `details`, and that unrelated error codes are not polluted.
- **Results**: `@cloudflare/sandbox` 560/560 pass; `@repo/sandbox-container` 622/622 pass; all three target packages typecheck clean.

### DIFF
--- a/.changeset/placement-id-api.md
+++ b/.changeset/placement-id-api.md
@@ -1,0 +1,32 @@
+---
+'@cloudflare/sandbox': patch
+---
+
+Add `sandbox.getContainerPlacementId()` to retrieve the Cloudflare placement ID
+observed for the underlying container.
+
+The placement ID identifies the current container placement and changes when
+the container is replaced by the platform. Compare the returned value against
+a stored value to detect container replacement and trigger reconciliation.
+
+The placement ID is captured during the session-create handshake and cached
+in Durable Object storage, so reads are cheap and do not require a healthy
+container. A fresh placement ID is captured on each subsequent handshake,
+so a replacement is reflected the next time the sandbox is used.
+
+Returns `undefined` when no session-create handshake has been observed yet on
+this sandbox — call any method that triggers session creation (such as
+`exec()`) first. Returns `null` when a handshake has completed but the
+container's `CLOUDFLARE_PLACEMENT_ID` environment variable is not set, such as
+in local development with wrangler.
+
+```ts
+await sandbox.exec('true'); // ensures the handshake has run
+const containerPlacementId = await sandbox.getContainerPlacementId();
+if (
+  typeof containerPlacementId === 'string' &&
+  containerPlacementId !== lastKnownPlacementId
+) {
+  // Container was replaced — reconcile state
+}
+```
--- a/packages/sandbox-container/src/handlers/session-handler.ts
+++ b/packages/sandbox-container/src/handlers/session-handler.ts
@@ -79,11 +79,25 @@ export class SessionHandler extends BaseHandler<Request, Response> {
       const response = {
         success: true,
         data: result.data,
+        containerPlacementId: process.env.CLOUDFLARE_PLACEMENT_ID ?? null,
         timestamp: new Date().toISOString()
       };
 
       return this.createTypedResponse(response, context);
     } else {
+      // Include placement ID on the already-exists path so the DO can capture
+      // it after a DO restart where the container is still the original
+      // placement and the session was created before the DO restarted.
+      if (result.error.code === ErrorCode.SESSION_ALREADY_EXISTS) {
+        const errorWithPlacement = {
+          ...result.error,
+          details: {
+            ...(result.error.details ?? {}),
+            containerPlacementId: process.env.CLOUDFLARE_PLACEMENT_ID ?? null
+          }
+        };
+        return this.createErrorResponse(errorWithPlacement, context);
+      }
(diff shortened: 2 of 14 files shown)


aron-cf:
@ghostwriternr I've renamed the field, wdyt?

aron-cf (approved):

PR #593, opened 2026-04-16
Fix local R2 mounts prefix handling

### LINKED ISSUE #590: `localBucket: true` silently syncs nothing when prefix starts with / (contradicts docs and prod FUSE path)
**Summary**

When using `sandbox.mountBucket()` with `localBucket: true` in `wrangler dev`, a `prefix` option starting with `/` causes the sync layer to match zero objects. The mount appears to succeed, `fullSyncR2ToContainer` runs cleanly, and subsequent `ls` calls on the mount path return an empty directory. Writes from inside the container round-trip into the container-local view but land in R2 under keys with a leading `/` that will also fail to sync back out.

The docs and the prod FUSE path both require prefixes to start with `/`, so following the docs produces silent data loss in local dev.

**Repro**

```ts
// wrangler.jsonc has: r2_buckets: [{ binding: "R2", bucket_name: "mybucket", remote: true }]
// R2 bucket contains objects with keys like "some/prefix/files/foo.txt" (no leading /).

const sandbox = getSandbox(env.Sandbox, "repro");
await sandbox.mountBucket("R2", "/data", {
  localBucket: true,
  prefix: "/some/prefix/",  // leading slash as the docs recommend
  readOnly: true,
});
const result = await sandbox.exec("ls /data");
// Expected: the mirrored files
// Actual:   empty
```

Stripping the leading slash (`prefix: "some/prefix/"`) makes it work for the local-sync path — but then `mountBucketFuse` rejects the same value via `validatePrefix` in production.

**Root cause**

`packages/sandbox/src/local-mount-sync.ts`:

- `LocalMountSyncManager.listAllR2Objects()` passes the prefix verbatim to `R2Bucket.list({ prefix })`. R2 object keys never have a leading `/`, so `list({ prefix: "/some/prefix/" })` matches zero objects, and `fullSyncR2ToContainer()` populates the mount with nothing. Subsequent `pollR2ForChanges()` cycles continue to find zero changes.
- `containerPathToR2Key()` does `path.join(this.prefix, relativePath)`, so files written from inside the container are uploaded to R2 under keys like `/some/prefix/foo.txt` — actual leading slash in the key. These keys will never match a future `list({ prefix: "/some/prefix/" })` either in practice with most R2 tooling, and they show up oddly in the R2 dashboard.

Meanwhile `packages/sandbox/src/storage-mount/validation.ts` `validatePrefix()` throws `"Prefix must start with '/'"`, which is only called from `mountBucketFuse` — so the prod path enforces `/`-prefixed while the local path silently requires the opposite.

**Impact**

Silent: the mount appears healthy, `mountBucket` returns success, no errors are logged, `ls` returns "empty directory" rather than an error. Easy to mistake for "the bucket is really empty" or "the sync hasn't run yet."

**Suggested fix**

Normalize the prefix at the top of `mountBucketLocal` / inside `LocalMountSyncManager`:

```ts
// In LocalMountSyncManager constructor, or where the prefix is stored:
this.prefix = options.prefix?.replace(/^\//, "") || undefined;
```

This makes the local-sync path use the same raw-R2-key semantics that `R2Bucket.list()` and `R2Bucket.put()` expect, while keeping the public API consistent with the docs and the FUSE path. Both `r2KeyToContainerPath()` and `containerPathToR2Key()` already use the `this.prefix` value internally, so stripping once at construction time keeps round-trips correct.

Alternatively, call `validatePrefix()` from `mountBucketLocal` too, but normalize the value before using it — enforcing the documented `/`-prefixed convention at the API boundary and translating internally.

**Environment**

- `@cloudflare/sandbox` (observed in current `main`, path: `packages/sandbox/src/local-mount-sync.ts`)
- `wrangler dev` with `r2_buckets` binding set to `remote: true`
- R2 bucket populated from outside the sandbox (via `@aws-sdk/client-s3` from a Node process, standard leading-slash-free key format)

scuffi:
# Summary

`LocalMountSyncManager` passed the prefix verbatim to R2 APIs (`bucket.list()`, `bucket.put()`), but R2 object keys never have leading slashes. Since the docs and the production FUSE path both require prefixes to start with `/`, following the documented convention caused `list()` to match zero objects — the mount appeared healthy but the directory was always empty. Writes from inside the container also landed under malformed R2 keys with a leading `/`.

Additionally, `validatePrefix()` was only called from the FUSE path, so the local dev path silently accepted prefixes that production would reject.

# Changes

- **packages/sandbox/src/local-mount-sync.ts** - Constructor now calls `validatePrefix()` (matching production) then strips the leading `/` for R2 key compatibility
- **packages/sandbox/src/sandbox.ts** - `mountBucketLocal()` also calls `validatePrefix()` for defence in depth
- **packages/sandbox/tests/local-mount-sync.test.ts** - Updated all prefix tests to use `/`-prefixed values matching production; added tests for leading-slash normalisation in both sync directions, bare `/` handling, and rejection of non-leading-slash prefixes

Closes #590

### DIFF
--- a/packages/sandbox/Dockerfile
+++ b/packages/sandbox/Dockerfile
@@ -1,7 +1,5 @@
 # Bun version — override via --build-arg BUN_VERSION=$(cat .bun-version)
 ARG BUN_VERSION=1
-ARG UBUNTU_MIRROR=http://mirrors.edge.kernel.org/ubuntu
-ARG UBUNTU_PORTS_MIRROR=http://mirrors.edge.kernel.org/ubuntu-ports
 FROM oven/bun:${BUN_VERSION} AS bun-binary
 
 # Sandbox container images (default and python variants)
@@ -53,9 +51,6 @@ RUN npx turbo run build
 # ============================================================================
 FROM ubuntu:22.04 AS python-builder
 
-ARG UBUNTU_MIRROR
-ARG UBUNTU_PORTS_MIRROR
-
 # Prevent interactive prompts during package installation
 ENV DEBIAN_FRONTEND=noninteractive
 
@@ -67,15 +62,6 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
     --mount=type=cache,target=/var/lib/apt,sharing=locked \
     rm -f /etc/apt/apt.conf.d/docker-clean && \
     echo 'Binary::apt::APT::Keep-Downloaded-Packages "true";' >/etc/apt/apt.conf.d/keep-cache && \
-    if [ "$(dpkg --print-architecture)" = "amd64" ] || [ "$(dpkg --print-architecture)" = "i386" ]; then \
-        M=$UBUNTU_MIRROR; \
-    else \
-        M=$UBUNTU_PORTS_MIRROR; \
-    fi && \
-    echo "deb $M jammy main restricted universe multiverse" > /etc/apt/sources.list && \
-    echo "deb $M jammy-updates main restricted universe multiverse" >> /etc/apt/sources.list && \
-    echo "deb $M jammy-backports main restricted universe multiverse" >> /etc/apt/sources.list && \
-    echo "deb $M jammy-security main restricted universe multiverse" >> /etc/apt/sources.list && \
     apt-get update && apt-get install -y --no-install-recommends \
     wget ca-certificates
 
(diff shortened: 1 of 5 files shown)


packages/sandbox/Dockerfile
scuffi, line 1: ignore - temporary mirror whilst `archive.ubuntu.com` is having issues

whoiskatrin:
@scuffi  maybe consider centralizing validatePrefix at the public API boundary only

whoiskatrin (approved):

PR #599, opened 2026-04-17
Cache preview URL token validation at the Worker level

whoiskatrin:
## Problem

`proxyToSandbox()` calls `validatePortToken()` over RPC to the Sandbox Durable Object on **every** preview URL request. A single Vite dev page load fans out to 20+ subresource fetches against the same preview URL, producing 20+ RPCs to the DO for the same `{sandboxId, port, token}` triple. This is the dominant source of DO queue pressure for customers serving dev-server traffic through preview URLs — even when tokens are valid and stable, the sheer volume overwhelms the DO.

## Change

Cache successful token validations at the Worker level, keyed by `{sandboxId}:{port}:{token}`. Cached hits short-circuit the RPC entirely.

This PR is a **bridge toward signed (HMAC) tokens** that the Worker can verify locally with zero RPCs. That is the right long-term answer but requires a token-format change and customer migration; the cache unblocks customers today without touching the token format.

### Cache policy

| Property | Default | Rationale |
|---|---|---|
| Scope | Per-isolate `Map` | No shared state across isolates needed — worst case is one re-validation per isolate per triple. |
| TTL | **10 s** (configurable) | Long enough to absorb a page-load burst (which happens in sub-second); short enough that `unexposePort()` takes effect quickly. |
| Size cap | 10,000 entries, LRU-ish eviction (configurable) | Hits refresh insertion order so hot entries survive; oldest evicted first under pressure. |
| Successes only | ✅ | Failures during transient windows (e.g. mid-restart, before ports are re-exposed) **must** retry on the next request, not be pinned as invalid. |
| Invalidation | Implicit (TTL + `unexposePort`) | No cross-isolate bus needed. Worst-case revocation latency = TTL. |

### Public API

```ts
import {
  proxyToSandbox,
  TokenValidationCache,
  clearTokenValidationCache
} from '@cloudflare/sandbox';

// Simple: use the module-level default (10 s TTL, 10k entries).
await proxyToSandbox(request, env);

// Tuned: hold a cache instance and share it across handlers.
const cache = new TokenValidationCache({ ttlMs: 5_000, maxEntries: 1_000 });
await proxyToSandbox(request, env, { tokenValidationCache: cache });

// Disable entirely when you want instant revocation.
const noCache = new TokenValidationCache({ ttlMs: 0 });
await proxyToSandbox(request, env, { tokenValidationCache: noCache });
```

`clearTokenValidationCache()` clears the default cache — useful in tests and for callers that want to force re-validation after an out-of-band token change.

## Why this is safe

- **No cached failures.** A `false` result is never stored, so a revoked or not-yet-exposed port cannot get pinned as invalid, and recovery after re-exposure is instantaneous.
- **Authorization still flows through the DO on first contact.** The cache only accelerates *subsequent* requests within the TTL for an already-authorized triple.
- **Bounded memory.** 10,000-entry cap with LRU eviction; each entry is a short string key + timestamp number (~100 bytes → ~1 MB per isolate max).
- **No change to the security model.** Token binding, port validation, and the hostname → `{port, sandboxId, token}` extraction are all unchanged.
- **Opt-out available.** Operators who need instant revocation can disable caching per call (`ttlMs: 0`).

## Performance impact

For a Vite dev page load against a preview URL:

- **Before:** 20+ RPCs to the Sandbox DO per page load.
- **After:** 1 RPC per `{sandboxId, port, token}` per TTL window (10 s).

Preview traffic is removed as an overload vector for the DO. Under sustained load on the same preview URL, RPCs drop to effectively zero for validation.

## Testing

Nine tests in `tests/request-handler.test.ts` covering the cache invariants:

1. Repeated requests for the same triple call `validatePortToken` exactly once.
2. Different tokens on the same port validate independently.
3. Same token on different ports validates independently.
4. Failed validations are **not** cached — every failing request retries the DO.
5. `clearTokenValidationCache()` forces re-validation.
6. TTL expiry (fake timers) — inside TTL still cached, past TTL re-validates.
7. Size-cap eviction — third entry evicts the first when `maxEntries: 2`.
8. LRU refresh-on-hit — hitting A before inserting C evicts B, not A.
9. `ttlMs: 0` disables caching entirely.

Results:

- `npm run typecheck -w @cloudflare/sandbox` — clean
- `npm test -w @cloudflare/sandbox` — **568/568** passing (559 pre-existing + 9 new)

## Review responses

Incorporated from the internal review against `REVIEW.md`:

- **"Magic 30 s is a design smell."** Dropped default to 10 s and made it configurable; documented the trade-off explicitly in the `ttlMs` doc.
- **"Revocation is now eventually consistent."** Acknowledged in the PR body and in the changeset, with an explicit opt-out (`ttlMs: 0`).
- **"Why not HMAC?"** Addressed head-on: HMAC is the intended long-term path; this cache is a non-breaking bridge that ships today.
- **"Test the TTL + eviction branches directly."** Added four new tests covering exactly those paths with fake timers and a size-capped instance.
- **"Don't export internal test hooks."** `clearTokenValidationCache` is kept public because it has a real external use case (forcing re-validation after out-of-band token rotation), but it operates only on the module-level default; tuned instances are controlled by the caller.

## Changeset

Bumped to **minor** — the change expands the public API and changes observable revocation semantics beyond a pure bugfix.

### DIFF
--- a/.changeset/cache-token-validation.md
+++ b/.changeset/cache-token-validation.md
@@ -1,5 +1,5 @@
 ---
-'@cloudflare/sandbox': minor
+'@cloudflare/sandbox': patch
 ---
 
 Cache successful preview URL token validations at the Worker level so page
--- a/.changeset/cache-token-validation.md
+++ b/.changeset/cache-token-validation.md
@@ -1,11 +1,16 @@
 ---
-'@cloudflare/sandbox': patch
+'@cloudflare/sandbox': minor
 ---
 
-Cache successful preview URL token validations at the Worker level to avoid
-calling the Sandbox Durable Object on every preview request. Page loads that
-previously triggered 20+ RPCs to the DO for the same port and token now
-trigger one, with the rest served from a per-isolate, TTL-bounded cache.
-Failed validations are never cached, so transient "port not exposed" states
-recover immediately once the port is re-exposed. A `clearTokenValidationCache`
-helper is exported for callers that need to force re-validation.
+Cache successful preview URL token validations at the Worker level so page
+loads on preview URLs stop saturating the Sandbox Durable Object. Page loads
+that previously made 20+ RPCs for the same port and token now make one per
+cache window (10 s by default), with the rest served from a per-isolate,
+TTL- and size-bounded cache. Failed validations are never cached, so
+transient "port not exposed" states recover on the next request. Exports
+`TokenValidationCache`, `proxyToSandbox` accepts an optional
+`tokenValidationCache` so operators can tune TTL and size limits or share
+one instance across handlers, and `clearTokenValidationCache()` is exposed
+for tests and forced re-validation. This is a bridge toward signed (HMAC)
+tokens that can be verified locally with zero RPCs; the cache unblocks
+customers today without requiring a token-format change.
--- a/.changeset/cache-token-validation.md
+++ b/.changeset/cache-token-validation.md
@@ -1,0 +1,11 @@
+---
+'@cloudflare/sandbox': patch
+---
+
+Cache successful preview URL token validations at the Worker level to avoid
+calling the Sandbox Durable Object on every preview request. Page loads that
+previously triggered 20+ RPCs to the DO for the same port and token now
+trigger one, with the rest served from a per-isolate, TTL-bounded cache.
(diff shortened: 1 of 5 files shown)


scuffi (approved):
nice solution until signed tokens 🔥

PR #603, opened 2026-04-20
Add backup/restore & mount performance tests

scuffi:
# Summary

Adds two new performance scenarios — **bucket mounting** and **backup & restore** — to the perf suite.

Both require FUSE access and R2 credentials that are only available on the deployed test worker, so they create their own `PerfSandboxManager` pointed at `PERF_DEPLOYED_WORKER_URL` and skip gracefully when it isn't set. This keeps them isolated from the rest of the suite, which continues to use the local wrangler dev server started by `global-setup.ts`.

To prevent existing scenarios from regressing back to running against the deployed worker, `TEST_WORKER_URL` is replaced with `PERF_DEPLOYED_WORKER_URL` throughout — existing scenarios never see the deployed URL and fall back to local dev as before.

# Changes

- **`tests/perf/scenarios/bucket-mounting.test.ts`** — measures s3fs mount/unmount latency and write, read, and write→read roundtrip I/O through a FUSE-mounted R2 bucket.
- **`tests/perf/scenarios/backup-restore.test.ts`** — measures backup creation (`mksquashfs` → R2) and restore latency across small, medium, and large directory sizes; includes post-restore read and write verification.
- **`tests/perf/helpers/perf-sandbox-manager.ts`** — adds `mountBucket`, `unmountBucket`, `createBackup`, `restoreBackup`, and bucket object CRUD helpers used by the new scenarios.
- **`tests/perf/helpers/constants.ts`** — adds `BACKUP_RESTORE` / `BUCKET_MOUNTING` scenario constants and associated metric keys.
- **`tests/e2e/test-worker/wrangler.perf.template.jsonc`** — adds `BACKUP_BUCKET` R2 binding required by the backup scenario.
- **`.github/workflows/performance.yml`** — registers both new scenarios in the dispatch list; swaps `TEST_WORKER_URL` → `PERF_DEPLOYED_WORKER_URL` and exposes `CLOUDFLARE_ACCOUNT_ID`, `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY` to the test process (used by `mountOptions()` to pass credentials via request body).
- **`turbo.json`** — mirrors the `TEST_WORKER_URL` → `PERF_DEPLOYED_WORKER_URL` swap; adds `CLOUDFLARE_ACCOUNT_ID`.

---

<a href="https://app.devin.ai/review/cloudflare/sandbox-sdk/pull/603" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/.github/workflows/performance.yml
+++ b/.github/workflows/performance.yml
@@ -16,6 +16,8 @@ on:
           - concurrent-creation
           - burst-startup
           - file-io
+          - backup-restore
+          - bucket-mounting
   schedule:
     - cron: '0 0 * * *'
 
@@ -180,8 +182,11 @@ jobs:
 
       - name: Run performance tests
         env:
-          TEST_WORKER_URL: ${{ needs.setup.outputs.worker_url }}
+          PERF_DEPLOYED_WORKER_URL: ${{ needs.setup.outputs.worker_url }}
           CI: true
+          CLOUDFLARE_ACCOUNT_ID: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           PERF_RUN_ID: ${{ github.run_id }}
         run: |
           if [ "${{ inputs.scenario }}" = "all" ] || [ -z "${{ inputs.scenario }}" ]; then
--- a/tests/e2e/test-worker/index.ts
+++ b/tests/e2e/test-worker/index.ts
@@ -556,22 +556,6 @@ console.log('Terminal server on port ' + port);
 
       // Bucket mount
       if (url.pathname === '/api/bucket/mount' && request.method === 'POST') {
-        // Pass R2 credentials from worker env to sandbox env
-        const sandboxEnvVars: Record<string, string> = {};
-        if (env.CLOUDFLARE_ACCOUNT_ID) {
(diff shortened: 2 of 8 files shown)


aron-cf:
@scuffi CI review has good feedback that needs addressing.

scuffi:
> @scuffi CI review has good feedback that needs addressing.

Yeah, should've commented underneath it but it's not actually true, this variable never propagated into the vitest process anyways so it doesn't break tests. I've ran the performance tests a few times in CI and they pass properly, this is ready for a review 🙏

.github/workflows/performance.yml
```diff
@@ -180,8 +182,11 @@ jobs:
 
       - name: Run performance tests
         env:
-          TEST_WORKER_URL: ${{ needs.setup.outputs.worker_url }}
+          PERF_DEPLOYED_WORKER_URL: ${{ needs.setup.outputs.worker_url }}
           CI: true
+          CLOUDFLARE_ACCOUNT_ID: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
```
aron-cf, line 189: What's the scope of these keys? As I understand it, this job runs daily and can only be triggered manually by those with write access. So I think this is fine. 

We should probably add read only permissions so the job can only clone.

```
permissions:
  contents: read  # Read the code to test
```

aron-cf (approved):

.github/workflows/performance.yml
scuffi: This workflow already has read only permissions for contents, so we should be good?

PR #614, opened 2026-04-22
Make transport configurable per instance

aron-cf:
This adds a new `transport` config option to the `getSandbox()` interface. Which allows custom transport to be specified per instance rather than for all sandbox instances via a worker level environment variable.

This enables two things:

 1. Different sandboxes to use different transports.
 2. The use of multiple transports on the same sandbox.

Primarily this will enable us to reduce the number of workers and container images down to a single set in CI. Ideally the entire transport mechanism will go away shortly when we replace with a unified websocket based transport.

The integration tests have been updated to use this method via a `TEST_TRANSPORT` file written into the test worker directory and loaded into the environment via the vite env. The follow up will remove the dual tests in actions. Then this hack can go away.

---

<a href="https://app.devin.ai/review/cloudflare/sandbox-sdk/pull/614" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/tests/e2e/test-worker/generate-config.sh
+++ b/tests/e2e/test-worker/generate-config.sh
@@ -64,6 +64,9 @@ echo "    Standalone: $IMAGE_STANDALONE"
 echo "    Musl: $IMAGE_MUSL"
 echo "    Desktop: $IMAGE_DESKTOP"
 
+# Write transport into file and pull it in in vite config.
+echo "$TRANSPORT" > TEST_TRANSPORT;
+
 # Read template and replace placeholders
 # Using | as delimiter since image URLs contain /
 sed -e "s|{{WORKER_NAME}}|$WORKER_NAME|g" \
--- a/tests/e2e/test-worker/generate-config.sh
+++ b/tests/e2e/test-worker/generate-config.sh
@@ -64,9 +64,6 @@ echo "    Standalone: $IMAGE_STANDALONE"
 echo "    Musl: $IMAGE_MUSL"
 echo "    Desktop: $IMAGE_DESKTOP"
 
-# Write transport into file and pull it in in vite config.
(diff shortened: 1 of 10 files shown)


tests/e2e/helpers/global-sandbox.ts
```diff
@@ -43,13 +49,18 @@ export interface CreateTestSandboxOptions {
 export async function createTestSandbox(
   options: CreateTestSandboxOptions = {}
 ): Promise<TestSandbox> {
-  const { type = 'default', initCommand = 'echo ready', sleepAfter } = options;
+  const {
+    type = 'default',
+    transport = process.env.TEST_TRANSPORT,
```
aron-cf: TEST_TRANSPOST isn't currently set, so the test worker will default to using `SANDBOX_WORKER`. A follow up PR will refactor the github actions to swap these out. For now this validates that the `getSandbox()` option works.

tests/e2e/test-worker/generate-config.sh
```diff
@@ -77,4 +76,7 @@ sed -e "s|{{WORKER_NAME}}|$WORKER_NAME|g" \
     -e "s|{{IMAGE_DESKTOP}}|$IMAGE_DESKTOP|g" \
   wrangler.template.jsonc > wrangler.jsonc
 
+# Write transport to a file so the vitest config can set TEST_TRANSPORT
+echo -n "$TRANSPORT" > TEST_TRANSPORT
```
aron-cf: This is a temporary workaround so that the existing CI flow will work using the `getSandbox()` flow. In a follow up PR we'll drop the generation of the two workers and flip this out to use an environment variable when running the harness.

ghostwriternr (approved):

PR #616, opened 2026-04-22
Update authentication example with egress intercept

scuffi:
## Summary

**Before:** The example ran a JWT-signed proxy inside the Worker. The sandbox had to be configured on every request — URL rewrites, env var injection, and git config mutations — and credentials were visible to the sandbox as JWT tokens.

**After:** The sandbox makes plain requests to real hosts (`api.anthropic.com`, `github.com`, `http://r2.worker/...`). Outbound handlers defined on the `Sandbox` class intercept those requests in the Worker runtime and inject credentials from `env`. The sandbox is dumb: no setup, no tokens, no URL rewrites.

## Changes

- **`src/index.ts`** — Extends `Sandbox` with `outboundByHost` mapping the three services; exports `ContainerProxy` (required for interception); removes all JWT/proxy/configure logic from request handlers
- **`src/services/*/index.ts`** — Each service is now a single handler function (~10 lines each) instead of a `ServiceConfig` + `configureFn` pair
- **`src/proxy/`** — Deleted entirely (JWT token creation/verification, proxy handler, error types, ~250 lines)
- **`package.json`** — Removed `jose` dependency
- **`wrangler.jsonc`/ `.dev.vars.example`** — Removed `PROXY_JWT_SECRET`
- **`README.md`** — Rewritten to explain the interception model


examples/authentication/src/services/anthropic/index.ts
```diff
@@ -1,35 +1,20 @@
...
-    req.headers.set('x-api-key', ctx.env.ANTHROPIC_API_KEY);
-    return req;
+export function anthropicHandler(
+  request: Request,
+  env: Env
+): Promise<Response> {
+  if (!env.ANTHROPIC_API_KEY) {
+    return Promise.resolve(
+      new Response('ANTHROPIC_API_KEY not configured', { status: 500 })
+    );
```
aron-cf: I don't think you need the promise resolve here. Just make the function `async`.

examples/authentication/src/index.ts
```diff
@@ -1,78 +1,34 @@
...
-  await configureGithub(sandbox, proxyBase, token);
+export class Sandbox extends BaseSandbox {
+  interceptHttps = true;
 }
 
+Sandbox.outboundByHost = {
+  'api.anthropic.com': anthropicHandler,
+  'github.com': githubHandler,
+  'r2.worker': r2Handler
+};
```
aron-cf, line 19: Can we stick this inside the class using `static`? Keeps everything together. No strong opinion though.

aron-cf (approved):
Soooo much better! We're making progress!!

examples/authentication/src/index.ts
scuffi: Unfortunately we can't, this was an issue we discovered when implementing the handler. This is now the official way to use the API as documented

PR #617, opened 2026-04-22
Refactor desktop management + shutdown logic

aron-cf:
This started as a pass over the desktop e2e suite to make it reliable and ended up touching a bunch of stuff covering the container-side desktop stack the tests exercise: the shutdown path, the worker process model, the IPC protocol between the service and its FFI worker, and the bun server's resilience to rogue failures bubbling up from any of them.

## E2E test refactor

Every fetch in the suite now carries an abort signal timeout so a container that stops responding fails the individual test rather than wedging the whole run. JSON body assertions use `.resolves` instead of awaiting and re-asserting, unused type aliases are gone, and the stop/error cases have been consolidated into a single sequential test — they were testing overlapping state transitions and running them as siblings produced order-dependent failures. Traces are enabled in the e2e test worker so future flakes come with a trail.

## Desktop stop determinism

The manager now spawns each managed process via `node:child_process` with `detached: true`, which makes every child its own process group leader. On stop, all groups receive SIGTERM simultaneously and a single eight-second deadline covers the entire teardown; survivors are SIGKILL'd once it expires. This replaces the old sequential `killProcess` loop and the per-process `PROCESS_KILL_TIMEOUT_MS` constant, and bounds the worst-case stop time to a predictable number.

Two follow-ups hardened the shutdown path further. The manager's shutdown now uses `Promise.allSettled` so a single failing teardown step doesn't prevent the follow-up hard shutdown from triggering. Concurrent `stop()` calls coalesce onto the same promise to avoid double-teardown races, and `stopping` is treated the same as `inactive` when gating operations so nothing slips through during the transition. A separate fix ensures the desktop has fully stopped before a new worker is spawned after a stop, closing a window where the worker could come up against half-dead X11 processes.

## Worker crash isolation

The desktop worker now runs in a dedicated child process via `Bun.spawn` instead of a Worker thread. A Worker thread shares the bun HTTP server's address space, so a segfault in the robotgo or X11 FFI layer takes everything down with it — which is exactly what was happening when X11 died while an FFI call was in flight. A child process means the kernel cleans up the crash and the parent just sees the subprocess exit.

Process isolation also made it safe to flip the stop ordering: the worker is now terminated before X11 processes die, where previously the service killed X11 first "so any in-flight FFI call fails fast." That prior ordering was a workaround for the in-process failure mode; with the worker in its own process, terminating it first is both safer and less racy. Worker termination is wrapped in a two-second SIGTERM deadline that escalates to SIGKILL, and every `sendToWorker` call carries its own ten-second timeout so no single operation can block callers indefinitely.

## Bun server resilience

Lastly, we no longer hard crash on uncaught promises. Instead we log an error. Ideally we should have Sentry or some form of telemetry here to catch this (at least in our integration tests).

### DIFF
--- a/packages/sandbox-container/src/managers/desktop-manager.ts
+++ b/packages/sandbox-container/src/managers/desktop-manager.ts
@@ -158,17 +158,19 @@ export class DesktopManager {
     }
 
     // 2. Wait for all to exit, bounded by STOP_DEADLINE_MS.
+    // allSettled never rejects — a single process crashing won't
+    // short-circuit the wait or skip the SIGKILL escalation.
     const exitPromises = allProcs
       .filter((p) => p.proc && !p.proc.killed)
       .map((p) => p.proc!.exited);
 
-    const allExited = Promise.all(exitPromises);
+    const allSettled = Promise.allSettled(exitPromises);
     const deadline = new Promise<'timeout'>((resolve) =>
       setTimeout(() => resolve('timeout'), STOP_DEADLINE_MS)
     );
 
     const result = await Promise.race([
-      allExited.then(() => 'exited' as const),
+      allSettled.then(() => 'exited' as const),
       deadline
     ]);
 
--- a/packages/sandbox-container/src/managers/desktop-manager.ts
+++ b/packages/sandbox-container/src/managers/desktop-manager.ts
@@ -3,14 +3,17 @@ import type {
   DesktopStartRequest,
   Logger
 } from '@repo/shared';
+import type { Subprocess } from 'bun';
 
 interface DesktopProcess {
   name: string;
   command: string;
   args: string[];
   priority: number;
-  proc: import('bun').Subprocess | null;
+  proc: Subprocess<'ignore', 'pipe', 'pipe'> | null;
   pid: number | undefined;
+  /** PGID for group-kill. Equals the child PID when spawned with `detached`. */
+  pgid: number | undefined;
   startTime: Date | null;
 }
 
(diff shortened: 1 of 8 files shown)


tests/e2e/test-worker/Dockerfile.desktop
```diff
@@ -1,6 +1,6 @@
 # E2E test Dockerfile for Desktop environment
 # Uses the -desktop variant for testing desktop/computer-use features
-FROM docker.io/cloudflare/sandbox-test:0.8.13-desktop
+FROM docker.io/cloudflare/sandbox-test:0.8.12-desktop
```
aron-cf:
```suggestion
FROM docker.io/cloudflare/sandbox-test:0.8.13-desktop
```

packages/sandbox-container/src/managers/desktop-manager.ts
```diff
@@ -183,39 +240,47 @@ export class DesktopManager {
...
+      args: desktopProcess.args
     });
 
     const env: Record<string, string> = {
-      ...(process.name !== 'xvfb' ? { DISPLAY: ':99' } : {})
+      ...(desktopProcess.name !== 'xvfb' ? { DISPLAY: ':99' } : {})
     };
 
-    const proc = Bun.spawn([process.command, ...process.args], {
+    // `detached: true` calls setsid(2), making the child a process group
```
ghostwriternr, line 254: no-op: nice, i remember trying to do this for our entire session execution layer because orphan processes ended up remaining after parent gets killed but it was messy. but in this case, much cleaner to handle this here.

ghostwriternr (approved):

ghostwriternr:
needs a rebase, lgtm otherwise. thanks for cleaning up the mess around the FFI worker lifecycle + isolation

PR #624, opened 2026-04-23
Make Sandbox DO configuration setters idempotent

ghostwriternr:
## The bug

When a Worker passes configuration options to `getSandbox()` on every request — a documented pattern — the per-isolate configuration cache starts empty in every freshly-created Worker isolate and reapplies those options to the Sandbox Durable Object.

Several DO setters treated "same value reapplied" as user activity:
- `setSleepAfter` called `renewActivityTimeout()` unconditionally.
- `setKeepAlive` rewrote storage on every invocation.
- `setContainerTimeouts` reconfigured the transport retry budget.

With enough isolate churn, the sandbox's idle timer is reset indefinitely and the sandbox never sleeps. This is the failure mode surfaced by `streaming-operations-workflow.test.ts`'s "should keep sandbox alive during execStream beyond sleepAfter value" scenario.

## Prior art on these setters

The setter behaviour this change modifies was introduced for specific reasons. None of it is cruft, and all of it is preserved for real value transitions:

- **#323** added `renewActivityTimeout()` to `setSleepAfter` so custom `sleepAfter` values passed to `getSandbox()` actually take effect. The base `Container` constructor arms the timer with defaults before the RPC arrives, so without a renewal the caller's value is ignored until the next activity.
- **#476** added `renewActivityTimeout()` to `setKeepAlive(false)` so sandboxes return to their configured sleep lifecycle after keepAlive is disabled.
- **#498** batched configuration into a single `configure()` RPC and introduced the per-isolate `WeakMap` cache to skip redundant apply work when the same isolate handles multiple requests.
- **#504** persisted `sleepAfter` to storage so custom values survive DO eviction.

The bug is not in any of those; it's that the setters ran their side effects on every invocation, including no-op reapply, and the cache cannot prevent a reapply when a fresh Worker isolate is created.

## The fix

Each affected setter now returns early when its argument matches the current value. Re-applying the same `sleepAfter`, `keepAlive`, or `containerTimeouts` is a true no-op — no storage write, no timer reset, no retry-budget recompute. Real transitions still run the same work they did before, so #323 and #476 continue to hold.

`setContainerTimeouts` needs a small addition: its defaults come from env vars with SDK fallbacks, so a first explicit call whose values happen to equal the env-derived defaults would silently not persist under naive in-memory equality, and the user's intent would be lost on DO eviction if env vars later change. A `hasStoredContainerTimeouts` flag gates the guard so the first durable write always happens regardless of value equality.

## Adjacent fixes

- `baseUrl` is now restored from storage in the constructor `blockConcurrencyWhile` block. It was persisted by `setBaseUrl` but never rehydrated, so after DO eviction the set-once immutability check could be bypassed by a different value.
- `setSandboxName` writes its two fields (`sandboxName` and `normalizeId`) via coalesced `storage.put` calls without an intervening await. On SQLite-backed DOs these commit atomically, so a mid-write failure cannot persist one field without the other.
- Every modified setter now writes storage before updating its in-memory mirror, so a failed write leaves the class field consistent with durable state.

## What's not in this change

The `WeakMap` cache and the fire-and-forget apply path are unchanged. With the setters now idempotent, the cache remains a valid optimization when an isolate is reused across requests, and reapply in a fresh isolate is harmless.

Revisiting that layer later — to surface apply errors to callers, or to share a single pending `configure()` promise across concurrent first calls on the same stub — is a separate refactor with its own tradeoffs.

---

<a href="https://app.devin.ai/review/cloudflare/sandbox-sdk/pull/624" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/packages/sandbox/src/sandbox.ts
+++ b/packages/sandbox/src/sandbox.ts
@@ -708,18 +708,22 @@ export class Sandbox<Env = unknown> extends Container<Env> implements ISandbox {
     }
   }
 
-  // RPC method to set the base URL
+  // RPC method to set the base URL. Set-once — subsequent calls with the
+  // same value are no-ops; subsequent calls with a different value throw.
+  //
+  // A failed storage write must not poison the in-memory immutability
+  // guard, so storage.put runs before the in-memory assignment. On the
+  // next retry from the same hot DO, this.baseUrl is still null and the
+  // guard will allow the write.
   async setBaseUrl(baseUrl: string): Promise<void> {
-    if (!this.baseUrl) {
-      this.baseUrl = baseUrl;
-      await this.ctx.storage.put('baseUrl', baseUrl);
-    } else {
-      if (this.baseUrl !== baseUrl) {
-        throw new Error(
-          'Base URL already set and different from one previously provided'
-        );
-      }
+    if (this.baseUrl === baseUrl) return;
+    if (this.baseUrl !== null) {
+      throw new Error(
+        'Base URL already set and different from one previously provided'
+      );
     }
+    await this.ctx.storage.put('baseUrl', baseUrl);
+    this.baseUrl = baseUrl;
   }
 
   // RPC method to set the sleep timeout.
--- a/packages/sandbox/src/sandbox.ts
+++ b/packages/sandbox/src/sandbox.ts
@@ -734,10 +734,15 @@ export class Sandbox<Env = unknown> extends Container<Env> implements ISandbox {
     this.renewActivityTimeout();
   }
 
-  // RPC method to enable keepAlive mode
+  // RPC method to enable keepAlive mode.
+  // Idempotent — applying the same value is a no-op. When disabling (going
+  // from true to false), the activity timer is renewed so the inactivity
+  // window starts counting from now rather than from whenever keepAlive was
+  // first enabled.
   async setKeepAlive(keepAlive: boolean): Promise<void> {
-    this.keepAliveEnabled = keepAlive;
+    if (this.keepAliveEnabled === keepAlive) return;
     await this.ctx.storage.put('keepAliveEnabled', keepAlive);
+    this.keepAliveEnabled = keepAlive;
 
     if (!keepAlive) {
       this.renewActivityTimeout();
--- a/packages/sandbox/src/sandbox.ts
+++ b/packages/sandbox/src/sandbox.ts
@@ -103,7 +103,6 @@ type SandboxConfiguration = {
     name: string;
     normalizeId?: boolean;
   };
-  baseUrl?: string;
   sleepAfter?: string | number;
   keepAlive?: boolean;
   containerTimeouts?: NonNullable<SandboxOptions['containerTimeouts']>;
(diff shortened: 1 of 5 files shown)


packages/sandbox/src/sandbox.ts
```diff
@@ -665,13 +676,24 @@ export class Sandbox<Env = unknown> extends Container<Env> implements ISandbox {
     });
   }
 
+  // RPC method to set the sandbox name and normalizeId. Set-once — a
+  // subsequent call is a no-op.
+  //
+  // sandboxName and normalizeId are one logical unit. Both storage.put
+  // calls are dispatched synchronously inside Promise.all, which coalesces
+  // them into a single atomic commit on SQLite-backed Durable Objects.
+  // In-memory state is updated only after both writes commit.
```
aron-cf: Is this because they both happen in the same microtask? Do we not ideally want all config to be set in the same commit?

packages/sandbox/src/sandbox.ts
```diff
@@ -699,32 +721,39 @@ export class Sandbox<Env = unknown> extends Container<Env> implements ISandbox {
     }
   }
 
-  // RPC method to set the base URL
+  // RPC method to set the base URL. Set-once — subsequent calls with the
+  // same value are no-ops; subsequent calls with a different value throw.
+  // Storage is written before the in-memory assignment so the immutability
+  // guard reflects durable state.
```
aron-cf: This will blow up the `getSandbox()` call right?

aron-cf (approved):
Nice, non-blocking comments inline.

packages/sandbox/src/sandbox.ts
ghostwriternr: it's our runtime behaviour - `storage.put()` writes to an in-memory write buffer that's flushed atomically as a single transaction. so as long as there's no await between two put() calls, they land in the same flush. see [write buffer behaviour](https://developers.cloudflare.com/durable-objects/api/sqlite-storage-api/#put:~:text=Write%20buffer%20behavior) and [DO best practices](https://developers.cloudflare.com/durable-objects/best-practices/rules-of-durable-objects/#understand-how-input-and-output-gates-work:~:text=Explain%20Code-,Write%20coalescing,-%3A%20Multiple%20storage) from the docs.

here since both puts are dispatched synchronously inside the Promise.all array literal before any await, they land in the same buffer and flush as one atomic commit.

packages/sandbox/src/sandbox.ts
ghostwriternr: the apply is fire-and-forget and wrapped with a try/catch, so errors will be swallowed? i think the primary difference in behaviour is that a caller passing a different `baseUrl` after the DO has one stored now silently fails instead of silently overwriting.

but now that you mention it... what the heck is baseUrl is even used for anymore? trying to remember why we introduced it

ghostwriternr: i chucked it out. there was zero reason for it to exist and cause unnecessary maintenance burden.

PR #632, opened 2026-04-24
Surface SESSION_TERMINATED when a session's shell exits

whoiskatrin:
## Summary

Fixes a bug where a sandbox session becomes permanently unusable after its underlying shell process exits. Once triggered, every subsequent operation on the affected sandbox fails with `Session '<id>' is not ready or shell has died`, and the only recovery is `sandbox.destroy()` — which loses `/workspace` state and forces a backup restore.

Reported by a customer running a multi-tenant agent platform; a single user command brought down the shell and their agent then failed 193 consecutive sandbox calls over 23 minutes before giving up. Reproducible from `exit 0`.

Closes the issue by modelling shell death as a first-class lifecycle event instead of silently papering over it.

## The bug

When a session's shell dies (user runs `exit`, shell crashes, OOM, child process takes the shell down):

1. `session.ts` exit handler flips `this.ready = false`, but leaves the `Session` object in `SessionManager.sessions`.
2. Every `getOrCreateSession(sessionId)` returns that dead handle unchecked.
3. `ensureReady()` throws `"Session is not ready or shell has died"` on every subsequent exec.
4. `createSession({ id })` refuses to overwrite and returns `SESSION_ALREADY_EXISTS`.
5. `deleteSession` is blocked at the SDK layer for the default session.
6. `sandbox.destroy()` is the only escape.

## The fix

Treat shell death as a typed, first-class error instead of a generic command failure, and give the SDK a deterministic recovery path.

### New error code: `SESSION_TERMINATED` (HTTP 410)

Distinct from the existing `SESSION_DESTROYED`. Mirrors the runtime's internal `ShellTerminatedError` vs `SessionDestroyedError` split:

- `SESSION_DESTROYED` → API-initiated `destroy()` during an in-flight command.
- `SESSION_TERMINATED` → shell exited on its own (user `exit`, crash, child killed parent). Carries the observed exit code.

Wired through `@repo/shared/errors` (code, status map, suggestion, context) and exposed at the SDK level as `SessionTerminatedError` with typed `sessionId` and `exitCode` accessors.

### Behavior changes in `SessionManager`

1. **`getOrCreateSession` fast path**: if the existing session isn't ready, evict it and return `SESSION_TERMINATED`. The next call finds no session and creates a fresh one through the normal path. **No silent recreate.**
2. **`executeInSession` / `withSession` / `executeStreamInSession`**: when the exec failure is a shell termination (typed `ShellTerminatedError` or an explicit `exit N` that raced past the type), return `SESSION_TERMINATED` (not `COMMAND_EXECUTION_ERROR`) and evict the dead handle under the already-held per-session lock.
3. **`createSession`** on a dead session id: acquire the per-session lock, evict, recreate. Deterministic explicit-recovery API for callers who prefer to be explicit. Healthy duplicates still return `SESSION_ALREADY_EXISTS`.

### Session state semantics

A fresh session does **not** inherit the dead session's env vars, cwd, shell functions, or background jobs — because they no longer exist. The `SESSION_TERMINATED` error (and its `suggestion` string) tells the caller this explicitly, so the automatic recovery is not a silent semantic swap.

## My thinking

The naive fix is to have `getOrCreateSession` transparently recreate a dead session and return the fresh handle. I started there. It's ~10 lines and the existing tests pass.

But transparently swapping the shell out from under the caller violates the semantic contract that a session id refers to a persistent shell. The caller's next `echo $MY_VAR` would return empty with `success: true` and no signal that anything happened, turning a recoverable lifecycle event into a hard-to-debug data loss bug — especially for agent platforms where session-local state matters.

So the design here is:

- **Surface once**: the first call after shell death returns `SESSION_TERMINATED` with the exit code. The caller learns their session-local state is gone.
- **Auto-recover after**: the dead handle is evicted as part of surfacing the error, so the next call creates a fresh session through the normal path. No manual recovery required.
- **Explicit recovery API**: `createSession({ id: "main" })` on a dead session id replaces it in place, for callers who prefer to be explicit (and because the reporter pointed out they already try this in their retry path and it currently fails).

This matches the existing `ShellTerminatedError` / `SessionDestroyedError` split that already lives in the container runtime — those types existed for exactly this distinction, but the error classification was collapsing both into `COMMAND_EXECUTION_ERROR` on the way out.

### What I deliberately didn't do

- **Did not relax the `deleteSession` default-session guard.** The reporter suggested this, but with `SESSION_TERMINATED` providing a typed signal and auto-recreation on the next call, there's no longer a forced path through `sandbox.destroy()`. Changing that guard is a public-API semantic decision that deserves its own PR
- **Did not touch `ProcessManager` / `PortManager` cleanup.** Verified in-flight background commands already flip to `'error'` status via `shellExitedPromise` propagating through `executeStreamInSession`; port registrations are container-scoped, not session-scoped, so unaffected by session recreation.

## Reviewer notes

Things worth scrutinizing, in roughly priority order:

1. **Is the semantic honesty worth the breaking change for `exit N` callers?** Previously, `exec('exit 1')` returned `COMMAND_EXECUTION_ERROR`. It now returns `SESSION_TERMINATED` (HTTP 410 instead of 500). Any caller code matching on `COMMAND_EXECUTION_ERROR` for `exit N` will need to update. I believe this is the right call (it's the same semantic event — shell gone — just now surfaced honestly), but it's the one visible behavior change for existing users. One existing unit test codified the old behavior and was updated (`session-manager-locking.test.ts:347`).

2. **Locking contract on `createSession` recovery branch.** `getOrCreateSession` documents that callers hold the per-session lock. `createSession` historically didn't. The new dead-replace branch takes the lock itself (the public HTTP handler doesn't). Documented inline (`session-manager.ts` around the `createSession` `existing` check) — please confirm the contract reads cleanly and matches the check-then-set pattern used in `setSandboxName` and the process pool reserve/spawn/register flow.

3. **Double-eviction is idempotent.** Two code paths evict: the `getOrCreateSession` fast path, and the exec catch blocks after a `ShellTerminatedError`. Both re-check `isReady()` before calling `evictDeadSession`, and both run under the per-session lock for the given `sessionId`, so concurrent callers serialize. Worth a second look.

4. **E2E coverage.** Added to `tests/e2e/session-state-isolation-workflow.test.ts` (reuses the existing shared container). Validates the full user path: create session with env var → `exit 42` → 410 `SESSION_TERMINATED` → next call succeeds with empty env → sandbox as a whole still alive. This is the test that would have caught the original bug.

5. **Why no `resetSession()` API?** Considered it. Decided `createSession({ id })` replacing a dead session is the explicit-recovery API — no new surface, and it matches what the reporter already tries in their retry path. Open to adding one if there's a reason the caller needs "reset even if alive".

6. **SDK error class plumbing.** `SessionTerminatedError` added to `packages/sandbox/src/errors/classes.ts`, wired in `adapter.ts` and `index.ts`. Follows the exact pattern of `SessionDestroyedError` — lockstep with the sibling error so there's nothing surprising.


whoiskatrin:
Closing — reopening from the upstream branch instead of the fork.

PR #633, opened 2026-04-24
Surface SESSION_TERMINATED when a session's shell exits

whoiskatrin:
## Summary

Fixes a bug where a sandbox session becomes permanently unusable after its underlying shell process exits. Once triggered, every subsequent operation on the affected sandbox fails with `Session '<id>' is not ready or shell has died`, and the only recovery is `sandbox.destroy()` — which loses `/workspace` state and forces a backup restore.

Reported by a customer running a multi-tenant agent platform; a single user command brought down the shell and their agent then failed 193 consecutive sandbox calls over 23 minutes before giving up. Reproducible from `exit 0`.

Closes the issue by modelling shell death as a first-class lifecycle event instead of silently papering over it.

## The bug

When a session's shell dies (user runs `exit`, shell crashes, OOM, child process takes the shell down):

1. `session.ts` exit handler flips `this.ready = false`, but leaves the `Session` object in `SessionManager.sessions`.
2. Every `getOrCreateSession(sessionId)` returns that dead handle unchecked.
3. `ensureReady()` throws `"Session is not ready or shell has died"` on every subsequent exec.
4. `createSession({ id })` refuses to overwrite and returns `SESSION_ALREADY_EXISTS`.
5. `deleteSession` is blocked at the SDK layer for the default session.
6. `sandbox.destroy()` is the only escape.

## Why this was possible in the first place

Not a race, not a missed edge case — a **missing contract between two layers**.

`git log -S "this.ready = false" -- packages/sandbox-container/src/session.ts` shows the pattern entered the codebase in `7aee736b` ("Rewrite SDK with cleaner design patterns and tests", #95, Oct 2025). That rewrite landed two things in the same commit that needed to stay in lockstep but didn't:

1. A `ready` flag on `Session` that flips to `false` on shell exit and gates `exec` via `ensureReady()`.
2. A `SessionManager.sessions` map that indexes sessions by id.

At the time of the rewrite, `SessionManager` was minimal — `createSession` + `getSession` + an `executeInSession` that created on-demand. The only failure mode modelled was *initial-create failure*. Dead sessions post-creation simply weren't in the model. Later PRs layered more behavior on top — per-session mutexes (#290), creation-race coordination via `creatingLocks`, the `getOrCreateSession` fast path — and each one inherited the unspoken "map presence ⇒ valid session" assumption without revisiting it. Even the PR closest to this area (#452, "Fix null pointer in shellExitedPromise during concurrent session destroy") was focused on correctness of the exit *signal*, not on what the manager does with the corpse afterwards.

The result: two halves of the system disagreed about what `sessions.has(id)` meant.

| Layer             | `sessions.has(id)` meant                                |
| ----------------- | ------------------------------------------------------- |
| `Session`         | "there is an object; its shell may or may not be alive" |
| `SessionManager`  | "there is a live, usable session"                       |

`Session` maintained its half correctly. `SessionManager` assumed the other half without checking. **Presence ≠ validity** — the cache key was being used as a liveness check, and the two aren't the same.

### Why the existing tests didn't catch it

- **Creation and destruction were covered, mid-life shell death was not.** No test existed where a shell exited *on its own* and the session was then reused. The new `session-manager-dead-session.test.ts` (added in this PR) is the first to exercise that path.
- **The happy path hides it.** Most commands either succeed or return a non-zero exit code; they don't take the shell down. The reproducer needs a command that actually terminates the PTY (`exit 0`, or a child that closes the controlling terminal — the customer's `monica-cli image generate` case).
- **The error message read as transient.** "Session is not ready or shell has died" sounds retryable. It isn't — it's permanent state corruption. The customer's Go retry layer retried 193 times on exactly that reasoning.
- **`SESSION_ALREADY_EXISTS` masked the manual recovery path.** A caller who correctly diagnosed the problem and tried `createSession` got "you already have one" and fell straight back into the dead handle.

The fix has to be more than evicting dead entries — it has to make the lifecycle event *typed* and *honestly surfaced*, so the same trap doesn't reopen next time someone adds a code path.

## The fix

Treat shell death as a typed, first-class error instead of a generic command failure, and give the SDK a deterministic recovery path.

### New error code: `SESSION_TERMINATED` (HTTP 410)

Distinct from the existing `SESSION_DESTROYED`. Mirrors the runtime's internal `ShellTerminatedError` vs `SessionDestroyedError` split:

- `SESSION_DESTROYED` → API-initiated `destroy()` during an in-flight command.
- `SESSION_TERMINATED` → shell exited on its own (user `exit`, crash, child killed parent). Carries the observed exit code.

Wired through `@repo/shared/errors` (code, status map, suggestion, context) and exposed at the SDK level as `SessionTerminatedError` with typed `sessionId` and `exitCode` accessors.

### Behavior changes in `SessionManager`

1. **`getOrCreateSession` fast path**: if the existing session isn't ready, evict it and return `SESSION_TERMINATED`. The next call finds no session and creates a fresh one through the normal path. **No silent recreate.**
2. **`executeInSession` / `withSession` / `executeStreamInSession`**: when the exec failure is a shell termination (typed `ShellTerminatedError` or an explicit `exit N` that raced past the type), return `SESSION_TERMINATED` (not `COMMAND_EXECUTION_ERROR`) and evict the dead handle under the already-held per-session lock.
3. **`createSession`** on a dead session id: acquire the per-session lock for the entire check → evict → construct → initialize → set sequence, then replace. Deterministic explicit-recovery API for callers who prefer it. Healthy duplicates still return `SESSION_ALREADY_EXISTS`.

### Session state semantics

A fresh session does **not** inherit the dead session's env vars, cwd, shell functions, or background jobs — because they no longer exist. The `SESSION_TERMINATED` error (and its `suggestion` string) tells the caller this explicitly, so the automatic recovery is not a silent semantic swap.

## Two follow-up gaps the initial fix didn't cover

Review against the first round of changes surfaced two more defects that fell under the same "presence ≠ validity" and "every code path that can observe shell death must route through `SESSION_TERMINATED`" umbrella. Both are fixed in this PR.

### Gap 1 — `withSession` didn't classify `ShellTerminatedError`

`withSession`'s catch block only unwrapped errors with a `code` field matching an `ErrorCode`. `ShellTerminatedError` and `SessionDestroyedError` are plain `Error` subclasses with **no** `code` field (see `packages/sandbox-container/src/errors.ts`), so they fell through to the generic branch and were wrapped as `INTERNAL_ERROR`. `withSession` also did not call `evictDeadSession` on the way out.

Impact: `withSession` is the foundation for `setEnvVars` (`session-manager.ts:1053`), `writeFile`, `readFile`, and several other file-service and git-service paths. If a user's prior command killed the shell and they then called any of those, they got `INTERNAL_ERROR` instead of `SESSION_TERMINATED`, and recovery was deferred until the *next* call caught `!isReady()` in the `getOrCreateSession` fast path. The retry logic that the customer widened to key on `SESSION_TERMINATED` would still miss this case.

**Fix:** match `SessionDestroyedError` / `ShellTerminatedError` explicitly at the top of the catch block, evict under the held lock, return the proper typed error. Mirrors what `executeInSession` already does.

### Gap 2 — `createSession` dead-replace and fresh-create races

The previous `createSession` comment claimed:

> The dead-replace branch below takes the lock itself so that any concurrent executeInSession serializes behind the recreate.

But the implementation only held the lock for the **eviction**. `new Session`, `await session.initialize()`, and `this.sessions.set(...)` all ran **outside** the lock. Two races:

**Dead-replace race:**

1. Request A: `createSession({id:'X'})` on a dead session. Enters dead-replace branch, acquires lock, evicts, releases lock.
2. Request B: `executeInSession('X', …)`. Acquires the now-free lock, `getOrCreateSession` sees empty map, constructs a Session, `initialize()`, `sessions.set('X', B)`, runs a command, releases lock.
3. Request A resumes, constructs its own Session, `initialize()`, `sessions.set('X', A)` — **overwrites B**.
4. B is orphaned: live bash PTY, session directory on disk, mutex entry, never destroyed.

**Fresh-create race:** two concurrent `createSession({id:'X'})` on a non-existent id both pass `sessions.get(...) === undefined`, both construct, both initialize, both `set`. Loser orphaned identically. `creatingLocks` was designed to coordinate this but `createSession` didn't consult it.

**Fix:** hold the per-session lock across the entire check → evict → construct → initialize → set sequence. Concurrent `executeInSession` / `withSession` / streaming callers now genuinely serialize behind the recreate — as the comment always claimed — and two concurrent `createSession` calls on the same id also serialize. Removes both races in one move.

## My thinking

The naive fix is to have `getOrCreateSession` transparently recreate a dead session and return the fresh handle. I started there. It's ~10 lines and the existing tests pass.

But transparently swapping the shell out from under the caller violates the semantic contract that a session id refers to a persistent shell. The caller's next `echo $MY_VAR` would return empty with `success: true` and no signal that anything happened, turning a recoverable lifecycle event into a hard-to-debug data loss bug — especially for agent platforms where session-local state matters.

So the design here is:

- **Surface once**: the first call after shell death returns `SESSION_TERMINATED` with the exit code. The caller learns their session-local state is gone.
- **Auto-recover after**: the dead handle is evicted as part of surfacing the error, so the next call creates a fresh session through the normal path. No manual recovery required.
- **Explicit recovery API**: `createSession({ id: "main" })` on a dead session id replaces it in place, for callers who prefer to be explicit (and because the reporter pointed out they already try this in their retry path and it currently fails).

This matches the existing `ShellTerminatedError` / `SessionDestroyedError` split that already lives in the container runtime — those types existed for exactly this distinction, but the error classification was collapsing both into `COMMAND_EXECUTION_ERROR` on the way out.

## Reviewer notes

Things worth scrutinizing, in roughly priority order:

1. **Is the semantic honesty worth the breaking change for `exit N` callers?** Previously, `exec('exit 1')` returned `COMMAND_EXECUTION_ERROR`. It now returns `SESSION_TERMINATED` (HTTP 410 instead of 500). Any caller code matching on `COMMAND_EXECUTION_ERROR` for `exit N` will need to update. I believe this is the right call (it's the same semantic event — shell gone — just now surfaced honestly), but it's the one visible behavior change for existing users. One existing unit test codified the old behavior and was updated (`session-manager-locking.test.ts:347`).

2. **Locking contract on `createSession`.** `createSession` historically did not hold the per-session lock. It now holds it across the entire lifecycle (check → evict → construct → initialize → set). That includes `session.initialize()` which has multiple await points. This is deliberate — it's the only way to make the stated concurrency contract true — and different session ids use different locks so cross-session parallelism is unaffected. Please confirm the scope reads cleanly.

3. **Double-eviction is idempotent.** Three code paths evict now: the `getOrCreateSession` fast path, the exec catch blocks after a `ShellTerminatedError`, and the `withSession` catch block. All re-check `isReady()` before calling `evictDeadSession`, and all run under the per-session lock, so concurrent callers serialize. Worth a second look.

4. **E2E coverage.** Added to `tests/e2e/session-state-isolation-workflow.test.ts` (reuses the existing shared container). Validates the full user path: create session with env var → `exit 42` → 410 `SESSION_TERMINATED` → next call succeeds with empty env → sandbox as a whole still alive. This is the test that would have caught the original bug.

5. **Unit coverage for the two follow-up gaps.** `session-manager-dead-session.test.ts` now includes:
   - `surfaces SESSION_TERMINATED from inside withSession callbacks` — guards Gap 1.
   - `surfaces SESSION_TERMINATED from setEnvVars after a prior shell death` — the highest-value `withSession` user.
   - `createSession holds the lock across evict, create, and set, so concurrent executeInSession cannot orphan a session` — guards Gap 2 dead-replace race.
   - `two concurrent createSession calls on the same fresh id serialize` — guards Gap 2 fresh-create race.

6. **Why no `resetSession()` API?** Considered it. Decided `createSession({ id })` replacing a dead session is the explicit-recovery API — no new surface, and it matches what the reporter already tries in their retry path. Open to adding one if there's a reason the caller needs "reset even if alive".

7. **SDK error class plumbing.** `SessionTerminatedError` added to `packages/sandbox/src/errors/classes.ts`, wired in `adapter.ts` and `index.ts`. Follows the exact pattern of `SessionDestroyedError` — lockstep with the sibling error so there's nothing surprising.


.changeset/session-terminated.md
```diff
@@ -0,0 +1,31 @@
+---
+'@cloudflare/sandbox': patch
```
aron-cf: I think this is a minor bump. We're changing the exec() contract with a minor breaking change. Not a strongly held opinion though, it's technically a bug fix.

.changeset/session-terminated.md
```diff
@@ -0,0 +1,31 @@
...
+Recover a sandbox session after its shell exits, instead of poisoning the sandbox.
+
+Previously, if a session's underlying shell process exited — because a user
+command ran `exit`, the shell crashed, or a child process took the shell down —
+the sandbox kept returning the dead session handle. Every subsequent call threw
+`Session is not ready or shell has died`, and `createSession` refused to replace
+the dead session with `SESSION_ALREADY_EXISTS`. The only recovery path was
+`sandbox.destroy()`, which loses `/workspace` state and forces a backup restore.
+
+The SDK now models shell death as a first-class error:
```
aron-cf: I think this is still technically correct and a "nicer" word.

```suggestion
The SDK now models shell termination as a first-class error:
```

.changeset/session-terminated.md
```diff
@@ -0,0 +1,31 @@
...
+- `createSession({ id })` on a dead session id now replaces the dead handle
+  instead of returning `SESSION_ALREADY_EXISTS`, giving callers a deterministic
+  recovery API. The evict + recreate sequence runs entirely under the
+  per-session lock, so it cannot race with a concurrent `executeInSession` or a
+  concurrent second `createSession` and orphan a Session with a live bash PTY.
+- Shell-death errors thrown from inside `withSession` callbacks (used by
+  `setEnvVars`, `writeFile`, `readFile`, git clone, etc.) now surface as
+  `SESSION_TERMINATED` and evict the dead handle eagerly, matching the behavior
+  of `executeInSession`. Previously they leaked through as `INTERNAL_ERROR` and
+  recovery was deferred to the next call.
```
aron-cf: This is a changelog entry so it needs to be rewritten from the perspective of an end user. A code example showing how to handle this with a short summary would probably be better.

packages/sandbox-container/src/services/session-manager.ts
```diff
@@ -78,10 +79,49 @@ export class SessionManager {
...
+      if (existing.isReady()) {
+        return { success: true, data: existing };
+      }
+
+      const exitCode = existing.getShellExitCode();
+      this.logger.warn('Evicting terminated session', {
+        sessionId,
+        exitCode
+      });
+      await this.evictDeadSession(sessionId, existing);
```
aron-cf:
```suggestion
      await this.evictTerminatedSession(sessionId, existing);
```

packages/sandbox-container/src/services/session-manager.ts
```diff
@@ -349,6 +503,20 @@ export class SessionManager {
...
         }
 
+        if (shellTerminated) {
+          // Shell exited during the command. Evict the dead handle under
+          // the lock we already hold, so the next call on this sessionId
+          // creates a fresh session instead of hitting the stale handle.
+          const dead = this.sessions.get(sessionId);
+          if (dead && !dead.isReady()) {
+            await this.evictDeadSession(sessionId, dead);
+          }
```
aron-cf:
```suggestion
          const session = this.sessions.get(sessionId);
          if (session && !session.isReady()) {
            await this.evictTerminatedSession(sessionId, session);
          }
```

packages/sandbox-container/src/services/session-manager.ts
```diff
@@ -298,6 +418,39 @@ export class SessionManager {
...
+   * and that the next call on this sessionId creates a fresh session.
+   */
+  private async evictDeadSession(
+    sessionId: string,
+    session: Session
+  ): Promise<void> {
+    try {
+      await session.destroy();
+    } catch (error) {
+      this.logger.debug('Dead session destroy() threw during eviction', {
```
aron-cf:
```suggestion
      this.logger.debug('Terminated session, destroy() threw during eviction', {
```

packages/sandbox-container/src/services/session-manager.ts
```diff
@@ -424,6 +592,27 @@ export class SessionManager {
 
         return serviceSuccess<T>(result);
       } catch (error) {
+        // Shell-death errors thrown from inside the callback's exec()
```
aron-cf:
```suggestion
        // Errors thrown from inside the callback's exec() for a terminated session
```

packages/sandbox-container/src/services/session-manager.ts
```diff
@@ -424,6 +592,27 @@ export class SessionManager {
...
+        // already hold so the next call creates a fresh session.
+        if (error instanceof SessionDestroyedError) {
+          return serviceError<T>(this.sessionDestroyedError(sessionId));
+        }
+
+        if (error instanceof ShellTerminatedError) {
+          const dead = this.sessions.get(sessionId);
+          if (dead && !dead.isReady()) {
+            await this.evictDeadSession(sessionId, dead);
+          }
```
aron-cf:
```suggestion
          const session = this.sessions.get(sessionId);
          if (session && !session.isReady()) {
            await this.evictTerminatedSession(sessionId, session);
          }
```

packages/sandbox-container/src/services/session-manager.ts
```diff
@@ -574,6 +764,17 @@ export class SessionManager {
           };
         }
 
+        if (shellTerminated) {
+          const dead = this.sessions.get(sessionId);
+          if (dead && !dead.isReady()) {
+            await this.evictDeadSession(sessionId, dead);
+          }
```
aron-cf:
```suggestion
          const session = this.sessions.get(sessionId);
          if (session && !session.isReady()) {
            await this.evictTerminatedSession(sessionId, session);
          }
```

packages/sandbox-container/src/services/session-manager.ts
```diff
@@ -689,6 +891,17 @@ export class SessionManager {
...
```
aron-cf:
```suggestion
          const session = this.sessions.get(sessionId);
          if (session && !session.isReady()) {
            await this.evictTerminatedSession(sessionId, session);
          }
```

packages/sandbox-container/tests/services/session-manager-terminated-session.test.ts
aron-cf, line 1: Rename file to `session-manager-terminated-session.test.ts`

packages/sandbox-container/tests/services/session-manager-dead-session.test.ts
```diff
@@ -0,0 +1,385 @@
+/**
+ * Session Manager dead-session semantics
```
aron-cf:
```suggestion
 * Session Manager terminated-session semantics
```

aron-cf (approved):
Conditionally approving. I'm happy to re-review if needed.

On the whole the change looks good, my two requests:

1. Let's reword the changelog to something understandable by an end-user. Giving an example of calling exec('exit 1') and recovering would be useful.
2. The changes interchange the terms "terminated", "dead" and "death". I don't want to be reading about dead process or process death, so can we consolidate the terminology on "terminate" and "terminated" which is a little more technical.

PR #641, opened 2026-04-27
Fix production backup restore mounts

whoiskatrin:
## Summary
- Keep successful production restores backed by their live backup session so FUSE-mounted restored files remain available after `restoreBackup()` returns.
- Pass the real backup ID through the restore API instead of deriving mount paths from `/var/backups/r2mount/<id>/data.sqsh`.
- Add unit and E2E regression coverage for production restore mount paths and post-restore readability/writability.

## Testing
- `bun test packages/sandbox-container/tests/services/backup-service.test.ts`
- `npm run typecheck -w @cloudflare/sandbox`
- `npm run typecheck` via pre-push hook

## Notes
Focused E2E was added but local execution previously stopped during Docker image build before the test ran; CI should exercise it in the configured environment.

### DIFF
--- a/.changeset/restore-backup-mount-paths.md
+++ b/.changeset/restore-backup-mount-paths.md
@@ -1,0 +1,5 @@
+---
+'@cloudflare/sandbox': patch
+---
+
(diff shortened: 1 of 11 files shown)


whoiskatrin:
Closing this fork-based draft in favor of an upstream-branch PR.

PR #645, opened 2026-04-27
Normalize globstar exclude patterns for backups

scuffi:
# Summary
This PR fixes `createBackup()` excludes that use gitignore-style globstar patterns like `**/node_modules` and `dist/**`.

Previously, these patterns could make it through to the backup pipeline in a form that `mksquashfs` does not support, which could produce empty archives. This change normalizes those exclude patterns before archive creation and applies the same behavior in both the Durable Object and container backup paths so local and production flows stay consistent.

# Testing
- Added regression coverage for SDK backup paths
- Verified container backup-service tests still pass
- Verified targeted SDK backup tests pass

### DIFF
--- a/.changeset/backup-globstar-excludes.md
+++ b/.changeset/backup-globstar-excludes.md
@@ -1,0 +1,7 @@
+---
+'@cloudflare/sandbox': patch
+---
+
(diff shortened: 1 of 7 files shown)


whoiskatrin (approved):
Nice 👍🏼 this was my miss, thanks for picking this up

PR #665, opened 2026-05-01
Separate container control from route transports

ghostwriternr:
The primary Sandbox Durable Object to container path was named around the capnweb/RPC implementation, while the route-based HTTP/WebSocket compatibility transports lived nearby. That made it harder to tell where future control-plane work belongs and which code exists for compatibility.

This renames the SDK side to `container-control` and the container side to `control-plane`, removes shim-only compatibility files, and documents the route-based HTTP/WebSocket transports as compatibility paths. The public `SandboxOptions.transport` contract still supports `rpc`; the route transport helpers remain internal to the route-based client layer.

No changeset is included because the package-level runtime API is unchanged.

### DIFF
--- a/packages/sandbox-container/src/control-plane/README.md
+++ b/packages/sandbox-container/src/control-plane/README.md
@@ -1,0 +1,14 @@
+# Control plane
+
+This directory owns the container-side control API called by the Sandbox Durable
+Object.
+
+The current implementation is exposed through capnweb over the `/rpc` WebSocket
+route, but the architectural boundary is the container control plane, not the
+transport mechanism.
+
+Container control operations are implemented here and call the underlying
+services directly. Route handlers under `handlers/` and `routes/` are kept for
+route-based HTTP/WebSocket compatibility.
+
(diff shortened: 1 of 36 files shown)


packages/sandbox-container/src/control-plane/README.md
```diff
@@ -0,0 +1,15 @@
...
+The current implementation is exposed through capnweb over the `/rpc` WebSocket
+route, but the architectural boundary is the container control plane, not the
+transport mechanism.
+
+Container control operations are implemented here and call the underlying
+services directly. Route handlers under `handlers/` and `routes/` are kept for
+route-based HTTP/WebSocket compatibility.
+
+The shared `@repo/shared` `SandboxAPI` interface remains named `SandboxAPI` for
+compatibility. Renaming that shared interface is deferred to Phase 2.
```
aron-cf: What does this mean? What is Phase 2?

ghostwriternr: my b, trying new agent harness and missed this bit when reviewing. fixing - but basically just not touching this for now and leaving it for when we clean out older transports.

whoiskatrin (approved):

PR #666, opened 2026-05-01
Improve backup/restore transfer speeds with parallelization

scuffi:
## Summary

This PR improves backup and restore performance for larger archives by:

- Switching the default squashfs compression to `lz4` with configurable compression threads.
- Adding multipart R2 uploads for large backup archives.
- Adding parallel ranged downloads during restore to reduce archive download time.
- Introducing backup options for `compression`, `compressThreads`, and `multipart`.

## Details

Large backups now upload through R2 multipart upload when enabled, while smaller backups continue using the existing single presigned PUT path. Restores now download larger archives in parallel ranges before mounting the squashfs archive in the container.

The new options allow callers to tune backup behavior:

- `compression`: choose `lz4`, `zstd`, or `gzip`
- `compressThreads`: configure mksquashfs processor count
- `multipart`: disable multipart upload when needed

### DIFF
--- a/packages/sandbox-container/src/services/backup-service.ts
+++ b/packages/sandbox-container/src/services/backup-service.ts
@@ -11,6 +11,8 @@ import type { SessionManager } from './session-manager';
 
 export const BACKUP_WORK_DIR = '/var/backups';
 const BACKUP_MOUNTS_DIR = '/var/backups/mounts';
+const BACKUP_ALLOWED_COMPRESSIONS = ['gzip', 'lz4', 'zstd'] as const;
+type BackupCompression = (typeof BACKUP_ALLOWED_COMPRESSIONS)[number];
 
 /**
  * Absolute paths for squashfs/FUSE binaries.
@@ -61,6 +63,10 @@ function validateBackupPaths(dir: string, archivePath: string): string | null {
   return null;
 }
 
+function isBackupCompression(value: string): value is BackupCompression {
+  return BACKUP_ALLOWED_COMPRESSIONS.includes(value as BackupCompression);
+}
+
 interface CreateArchiveResult {
   sizeBytes: number;
   archivePath: string;
@@ -101,7 +107,7 @@ export class BackupService {
     sessionId = 'default',
     gitignore = false,
     excludes: string[] = [],
-    compression: 'gzip' | 'lz4' | 'zstd' = 'lz4',
+    compression: string = 'lz4',
     compressThreads = 8
   ): Promise<ServiceResult<CreateArchiveResult>> {
     const opLogger = this.logger.child({ operation: Operation.BACKUP_CREATE });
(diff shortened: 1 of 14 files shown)


packages/shared/src/request-types.ts
```diff
@@ -159,9 +159,56 @@ export interface CreateBackupRequest {
   gitignore?: boolean;
   /** Glob patterns to exclude from the backup */
   excludes?: string[];
+  /** Compression algorithm for mksquashfs (default: lz4) */
+  compression?: 'gzip' | 'lz4' | 'zstd';
+  /** Number of parallel compression threads (default: 8) */
+  compressThreads?: number;
```
aron-cf: Maybe combine this into a single options object? 

```
compression: { format?: "gzip" | "lz4" | "zsdt"; threads?: number }
```

packages/sandbox-container/src/services/backup-service.ts
```diff
@@ -407,6 +424,105 @@ export class BackupService {
...
+    parts: Array<{
+      partNumber: number;
+      url: string;
+      offset: number;
+      size: number;
+    }>,
+    sessionId = 'default'
+  ): Promise<
+    ServiceResult<{ parts: Array<{ partNumber: number; etag: string }> }>
+  > {
```
aron-cf: Why is this using bash and not just doing the read/post in JS?

packages/sandbox-container/src/services/backup-service.ts
```diff
@@ -407,6 +424,105 @@ export class BackupService {
...
+        `-D ${shellEscape(hdrFiles[i])}`,
+        '--data-binary @-',
+        shellEscape(part.url)
+      ].join(' ');
+      return `(set -o pipefail; ${ddCmd} | ${curlCmd}) & J${i}=$!`;
+    });
+
+    const waitLines = parts.map((_, i) => `wait $J${i}; E${i}=$?`);
+    const exitVars = parts.map((_, i) => `$E${i}`).join(' + ');
+    const etagLines = parts.map(
```
aron-cf: This feels like we're doing this in the wrong layer. We have a Bun here that has first class support for efficient file streaming in an easily testable typed language. Let's just use JavaScript.

We also have RPC support now, which has good support for promise pipelining, this can help us with the error handling.

We can ultimately do something like:

```ts
try {
     const archive = container.backup.createArchive(dir, opts);
     const uploaded = container.backup.uploadParts(archive.path, partsFor(archive.size));
     await archive.cleanup(uploaded);   // pipelined: no extra RTT to start
} catch {
      // handle backup failure
}
```

aron-cf (approved):
Huge improvement 👏  well done for working through this.

I think we can make the code a little more robust if we lean on the new RPC pipeline and extract as much as possible out of the shell exec calls.

For now can we make this feature RPC only and take a moment to decide if we want to refactor it.

aron-cf (approved):
Very nice!

PR #667, opened 2026-05-01
Add default session free exec mode

whoiskatrin:
## Summary

This reworks the original isolated exec PR around the broader API direction from the RFC discussion:

- Add an opt-in sandbox-level mode: `getSandbox(env.Sandbox, id, { defaultSession: false })`.
- In that mode, top-level `sandbox.exec()` runs without using the implicit persistent default session.
- Keep existing behavior unchanged by default: `sandbox.exec()` and `session.exec()` still preserve shell state unless callers opt out.
- Keep the per-call escape hatch from the earlier version: `sandbox.exec(command, { sessionId: false })`.
- Remove `preserveShellState` from the public/API-facing request shapes so the model stays focused on sessions rather than shell-state flags.
- Keep explicit sessions unchanged for callers that want persistent shell state.
- Continue rejecting `sessionId: false` for streaming/background command paths, because this PR only implements isolated foreground execution.

## API Shape

Default behavior remains persistent and backwards-compatible:

```ts
const sandbox = getSandbox(env.Sandbox, "my-sandbox");

await sandbox.exec("cd repo");
await sandbox.exec("pwd"); // repo
```

New sandbox-level opt-in mode:

```ts
const sandbox = getSandbox(env.Sandbox, "my-sandbox", {
  defaultSession: false
});

await sandbox.exec("cd repo");
await sandbox.exec("pwd"); // /workspace, not repo
```

Per-call opt-out still works:

```ts
await sandbox.exec(commitAndPushScript, {
  sessionId: false
});
```

Explicit sessions keep persistent semantics:

```ts
const session = await sandbox.createSession();
await session.exec("cd repo");
await session.exec("pwd"); // repo
```

## Why

The original PR fixed the acute `exec()` failure mode by allowing a single foreground command to avoid mutating or terminating the persistent default session shell. That handles generated scripts that call `exit`, `exec`, or use `set -e`.

The RFC discussion broadened the framing: the core issue is the implicit default session behind top-level Sandbox APIs. This version keeps `exec` as the public verb, avoids adding a new `run()` entry point, and introduces a sandbox-level opt-in that starts moving top-level APIs away from hidden persistent shell state.

This PR intentionally starts with top-level `exec()` only. File operations and other top-level methods still use the existing default-session behavior and should be consolidated in follow-up work.

## Implementation Notes

- `SandboxOptions` now accepts `defaultSession?: boolean`.
- The setting is passed through the existing `getSandbox()` configuration flow and persisted on the Sandbox DO.
- When `defaultSession` is false and no explicit `sessionId` is provided, top-level `exec()` routes to the existing isolated foreground path.
- `sessionId: false` remains unsupported for streaming/background commands.
- `preserveShellState` remains an internal container/session implementation detail only; it is no longer exposed through SDK command options or shared request/RPC types.

## Testing

- `npm run check`
- `npm test -w @cloudflare/sandbox -- --run tests/get-sandbox.test.ts tests/sandbox.test.ts -t "defaultSession|top-level exec|sessionId: false|configure|streaming exec"`

Both pass locally.


aron-cf:
@whoiskatrin funny I ran into this last week and was thinking about this over the weekend. Will take a look.

aron-cf (changes requested):
I think this generally looks good and I fully agree with the direction it's going.

My main feedback is the `preserveShellState` option doesn't feel like the right fit. The fact it leaks into execStream and needs to be handled is a sign of this. 

Essentially we're saying run this outside of a session. Ideally top level exec wouldn't have a session at all, but we need to keep backwards compatibility.

So my suggestions would be one of:

 1. Add a new method `execSync()` that mirrors `execStream` but blocks.
 2. Overload `sessionId` in `exec()` to support passing `false` to mean, no session, rather than the default session.
3. Rename `preserveShellState` to `session: boolean`.

I think, I'm leaning towards (2) because it provides a "fix" now but leaves us room to remove the whole sessionId field in a future breaking change and doesn't introduce new concepts.

ghostwriternr (commented):
i'm ngl, i'm increasingly anxious about adding more to the api surface and overall complexity of the already complex session execution layer. are there no other answers to how such scripts can be run? do we have an example of a script that we can use to understand where exactly our session layer breaks and try to fix the session layer rather than introducing new code paths?

whoiskatrin:
Reopening as draft to rework the isolated exec approach around a broader default-session-free top-level API mode.

PR #673, opened 2026-05-06
Inject credentials into HTTP requests in Claude Code example

aron-cf:
We use the new outbound HTTP request handler to intercept requests to the Anthropic API passing fake credentials into the container as environment variables as well as blocking all outbound internet access allowing the agent to run free.

This means that the container no longer has visibility into these API keys.

We also escape the input parameters passed into the `exec()` call.

I wanted to also allow `--permission-mode bypassPermissions` but Claude Code prohibits this when running as root for security reasons. I've created #677 and #678 to track this.

---

<a href="https://app.devin.ai/review/cloudflare/sandbox-sdk/pull/673" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


scuffi (approved):
nice 🚀

PR #674, opened 2026-05-06
Add experimental sandbox.tunnels interface

aron-cf:
Adds a new `sandbox.tunnels.*` API that exposes a service running inside a sandbox container to the public internet via [Cloudflare Tunnel], sitting alongside the existing `sandbox.exposePort()` mechanism rather than replacing it.

Two modes:

```ts
// Quick tunnel — zero config, no credentials.
const tunnel = await sandbox.tunnels.create(8080, { mode: 'quick' });
// → https://<random-words>.trycloudflare.com

// Named tunnel — bound to a hostname under a zone you own.
const tunnel = await sandbox.tunnels.create(8080, {
  mode: 'named',
  hostname: 'preview.example.com',
});
// → https://preview.example.com

// Or with creds set in env, no hostname → derived as `<sandbox-id>.<zone-name>`
const tunnel = await sandbox.tunnels.create(8080);

await sandbox.tunnels.list();
await sandbox.tunnels.destroy(tunnel);
```

A worked end-to-end example lives at `examples/tunnel-preview/` (follow up PR).

## Why not just route through the Worker?

Today's `exposePort()` uses a Worker-fronted preview URL: traffic hits the Worker, the Worker calls into the Durable Object, the DO routes to the container. That's a great fit for HTTP request/response, but it has real ceilings for streaming / long-lived workloads:

- **Wildcard DNS.** Preview URLs encode the port + sandbox + token in the subdomain (`8080-sandbox-id-token.example.com`), which requires a wildcard CNAME *and* a custom domain (no `.workers.dev`).
- **Local Dev** The shared origin in local development means that tools like vite often accidentally intercept traffic intended for the sandbox.
- **CPU time per request.** A Worker invocation is bounded by your plan's CPU limit. A long-lived WebSocket or a multi-minute SSE stream can run into that ceiling well before the workload would naturally end.
- **Subrequest count.** Each Worker invocation has a hard subrequest cap. Anything that fans out (one request → many container calls) hits it.
- **Token-based auth on every request.** Adds latency on the hot path.

A Cloudflare Tunnel sidesteps all of those. Traffic terminates at the CF edge, gets routed straight into a `cloudflared` process inside the container over a persistent QUIC/HTTP2 connection, and never hits your Worker. Your Worker's CPU and subrequest budgets aren't touched. The hostname is whatever you point at the tunnel — no wildcard required.

This is purely additive. `exposePort()` continues to work; the two mechanisms can coexist on the same sandbox. Though we can deprecate `exposePort()` if this proves to be a better fit.

### Quick tunnels

`cloudflared tunnel --url http://localhost:<port>` running inside the container talks to `api.trycloudflare.com`, gets handed a random `<words>.trycloudflare.com` hostname + ephemeral creds, and opens a persistent connection to the CF edge. Zero credential setup; ideal for debug / demo / "throwaway" use.

### Named tunnels

The DO handles the creation of the tunnel for the account provided as well as setting up the DNS. The resources created are tagged with sandbox id so they can be audited later (on enterprise plans we use the `tags` feature).

The DO orchestrates four steps:

1. `POST /accounts/:id/cfd_tunnel` — create a CF tunnel resource and receive its opaque `--token`. Tagged with `metadata: { sandboxId, createdBy: 'sandbox-sdk' }` for later reconciliation.
2. `POST /zones/:id/dns_records` — upsert a `CNAME` at the requested hostname, pointing at `<tunnel-id>.cfargotunnel.com`, `proxied: true`, with `comment: 'sandbox-<id>'` (universal) and `tags: ['sandbox-sdk', 'sandbox-<id>']` (Enterprise; falls back to no-tags on 4xx).
3. RPC call into the container: `tunnels.runTokenTunnel(id, token, port)`. The container spawns `cloudflared tunnel run --token <…> --url http://localhost:<port>`.
4. Container's `TunnelManager` polls cloudflared's `/ready` metrics endpoint; once `readyConnections >= 1` the call returns.

If any step fails, the DO rolls back the previous resources via `Promise.allSettled` so a single failure doesn't leak.

```mermaid
sequenceDiagram
    autonumber
    participant App as User Worker
    participant DO as Sandbox DO
    participant CF as Cloudflare API
    participant Container as Sandbox Container
    participant Edge as Cloudflare Edge

    App->>DO: tunnels.create(port, { mode: 'named', hostname })
    DO->>CF: POST /accounts/:id/cfd_tunnel
    CF-->>DO: { id, token }
    DO->>CF: POST /zones/:id/dns_records (CNAME → <id>.cfargotunnel.com)
    CF-->>DO: { recordId }
    DO->>Container: RPC tunnels.runTokenTunnel(id, token, port)
    Container->>Container: spawn cloudflared --token <…> --url http://localhost:<port>
    Container->>Edge: persistent QUIC/HTTP2 connection
    Container->>Container: poll /ready until readyConnections ≥ 1
    Container-->>DO: { id, port, createdAt }
    DO-->>App: { id, mode: 'named', url, hostname, ... }

    Note over App,Edge: Public traffic now flows: client → Edge → cloudflared → localhost:port
```

> [!NOTE]
> The implementation is split across the SDK (DO) and the container runtime, with the **Cloudflare API token kept strictly on the DO side**. The container only ever receives a per-tunnel `--token` string — never the API token, account id, or zone id.

### Decision tree on `tunnels.create()`

| Inputs                                | Mode                                                             |
| ------------------------------------- | ---------------------------------------------------------------- |
| `{ mode: 'quick' }`                   | quick                                                            |
| `{ mode: 'named', hostname }`         | named with that hostname                                         |
| no options + creds in `env`           | named with auto-derived `<sandbox-id>.<zone-name>`               |
| no options + no creds in `env`        | quick                                                            |

For named tunnels the DO reads three env vars (never forwarded to the
container):

```
CLOUDFLARE_API_TOKEN
CLOUDFLARE_ACCOUNT_ID
CLOUDFLARE_ZONE_ID
```

The token needs two scopes:

| Permission                            | Scope                          |
| ------------------------------------- | ------------------------------ |
| Account → Cloudflare Tunnel → Edit    | the account in `..._ACCOUNT_ID` |
| Zone → DNS → Edit                     | the zone in `..._ZONE_ID`       |

### Quick tunnels (`*.trycloudflare.com`)

- **No SSE.** The trycloudflare edge buffers `text/event-stream` responses, so server-sent events don't reach the client. WebSockets work fine. Named tunnels don't have this issue.
- **No uptime guarantee.** Cloudflare positions trycloudflare as a debug aid, not a production target.
- **No persistent hostname.** Every restart picks a new `<words>.trycloudflare.com`.

### Named tunnels

- **DNS propagation latency.** First-load on a fresh hostname adds a few seconds before the URL becomes globally reachable. The `tunnels.create()` call itself returns once cloudflared has registered with the edge (~3s); the public URL becomes resolvable shortly after.
- **TLS depth.** Universal SSL covers the apex + one wildcard level (`*.example.com`). Going two levels deep (`*.preview.example.com`) needs either a delegated subdomain zone or Advanced Certificate Manager.
- **Quota.** Free Cloudflare accounts cap at 1,000 tunnels per account. Orphan cleanup is a follow-up (see below).
- **DO eviction = orphaned CF resources.** Today the DO holds the only in-memory record of which CF tunnel/DNS resources belong to which sandbox. If the DO is evicted before `destroy()` runs, the resources persist in your account. The follow-up reconciler addresses this via `metadata.sandboxId` filtering.
- **WARP / Zero Trust intercepts cloudflared.** Local dev behind a TLS-intercepting proxy can't establish the cloudflared edge handshake. `DOCKER_README.md` documents the workaround for the CA-bundle setup; cloudflared still won't authenticate against an intercepted edge, so end-to-end runs need WARP off.

## Tested

- Unit tests for `TunnelService` with `Bun.spawn` mocked at the cloudflared boundary; the real `TunnelManager` is exercised end-to-end.
- Unit tests for the Cloudflare API client (CF API mocked).
- E2E workflows for quick and named tunnels in `tests/e2e/tunnel-{quick,named}-workflow.test.ts`.

## Follow-ups (separate PRs)

1. **Sleep / restart cleanup.** A cron-Worker reconciler that uses `metadata.sandboxId` and `comment` filters to find and delete orphaned tunnels + DNS records.
2. **Persist tunnel state in DO storage** so `destroy()` after a DO restart can still tear down CF resources without scanning.
3. Discover the account id + zone id from the token itself.

[Cloudflare Tunnel]: https://developers.cloudflare.com/cloudflare-one/connections/connect-networks/


eashish93:
This is much needed to solve HMR issue #689

aron-cf:
Closing in favor of #674 this stands as a draft implementation reference

PR #679, opened 2026-05-07
Fix local backup/restore when using RPC transport

aron-cf:
This change fixes restore flow on the RPC transport to stream the archive to the sandbox using the streaming write file flow. This avoids the 32 MiB limit on RPC payloads.

For backup we now always stream the data to R2 regardless of transport type avoiding buffering the archive into memory.

### DIFF
--- a/.changeset/fix-local-backup-rpc-streaming.md
+++ b/.changeset/fix-local-backup-rpc-streaming.md
@@ -1,0 +1,5 @@
+---
+'@cloudflare/sandbox': patch
+---
+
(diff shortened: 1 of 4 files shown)


whoiskatrin:
Nice regression coverage! 😄  One small thing I’d tighten up: after mutating blob.bin, the test actually doesn’t currently assert that the mutation actually succeeded or that the hash changed before restore. If tha request failed for some reason, the final SHA check could still pass because the original file was never modified. Could we parse the mutation response and assert exitCode === 0 plus mutatedSha!== originalSha before calling restore?

whoiskatrin (approved):

PR #683, opened 2026-05-08
Add RPC only readFile({ encoding: 'none' }) to return ReadableStream

aron-cf:
Adds encoding: 'none' to `readFile()` which returns a `ReadableStream<Uint8Array>`
of raw binary data on the rpc transport — no base64 encoding or buffering. Mirrors the existing `writeFile()` support for `ReadableStream` input.

```ts
const { content, size, mimeType } = await sandbox.readFile(
  '/workspace/image.png',
  { encoding: 'none' }
);

return new Response(content, {
  headers: { "Content-Type": mimeType }
});
```

This removes the need for parsing SSE streams, `streamFile`, `collectFile` etc.

Also moves `writeFileStream` from a top-level method on `SandboxClient` / `ContainerControlClient` onto `FileClient` (client.files), where all other file operations live. No change to the public sandbox.writeFile(path, stream) API.

---

<a href="https://app.devin.ai/review/cloudflare/sandbox-sdk/pull/683" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/packages/sandbox-container/src/services/file-service.ts
+++ b/packages/sandbox-container/src/services/file-service.ts
@@ -1686,6 +1686,97 @@ export class FileService implements FileSystemOperations {
    * Resolve a complete path in the context of the session's current working directory.  If the
    * provided path is relative, we append the session's current working directory to it.
    */
+  /**
+   * Stream raw binary file contents without SSE framing or base64 encoding.
+   * Intended for the RPC transport where the stream passes directly to the
+   * caller over the capnp binary channel — no encoding overhead.
+   *
+   * Validates the path, resolves relative paths via the session's cwd, then
+   * returns Bun's native file stream alongside the file size and MIME type.
+   *
+   * Returns a `ServiceResult` so callers receive typed `ErrorCode` values
+   * (e.g. `FILE_NOT_FOUND`, `VALIDATION_FAILED`) rather than plain throws.
+   */
+  async readFileBinaryStream(
+    path: string,
+    sessionId = 'default'
+  ): Promise<
+    ServiceResult<{
+      content: ReadableStream<Uint8Array>;
+      size: number;
+      mimeType: string;
+    }>
+  > {
+    // Validate path for security.
+    const validation = this.security.validatePath(path);
+    if (!validation.isValid) {
+      return {
+        success: false,
+        error: {
+          message: `Invalid path format for '${path}': ${validation.errors.join(', ')}`,
+          code: ErrorCode.VALIDATION_FAILED,
+          details: {
+            validationErrors: validation.errors.map((e) => ({
+              field: 'path',
+              message: e,
+              code: 'INVALID_PATH'
+            }))
+          } satisfies ValidationFailedContext
+        }
+      };
+    }
+
+    // Resolve relative paths via the session's working directory.
+    let resolvedPath = path;
+    if (!path.startsWith('/')) {
+      const result = await this.sessionManager.withSession(
+        sessionId,
+        async (exec) => this.resolvePathInSession(path, exec)
+      );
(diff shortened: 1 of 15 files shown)


packages/sandbox-container/src/services/file-service.ts
```diff
@@ -1686,6 +1686,56 @@ export class FileService implements FileSystemOperations {
...
+    path: string,
+    sessionId = 'default'
+  ): Promise<{
+    content: ReadableStream<Uint8Array>;
+    size: number;
+    mimeType: string;
+  }> {
+    const validation = this.security.validatePath(path);
+    if (!validation.isValid) {
+      throw new Error(
```
whoiskatrin: tiny thing, should this return a `ServiceResult` like the other file service methods?

i think plain errors here might skip the normal `ErrorCode` mapping, so a missing file could end up looking like a generic rpc error instead of `FILE_NOT_FOUND`

packages/sandbox/src/sandbox.ts
```diff
@@ -3468,11 +3469,48 @@ export class Sandbox<Env = unknown> extends Container<Env> implements ISandbox {
...
+   *   - `undefined` (default): auto-detect from MIME type (text → UTF-8 string, binary → base64 string)
+   *   - `'utf-8'` / `'utf8'`: always return as UTF-8 string
+   *   - `'base64'`: always return as base64-encoded string
+   *   - `'none'`: return a result whose `content` is a raw binary `ReadableStream<Uint8Array>`
+   *              with no encoding overhead. **Requires `SANDBOX_TRANSPORT=rpc`.** Throws on HTTP/WebSocket transports.
+   */
+  async readFile(
+    path: string,
+    options: { encoding: 'none'; sessionId?: string }
+  ): Promise<{
```
whoiskatrin: could we add the new `encoding: 'none'` overload to the shared types too?

looks like `Sandbox.readFile` has the stream return type now, but `ISandbox` / `ExecutionSession` still only say `ReadFileResult`, so callers typed through those lose the nice stream type

packages/sandbox/src/sandbox.ts
```diff
@@ -3468,11 +3469,48 @@ export class Sandbox<Env = unknown> extends Container<Env> implements ISandbox {
...
+    success: true;
+    content: ReadableStream<Uint8Array>;
```
whoiskatrin: nit, this result shape is showing up in a few places now

maybe worth making a shared type for the stream read result so the sdk + rpc types stay in sync?

packages/sandbox/src/clients/file-client.ts
```diff
@@ -102,11 +102,43 @@ export class FileClient extends BaseHttpClient {
...
+    content: ReadableStream<Uint8Array>;
+    path: string;
+    size: number;
+    mimeType: string;
+    timestamp: string;
+  }>;
   async readFile(
     path: string,
     sessionId: string,
     options?: { encoding?: string }
```
whoiskatrin: small type thought here, since this only works on rpc

if someone has options typed wider as `{ encoding?: string }`, ts might still think this is the normal `ReadFileResult` path even when they pass `none`

whoiskatrin (commented):
left a couple small thoughts

whoiskatrin (approved):

packages/sandbox-container/src/services/file-service.ts
aron-cf: Good catch, fixed.

packages/sandbox/src/sandbox.ts
aron-cf: I've refactored all the types for encoding so this is shared everywhere.

PR #691, opened 2026-05-12
Implement no credential R2 binding mount support

scuffi:
### Summary

This PR implements a new mount path for R2 bindings that never passes credentials into the container. Instead of supplying an S3-compatible endpoint and access keys, users pass a `bindingName` referencing an R2 binding on their Worker. The DO intercepts s3fs HTTP traffic at the Worker boundary, translates S3 API calls into native R2 binding calls, and proxies the result back — keeping credentials entirely in the Worker runtime.

### How it works

1. When `mountBucket()` detects no `endpoint` on the options, it takes the R2 egress path.
2. A one-time password file is written to `/tmp/.passwd-s3fs-<uuid>` inside the container (dummy credentials — s3fs requires them syntactically, but they are never validated by R2).
3. s3fs is launched pointing at `http://r2.internal/<bucket>` (not HTTPS, meaning `interceptHttps` is not required for this feature).
4. Outbound traffic to `r2.internal` is intercepted by the new R2 egress handler, which:
   - Validates the bucket name against an allowlist of mounted buckets.
   - Enforces `readOnly` by rejecting mutating operations with 403.
   - Translates S3 path-style operations (list, get, put, delete, multipart, copy, range) to native R2 binding API calls.
   - Returns S3-compatible XML responses that s3fs can parse.
5. On unmount or sandbox destroy, the password file and outbound handler registration are cleaned up.

<details>
<summary><h3>Speed comparison R2 binding vs S3 native</h3></summary>

_The R2 path does introduce a slight overhead for most operations (mostly negligible ~few ms), except for writes. We have a pretty noticeable overhead for write operations through R2 bindings (~1.5x-2x slower). I'm yet to find an approach to speed this up, so this will likely be come back to in future._

| Operation | R2 (ms) | S3 (ms) | Delta | Winner |
|---|---|---|---|---|
| Mount bucket | 2270 | 8406 | -6136ms | R2 |
| List directory (initial) | 560 | 594 | -34ms | R2 |
| Read seeded sample file | 145 | 158 | -13ms | R2 |
| Stat sample file (exec) | 42 | 54 | -12ms | R2 |
| Write tiny file (~150 B) | 1703 | 1430 | +273ms | S3 |
| Read tiny file back | 167 | 186 | -19ms | R2 |
| Write 1 KB file | 1788 | 1367 | +421ms | S3 |
| Read 1 KB file back | 361 | 366 | -5ms | R2 |
| Write 10 KB file | 1366 | 663 | +703ms | S3 |
| Read 10 KB file back | 420 | 273 | +147ms | S3 |
| Write 100 KB file | 1416 | 735 | +681ms | S3 |
| Read 100 KB file back | 324 | 283 | +41ms | S3 |
| Write 512 KB file | 1405 | 567 | +838ms | S3 |
| Read 512 KB file back | 944 | 382 | +562ms | S3 |
| Overwrite test: write V1 | 1626 | 1319 | +307ms | S3 |
| Overwrite test: replace with V2 | 1406 | 306 | +1100ms | S3 |
| Overwrite test: read back | 167 | 133 | +34ms | S3 |
| Concurrent write 3 × 2 KB files | 4340 | 2366 | +1974ms | S3 |
| Concurrent read 3 files back | 890 | 816 | +74ms | S3 |
| Check file exists | 99 | 82 | +17ms | S3 |
| List directory (post-write) | 534 | 704 | -170ms | R2 |
| Delete written files (exec rm) | 1952 | 1998 | -46ms | R2 |
| Unmount bucket | 173 | 160 | +13ms | S3 |

</details>

### New usage

**SDK (production — R2 binding, no credentials):**

```ts
await sandbox.mountBucket('MY_BUCKET', '/mnt/data');

// with options:
await sandbox.mountBucket('MY_BUCKET', '/mnt/data', {
  prefix: '/uploads',
  readOnly: true,
  s3fsOptions: ['parallel_count=16'],
});
```

**SDK (existing — remote S3/R2 endpoint with explicit credentials, unchanged):**

```ts
await sandbox.mountBucket('my-bucket', '/mnt/data', {
  endpoint: 'https://<account>.r2.cloudflarestorage.com',
  credentials: { accessKeyId: '...', secretAccessKey: '...' },
});
```

**Bridge HTTP API:**

```jsonc
// POST /v1/sandbox/:id/mount
{
  "bucket": "MY_BUCKET",       // binding name — no endpoint means R2 egress path
  "mountPath": "/mnt/data",
  "options": {
    "prefix": "/uploads",      // optional — must start with /
    "readOnly": false,
    "s3fsOptions": ["parallel_count=8"]
  }
}
```

### Bridge types

The old bridge-specific mount option types were replaced with a direct union alias over the SDK's own `MountBucketOptions`. This removes the impedance mismatch between the bridge API surface and the SDK:

- `MountBucketRequestOptions` is now a union of `RemoteMountBucketOptions | R2BindingMountBucketOptions | LocalMountBucketOptions` — the same union exported from the SDK package.
- Mount option validation and conversion to SDK types are centralised in `routes.ts` — all type narrowing happens once before `sandbox.mountBucket()` is called.
- Input validation now explicitly rejects: non-string `endpoint`, non-object or array `options`, non-string `credentials.accessKeyId` / `secretAccessKey`, non-boolean `readOnly`, non-string `prefix`.

### Files changed

| File                                                      | Change                                                                                                                                                                                               |
| -----------------------------------------------------------| ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| `packages/sandbox/src/storage-mount/r2-egress-handler.ts` | **New** — full R2 egress HTTP handler: list, get, put, delete, multipart, copy, range requests, read-only enforcement, S3-compatible XML serialisation                                               |
| `packages/sandbox/src/sandbox.ts`                         | New `mountBucketR2Egress()` path; password file lifecycle; outbound handler registration and cleanup; mountpoint check enforcement; unmount/destroy cleanup                                          |
| `packages/sandbox/src/bridge/types.ts`                    | Replaced bridge-specific mount option types with SDK union alias                                                                                                                                     |
| `packages/sandbox/src/bridge/routes.ts`                   | Centralised `validateMountOptions()` and `toSDKMountOptions()` helpers; tightened primitive type validation                                                                                          |
| `packages/sandbox/src/bridge/openapi.ts`                  | Updated `s3fsOptions` description to cover both remote and R2 mounts; fixed `prefix` description                                                                                                     |
| `packages/sandbox/src/storage-mount/validation.ts`        | Bucket name, binding name, and prefix validators; `isR2BindingMount` type guard                                                                                                                      |
| `packages/sandbox/src/storage-mount/types.ts`             | Added `R2EgressMountInfo` discriminant to the `MountInfo` union                                                                                                                                      |
| `packages/sandbox/src/storage-mount/index.ts`             | Re-exports for new types                                                                                                                                                                             |
| `packages/shared/src/types.ts`                            | Added `R2BindingMountBucketOptions`; updated `MountBucketOptions` union                                                                                                                              |
| `bridge/worker/README.md`                                 | Documented R2 binding mount, `s3fsOptions`, `prefix`, and `readOnly` in the mount API reference                                                                                                      |
| `packages/sandbox/tests/r2-egress-handler.test.ts`        | 28 unit tests for egress handler operations                                                                                                                                                          |
| `packages/sandbox/tests/r2-egress-mount.test.ts`          | 11 unit tests for mount lifecycle, prefix handling, failure cleanup, protected option rejection, and mountpoint check enforcement                                                                    |
| `packages/sandbox/tests/storage-mount/validation.test.ts` | 27 unit tests for the new validators                                                                                                                                                                 |
| `bridge/worker/src/__tests__/mount.test.ts`               | Extended with malformed credential, invalid primitive, and array options rejection tests                                                                                                             |
| `tests/e2e/bucket-mounting.test.ts`                       | E2E test: mount via R2 binding, verify pre-existing R2 file is visible in the container, write a file from the container, verify it appears in R2, then unmount and confirm mountpoint is cleaned up |

### DIFF
--- a/bridge/worker/README.md
+++ b/bridge/worker/README.md
@@ -261,13 +261,23 @@ curl -X POST http://localhost:8787/v1/sandbox/mfrggzdfmy2tqnrz/mount \
   -d '{"bucket": "my-bucket", "mountPath": "/mnt/data", "options": {"endpoint": "https://ACCT.r2.cloudflarestorage.com"}}'
 ```
 
+To mount a Worker R2 binding without credentials, omit `options.endpoint`. The `bucket`
+value must match the R2 binding name on the Worker:
+
+```sh
+curl -X POST http://localhost:8787/v1/sandbox/mfrggzdfmy2tqnrz/mount \
+  -H "Authorization: Bearer $SANDBOX_API_KEY" \
+  -H "Content-Type: application/json" \
+  -d '{"bucket": "MY_BUCKET", "mountPath": "/mnt/data", "options": {"prefix": "/uploads/"}}'
+```
+
 **Request body:**
 
 | Field                                 | Type    | Required | Description                                    |
 | ------------------------------------- | ------- | -------- | ---------------------------------------------- |
 | `bucket`                              | string  | yes      | Bucket name                                    |
 | `mountPath`                           | string  | yes      | Absolute path in the container to mount at     |
-| `options.endpoint`                    | string  | yes      | S3-compatible endpoint URL                     |
+| `options.endpoint`                    | string  | no       | S3-compatible endpoint URL                     |
 | `options.readOnly`                    | boolean | no       | Mount as read-only (default: false)            |
 | `options.prefix`                      | string  | no       | Subdirectory prefix within the bucket          |
 | `options.credentials.accessKeyId`     | string  | no       | Explicit access key (auto-detected if omitted) |
--- a/bridge/worker/README.md
+++ b/bridge/worker/README.md
@@ -271,14 +271,11 @@ curl -X POST http://localhost:8787/v1/sandbox/mfrggzdfmy2tqnrz/mount \
   -d '{"binding": "MY_BUCKET", "mountPath": "/mnt/data", "options": {"prefix": "/uploads/"}}'
 ```
 
-For compatibility, `bucket` is also accepted as the binding name when `binding`
-is omitted and `options.endpoint` is not provided.
-
 **Request body:**
 
 | Field                                 | Type    | Required | Description                                                                     |
 | ------------------------------------- | ------- | -------- | ------------------------------------------------------------------------------- |
-| `bucket`                              | string  | no       | Remote bucket name, or legacy R2 binding name when `binding` is omitted         |
+| `bucket`                              | string  | no       | Remote bucket name for endpoint-based S3-compatible mounts                      |
 | `binding`                             | string  | no       | Worker R2 binding name for credential-less R2 binding mounts                    |
 | `mountPath`                           | string  | yes      | Absolute path in the container to mount at                                      |
 | `options.endpoint`                    | string  | no       | S3-compatible endpoint URL for remote mounts; mutually exclusive with `binding` |
--- a/bridge/worker/README.md
+++ b/bridge/worker/README.md
@@ -261,27 +261,31 @@ curl -X POST http://localhost:8787/v1/sandbox/mfrggzdfmy2tqnrz/mount \
   -d '{"bucket": "my-bucket", "mountPath": "/mnt/data", "options": {"endpoint": "https://ACCT.r2.cloudflarestorage.com"}}'
 ```
 
-To mount a Worker R2 binding without credentials, omit `options.endpoint`. The `bucket`
-value must match the R2 binding name on the Worker:
+To mount a Worker R2 binding without credentials, provide the top-level `binding`
+field and omit `options.endpoint`:
 
 ```sh
 curl -X POST http://localhost:8787/v1/sandbox/mfrggzdfmy2tqnrz/mount \
   -H "Authorization: Bearer $SANDBOX_API_KEY" \
   -H "Content-Type: application/json" \
-  -d '{"bucket": "MY_BUCKET", "mountPath": "/mnt/data", "options": {"prefix": "/uploads/"}}'
+  -d '{"binding": "MY_BUCKET", "mountPath": "/mnt/data", "options": {"prefix": "/uploads/"}}'
 ```
+
+For compatibility, `bucket` is also accepted as the binding name when `binding`
+is omitted and `options.endpoint` is not provided.
 
 **Request body:**
 
-| Field                                 | Type    | Required | Description                                    |
-| ------------------------------------- | ------- | -------- | ---------------------------------------------- |
-| `bucket`                              | string  | yes      | Bucket name                                    |
-| `mountPath`                           | string  | yes      | Absolute path in the container to mount at     |
-| `options.endpoint`                    | string  | no       | S3-compatible endpoint URL                     |
-| `options.readOnly`                    | boolean | no       | Mount as read-only (default: false)            |
-| `options.prefix`                      | string  | no       | Subdirectory prefix within the bucket          |
-| `options.credentials.accessKeyId`     | string  | no       | Explicit access key (auto-detected if omitted) |
-| `options.credentials.secretAccessKey` | string  | no       | Explicit secret key (auto-detected if omitted) |
+| Field                                 | Type    | Required | Description                                                                     |
+| ------------------------------------- | ------- | -------- | ------------------------------------------------------------------------------- |
+| `bucket`                              | string  | no       | Remote bucket name, or legacy R2 binding name when `binding` is omitted         |
+| `binding`                             | string  | no       | Worker R2 binding name for credential-less R2 binding mounts                    |
+| `mountPath`                           | string  | yes      | Absolute path in the container to mount at                                      |
+| `options.endpoint`                    | string  | no       | S3-compatible endpoint URL for remote mounts; mutually exclusive with `binding` |
+| `options.readOnly`                    | boolean | no       | Mount as read-only (default: false)                                             |
+| `options.prefix`                      | string  | no       | Subdirectory prefix within the bucket                                           |
+| `options.credentials.accessKeyId`     | string  | no       | Explicit access key (auto-detected if omitted)                                  |
+| `options.credentials.secretAccessKey` | string  | no       | Explicit secret key (auto-detected if omitted)                                  |
 
 Credentials are optional — the SDK auto-detects from Worker secrets (`R2_ACCESS_KEY_ID`/`R2_SECRET_ACCESS_KEY` or `AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY`).
 
--- a/packages/sandbox/src/bridge/openapi.ts
+++ b/packages/sandbox/src/bridge/openapi.ts
@@ -98,11 +98,11 @@ export const OPENAPI_SCHEMA = {
       },
       MountBucketRequestOptions: {
         type: 'object',
-        required: ['endpoint'],
         properties: {
           endpoint: {
             type: 'string',
-            description: 'S3-compatible endpoint URL.',
+            description:
+              'S3-compatible endpoint URL. Omit to mount a Worker R2 binding with the same name as `bucket`.',
             example: 'https://YOUR_ACCOUNT_ID.r2.cloudflarestorage.com'
           },
           readOnly: {
@@ -120,6 +120,13 @@ export const OPENAPI_SCHEMA = {
             $ref: '#/components/schemas/MountBucketCredentials',
             description:
               'Explicit credentials. When omitted, the SDK auto-detects from Worker secrets (R2_ACCESS_KEY_ID / R2_SECRET_ACCESS_KEY or AWS equivalents).'
+          },
+          s3fsOptions: {
+            type: 'array',
+            items: { type: 'string' },
+            description:
+              'Advanced: Override or extend s3fs mount options. Only applies to R2 binding mounts (when endpoint is omitted).',
+            example: ['nomultipart']
           }
         }
       },
--- a/packages/sandbox/src/bridge/openapi.ts
+++ b/packages/sandbox/src/bridge/openapi.ts
@@ -125,7 +125,7 @@ export const OPENAPI_SCHEMA = {
             type: 'array',
             items: { type: 'string' },
             description:
-              'Advanced: Override or extend s3fs mount options. Only applies to R2 binding mounts (when endpoint is omitted).',
+              'Advanced: Override or extend s3fs mount options for both remote mounts and R2 binding mounts.',
             example: ['nomultipart']
           }
         }
--- a/packages/sandbox/src/bridge/openapi.ts
+++ b/packages/sandbox/src/bridge/openapi.ts
@@ -102,7 +102,7 @@ export const OPENAPI_SCHEMA = {
           endpoint: {
             type: 'string',
             description:
-              'S3-compatible endpoint URL. Omit to mount a Worker R2 binding with the same name as `bucket`.',
+              'S3-compatible endpoint URL for remote mounts. Mutually exclusive with top-level `binding`.',
             example: 'https://YOUR_ACCOUNT_ID.r2.cloudflarestorage.com'
           },
           readOnly: {
@@ -132,12 +132,19 @@ export const OPENAPI_SCHEMA = {
       },
       MountBucketRequest: {
         type: 'object',
-        required: ['bucket', 'mountPath', 'options'],
+        required: ['mountPath', 'options'],
         properties: {
           bucket: {
             type: 'string',
-            description: 'Bucket name.',
+            description:
+              'Remote bucket name. For compatibility, also used as the R2 binding name when `binding` is omitted.',
             example: 'my-r2-bucket'
+          },
+          binding: {
+            type: 'string',
+            description:
+              'Worker R2 binding name for credential-less R2 binding mounts. Mutually exclusive with `options.endpoint`.',
+            example: 'MY_BUCKET'
           },
           mountPath: {
             type: 'string',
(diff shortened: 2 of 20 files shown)


packages/sandbox/src/storage-mount/types.ts
```diff
@@ -27,3 +27,13 @@ export interface LocalSyncMountInfo {
   syncManager: LocalMountSyncManager;
   mounted: boolean;
 }
+
+export interface R2EgressMountInfo {
```
aron-cf:
```suggestion
export interface R2BindingMountInfo {
```

packages/sandbox/src/storage-mount/validation.ts
```diff
@@ -26,6 +47,26 @@ export function validateBucketName(bucket: string, mountPath: string): void {
...
+  if (bucketBinding.includes(':')) {
+    const [bucketName, prefixPart] = bucketBinding.split(':');
+    throw new InvalidMountConfigError(
+      `Bucket name cannot contain ':'. To mount a prefix, use the 'prefix' option:\n` +
+        `  mountBucket('${bucketName}', '${mountPath}', { ...options, prefix: '${prefixPart}' })`
+    );
+  }
+
+  const bindingNameRegex = /^[A-Za-z_][A-Za-z0-9_]*$/;
+  if (!bindingNameRegex.test(bucketBinding)) {
```
aron-cf, line 64: Is this defined somewhere external? Linking to it either in docs or code would be useful for maintainability.

packages/sandbox/src/sandbox.ts
aron-cf, line 1220: Is this out of date now?

packages/sandbox/src/sandbox.ts
```diff
@@ -1441,6 +1616,37 @@ export class Sandbox<Env = unknown> extends Container<Env> implements ISandbox {
...
+          );
+        }
+
+        try {
+          await this.execInternal(
+            `mountpoint -q ${shellEscape(mountPath)} || rmdir ${shellEscape(mountPath)}`
+          );
+        } catch {
+          // Best-effort directory removal
+        }
```
aron-cf: Theres a lot of duplication here between this and the fuse unmount below it. Can the deletion of the password file and teardown of egress be moved into that block?

packages/sandbox/src/sandbox.ts
```diff
@@ -5882,3 +6124,6 @@ export class Sandbox<Env = unknown> extends Container<Env> implements ISandbox {
     }
   }
 }
+
+Sandbox.outboundByHost = { 'r2.internal': r2EgressHandler };
+Sandbox.outboundHandlers = { r2EgressMount: r2EgressHandler };
```
aron-cf: Ooh, this is tricky, what happens if a user overrides this after reading our egress docs... ideally we shouldn't be touching the public surface area.

packages/sandbox/src/bridge/openapi.ts
```diff
@@ -98,11 +98,11 @@ export const OPENAPI_SCHEMA = {
       },
       MountBucketRequestOptions: {
         type: 'object',
-        required: ['endpoint'],
         properties: {
           endpoint: {
             type: 'string',
-            description: 'S3-compatible endpoint URL.',
+            description:
+              'S3-compatible endpoint URL. Omit to mount a Worker R2 binding with the same name as `bucket`.',
```
aron-cf: I think I'd prefer an explicit `binding` param here. And then the caller must provide either `endpoint` or `binding`

aron-cf (commented):
I'm halfway, but wanted to give some feedback while I review the egress handler. This is looking really good. I like the API and the structure.

I've got two major thoughts:

1. We need to figure out how to implement the egress handler without touching the public egress methods. Currently it's very easy for a user to wipe out the handler and break the implementation. Ideally we need to isolate our handler.
2. I think we need a clearer distinction between bucket and binding. Having both represented as strings isn't ideal. If you pass the binding directly into the mount/unmount functions it would remove the ambiguity. Are there benefits to also supporting the string I'm missing?

packages/sandbox/src/storage-mount/r2-egress-handler.ts
```diff
@@ -0,0 +1,738 @@
...
+  const contents = result.objects
+    .map(
+      (obj) =>
+        `<Contents>` +
+        `<Key>${escapeXml(obj.key)}</Key>` +
+        `<LastModified>${obj.uploaded.toISOString()}</LastModified>` +
+        `<ETag>${escapeXml(obj.httpEtag)}</ETag>` +
+        `<Size>${obj.size}</Size>` +
+        `<StorageClass>STANDARD</StorageClass>` +
+        `</Contents>`
```
aron-cf: Why is this mixing string concatenation with backtick strings? Would be clearer if it was just one big string no?

packages/sandbox/src/storage-mount/r2-egress-handler.ts
```diff
@@ -0,0 +1,738 @@
...
+};
+
+// ---------------------------------------------------------------------------
+// XML helpers
+// ---------------------------------------------------------------------------
+
+const XML_NS = 'xmlns="http://s3.amazonaws.com/doc/2006-03-01/"';
+const XML_DECL = '<?xml version="1.0" encoding="UTF-8"?>\n';
+
+function escapeXml(s: string): string {
```
aron-cf: This is a pendantic nit but I would really appreciate this being `escapeXML`

packages/sandbox/src/storage-mount/r2-egress-handler.ts
```diff
@@ -0,0 +1,738 @@
...
+): Promise<Response> {
+  const range = parseRange(request.headers.get('Range'));
+
+  if (!range) {
+    const obj = await r2.get(key);
+    if (!obj) {
+      return new Response(null, { status: 404 });
+    }
+    const body = await obj.arrayBuffer();
+    return new Response(body, {
```
aron-cf: @scuffi I've got the same concern as bonk here, how come we resolved the issue? Without streaming this is going to blow the DO up.

packages/sandbox/src/storage-mount/r2-egress-handler.ts
```diff
@@ -0,0 +1,738 @@
...
+    });
+  }
+
+  const [headObj, rangeObj] = await Promise.all([
+    r2.head(key),
+    r2.get(key, { range })
+  ]);
+  if (!headObj || !rangeObj) {
+    return new Response(null, { status: 404 });
+  }
```
aron-cf, line 475: This isn't clear what's going on here, why are we doing both in parallel? Needs a comment.

packages/sandbox/src/storage-mount/r2-egress-handler.ts
```diff
@@ -0,0 +1,738 @@
...
+
+  const rangeBody = await rangeObj.arrayBuffer();
```
aron-cf: Same concern as above, we want to be streaming here, not holding any of the data in memory.

packages/sandbox/src/storage-mount/r2-egress-handler.ts
```diff
@@ -0,0 +1,738 @@
...
+      return new Response(null, { status: 404 });
+    }
+
+    const metadataDirective = request.headers.get('x-amz-metadata-directive');
+    const httpMetadata =
+      metadataDirective?.toUpperCase() === 'REPLACE'
+        ? extractHttpMetadata(request)
+        : sourceObject.httpMetadata;
+    const sourceBody = await sourceObject.arrayBuffer();
+    const result = await r2.put(key, sourceBody, {
```
aron-cf: Again, let's stream here but this is painful that there isn't a copy/move api

packages/sandbox/src/storage-mount/r2-egress-handler.ts
aron-cf, line 1: Every call to `arrayBuffer()` needs to be checked here to see if it can stream instead. Otherwise we're going to consume a lot of CPU/memory buffering data and there's a real risk of OOM issues.

aron-cf (changes requested):
Okay, left some comments on the egress handler too.

packages/sandbox/src/sandbox.ts
```diff
@@ -146,6 +197,12 @@ const sandboxConfigurationCache = new WeakMap<
   Map<string, CachedSandboxConfiguration>
 >();
 
+const R2_DEFAULT_S3FS_OPTIONS: readonly string[] = [
+  'stat_cache_expire=60',
+  'enable_noobj_cache',
+  'multipart_size=5'
+];
```
aron-cf: In the container side I refactored this to handle options as an object. It makes it much easier to work with.

```
{
  stat_cache_expire: "60",
  enable_noobj_cache: true,
  multipart_size: "5",
}
```

Then just serialize the object when needed (`true` become single strings only, `false` is dropped).

packages/sandbox/src/sandbox.ts
```diff
@@ -1751,6 +1967,31 @@ export class Sandbox<Env = unknown> extends Container<Env> implements ISandbox {
...
+            } catch (error) {
+              mountFailures++;
+              const errorMsg =
+                error instanceof Error ? error.message : String(error);
+              this.logger.warn(
+                `Failed to unmount R2 egress bucket ${mountInfo.bucket} from ${mountPath}: ${errorMsg}`
+              );
+            }
+          }
+          await this.deletePasswordFile(mountInfo.passwordFilePath);
```
aron-cf: This still duplicates most of the following `else` block.

packages/sandbox/src/sandbox.ts
```diff
@@ -5881,4 +6129,46 @@ export class Sandbox<Env = unknown> extends Container<Env> implements ISandbox {
...
+            params
+          }
+        }
+      }
+    });
+    if (!isFetcher(fetcher)) {
+      throw new InvalidMountConfigError(
+        'R2 binding mounts require ContainerProxy to return a valid Fetcher'
+      );
+    }
```
aron-cf, line 6155: This is a nice fix!

packages/sandbox/src/sandbox.ts
```diff
@@ -5881,4 +6129,46 @@ export class Sandbox<Env = unknown> extends Container<Env> implements ISandbox {
       });
     }
   }
+
+  private async configureR2EgressOutbound(
+    params: R2EgressParams = this.getR2EgressParams()
```
aron-cf: This is a bit weird, I think I'd prefer it to always just be passed `params` for clarity at the call site.

packages/sandbox/src/sandbox.ts
```diff
@@ -5881,4 +6129,46 @@ export class Sandbox<Env = unknown> extends Container<Env> implements ISandbox {
...
+        'R2 binding mounts require ContainerProxy to return a valid Fetcher'
+      );
+    }
+
+    await ctx.container.interceptOutboundHttp('r2.internal', fetcher);
+  }
+
+  private async teardownR2EgressMount(): Promise<void> {
+    const remainingR2Params = this.getR2EgressParams();
+    await this.configureR2EgressOutbound(remainingR2Params);
```
aron-cf: This is equivalent to just calling `configureR2EgressOutbound()` no?

packages/sandbox/src/storage-mount/r2-egress-handler.ts
```diff
@@ -0,0 +1,806 @@
...
+): Promise<Response> {
+  const uploadId = url.searchParams.get('uploadId') ?? '';
+  const partNumber = parseInt(url.searchParams.get('partNumber') ?? '0', 10);
+  if (!uploadId || !partNumber) {
+    return new Response('Bad Request: missing uploadId or partNumber', {
+      status: 400
+    });
+  }
+  const upload = r2.resumeMultipartUpload(key, uploadId);
+  const body = await readRequestBody(request);
```
aron-cf: This can't do this. Can we not make this work with streams? As we're acting as a proxy, I'm assuming the request headers must contain the data we need about content length and etag etc?

bridge/worker/README.md
```diff
@@ -261,17 +261,31 @@ curl -X POST http://localhost:8787/v1/sandbox/mfrggzdfmy2tqnrz/mount \
...
+
+```sh
+curl -X POST http://localhost:8787/v1/sandbox/mfrggzdfmy2tqnrz/mount \
+  -H "Authorization: Bearer $SANDBOX_API_KEY" \
+  -H "Content-Type: application/json" \
+  -d '{"binding": "MY_BUCKET", "mountPath": "/mnt/data", "options": {"prefix": "/uploads/"}}'
+```
+
+For compatibility, `bucket` is also accepted as the binding name when `binding`
+is omitted and `options.endpoint` is not provided.
```
aron-cf: We don't need compatibility, this is a new feature.

aron-cf (changes requested):
This is getting really close. Well done fixing that egress issue. It looks like some tests are failing now though.

packages/sandbox/src/sandbox.ts
whoiskatrin: really like this shape! (re: object)

aron-cf (approved):
Wow, well done for powering through this 👏 . I think my main concerns have now been addressed. The egress handler isn't buffering anymore and the bridge API clarifies the binding. I'm still not 100% sure we should allow mountBucket to take strings that represent both bindings and buckets but for now I think the endpoint requirement for buckets is okay.

PR #695, opened 2026-05-13
Add support for Cloudflare tunnels

aron-cf:
Add `sandbox.tunnels` namespace with quick-tunnel support. Call `sandbox.tunnels.create(port)` to spawn a `cloudflared` quick tunnel inside the sandbox and get back a `https://<words>.trycloudflare.com` URL that proxies to `localhost:<port>` inside the container. No Cloudflare account or DNS setup required.

```ts
const tunnel = await sandbox.tunnels.create(8080);
console.log(tunnel.url);
// → https://random-words-here.trycloudflare.com

await sandbox.tunnels.list();
await sandbox.tunnels.destroy(tunnel);
```

This PR supercedes #674 which has a fuller implementation. The plan is to ship the tunnels interface in smaller chunks, starting with the basic "quick" tunnels that provide ephemeral URLs to single ports running inside the container.

> [!NOTE]
> Quick tunnels require the RPC transport and the glibc sandbox image variants (default, python, opencode, desktop) — the musl/Alpine variant does not include cloudflared.

**Limitations**

- **No SSE.** The trycloudflare edge buffers `text/event-stream` responses, so server-sent events don't reach the client. WebSockets work fine. Named tunnels don't have this issue.
- **No uptime guarantee.** Cloudflare positions trycloudflare as a debug aid, not a production target.
- **No persistent hostname.** Every restart picks a new `<words>.trycloudflare.com`.

**Background**

Today's `exposePort()` uses a Worker-fronted preview URL: traffic hits the Worker, the Worker calls into the Durable Object, the DO routes to the container. That's a great fit for HTTP request/response, but it has real ceilings for streaming / long-lived workloads:

- **Wildcard DNS.** Preview URLs encode the port + sandbox + token in the subdomain (`8080-sandbox-id-token.example.com`), which requires a wildcard CNAME *and* a custom domain (no `.workers.dev`).
- **Local Dev** The shared origin in local development means that tools like vite often accidentally intercept traffic intended for the sandbox.
- **CPU time per request.** A Worker invocation is bounded by your plan's CPU limit. A long-lived WebSocket or a multi-minute SSE stream can run into that ceiling well before the workload would naturally end.
- **Subrequest count.** Each Worker invocation has a hard subrequest cap. Anything that fans out (one request → many container calls) hits it.
- **Token-based auth on every request.** Adds latency on the hot path.

A Cloudflare Tunnel sidesteps all of those. Traffic terminates at the CF edge, gets routed straight into a `cloudflared` process inside the container over a persistent QUIC/HTTP2 connection, and never hits your Worker. Your Worker's CPU and subrequest budgets aren't touched. The hostname is whatever you point at the tunnel — no wildcard required.

**Overview**

This PR is broken into a couple of core pieces:

 1. We now install `cloudflared` in all non-musl images. There are currently no pre-built Alpine/musl binaries so this will be supported in a follow-up PR.
 2. The sandbox app in the container gained a `TunnelService` that handles spawning the `cloudflared` process and extracting the URL.
 3. The `Sandbox` interface now has a `tunnels` object with `create()/list()/destroy()` methods.

**API**

```ts
const tunnel = await sandbox.tunnels.create(8080);
// → https://<random-words>.trycloudflare.com

await sandbox.tunnels.list();
await sandbox.tunnels.destroy(tunnel);
```

**Testing**

There are unit tests for the TunnelService and E2E tests covering creating one & multiple tunnels.

### DIFF
--- a/packages/sandbox/Dockerfile
+++ b/packages/sandbox/Dockerfile
@@ -152,6 +152,38 @@ COPY --from=builder /app/packages/sandbox-container/dist/runtime/executors/javas
 # Users with custom startup scripts that call `bun /container-server/dist/index.js` need this
 COPY --from=builder /app/packages/sandbox-container/dist/index.js ./dist/
 
+# ============================================================================
+# cloudflared (Cloudflare Tunnel daemon) — enables sandbox.tunnels.*
+# Pinned version with sha256 verification per architecture.
+# Skipped in the musl/Alpine stage (no musl prebuilt).
+# ============================================================================
+ARG CLOUDFLARED_VERSION=2026.3.0
+ARG CLOUDFLARED_SHA256_AMD64=4a9e50e6d6d798e90fcd01933151a90bf7edd99a0a55c28ad18f2e16263a5c30
+ARG CLOUDFLARED_SHA256_ARM64=0755ba4cbab59980e6148367fcf53a8f3ec85a97deefd63c2420cf7850769bee
+RUN set -eux; \
+    arch="$(dpkg --print-architecture)"; \
+    case "$arch" in \
+      amd64) suffix=amd64; sha="${CLOUDFLARED_SHA256_AMD64}" ;; \
+      arm64) suffix=arm64; sha="${CLOUDFLARED_SHA256_ARM64}" ;; \
+      *) echo "Unsupported arch for cloudflared: $arch" >&2; exit 1 ;; \
+    esac; \
+    url="https://github.com/cloudflare/cloudflared/releases/download/${CLOUDFLARED_VERSION}/cloudflared-linux-${suffix}"; \
+    curl -fsSL -o /usr/local/bin/cloudflared "$url"; \
+    echo "$sha  /usr/local/bin/cloudflared" | sha256sum -c -; \
+    chmod +x /usr/local/bin/cloudflared; \
+    cloudflared --version
+
+# Inject the host's extra CA bundle when running locally behind a TLS-
+# intercepting proxy (e.g. Cloudflare WARP / Zero Trust). See
+# DOCKER_README.md for the local-dev setup. No-op when the secret
+# isn't passed.
+RUN --mount=type=secret,id=wrangler_ca \
+    if [ -f /run/secrets/wrangler_ca ] && [ -s /run/secrets/wrangler_ca ]; then \
+        cp /run/secrets/wrangler_ca /usr/local/share/ca-certificates/wrangler-dev-ca.crt && \
+        cat /run/secrets/wrangler_ca >> /etc/ssl/certs/ca-certificates.crt && \
+        update-ca-certificates; \
+    fi
+
 RUN mkdir -p /workspace
 
 # Expose the application port (3000 for control)
--- a/packages/sandbox/README.md
+++ b/packages/sandbox/README.md
@@ -98,6 +98,40 @@ export default {
 };
 ```
 
+## Quick tunnels
+
+`sandbox.tunnels.create(port)` exposes a service running inside the
+sandbox on a randomly-assigned `*.trycloudflare.com` URL. No Cloudflare
+account or DNS setup required — cloudflared opens a persistent QUIC
+connection to Cloudflare's edge and Cloudflare hands back a hostname.
+
+```ts
+// Inside a Worker with an RPC-transport sandbox:
+const tunnel = await sandbox.tunnels.create(8080);
+console.log(tunnel.url);
+// → https://random-words-here.trycloudflare.com
+
+// Later:
+await sandbox.tunnels.destroy(tunnel);
+```
+
+Notes:
+
+- Requires the RPC transport. The route-based transport's `tunnels`
+  stub throws "RPC transport required".
+- Hostnames are assigned per `create()` call — restarting a sandbox or
+  calling `create()` twice yields different URLs.
+- The first fetch can take a few seconds while DNS propagates, even
+  after `create()` resolves.
+- `*.trycloudflare.com` buffers `text/event-stream` responses. WebSockets
+  work fine.
+- The musl/Alpine image variant does not ship cloudflared (no upstream
+  musl prebuilt); `sandbox.tunnels` is unavailable on that variant.
+- Local builds behind a TLS-intercepting proxy (e.g. Cloudflare WARP)
+  need the host CA bundle injected at build time — see
+  [DOCKER_README.md](../../DOCKER_README.md).
+
+
 ## Documentation
 
 **📖 [Full Documentation](https://developers.cloudflare.com/sandbox/)**
@@ -115,6 +149,7 @@ export default {
 - **File System Access** - Read, write, and manage files
 - **Command Execution** - Run any command with streaming support
 - **Preview URLs** - Expose services with public URLs
+- **Quick tunnels** - Zero-config `*.trycloudflare.com` URLs via `sandbox.tunnels.create(port)`
 - **Git Integration** - Clone repositories directly
 
 ## Contributing
--- a/packages/sandbox/README.md
+++ b/packages/sandbox/README.md
@@ -100,37 +100,48 @@ export default {
 
 ## Quick tunnels
 
-`sandbox.tunnels.create(port)` exposes a service running inside the
-sandbox on a randomly-assigned `*.trycloudflare.com` URL. No Cloudflare
-account or DNS setup required — cloudflared opens a persistent QUIC
-connection to Cloudflare's edge and Cloudflare hands back a hostname.
+`sandbox.tunnels.get(port)` exposes a service running inside the
+sandbox on a `*.trycloudflare.com` URL. No Cloudflare account or DNS
+setup required — cloudflared opens a persistent QUIC connection to
+Cloudflare's edge and Cloudflare hands back a hostname.
 
 ```ts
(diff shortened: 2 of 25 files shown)


packages/sandbox/src/tunnels/tunnels-handler.ts
```diff
@@ -0,0 +1,97 @@
...
+    } finally {
+      logCanonicalEvent(host.logger, {
+        event: 'tunnel.create',
+        outcome,
+        port,
+        durationMs: Date.now() - startTime,
+        error: caughtError
+      });
+    }
+  }
```
aron-cf: @ghostwriternr interested in your thoughts here. I think we're probably going to see this called on the hot path in every worker call. So I'm wondering if this should actually be an idempotent getOrCreate(). But if that's the case do we need to cache the URL somewhere to avoid hitting the container each time?

ghostwriternr: yet to go over the pr overall but i agree on having this be `getOrCreate`. the method could even just be `.get`, which does a `get` if exits or `create` if not, so user don't have to spend too much time reasoning about this and just always do `handler.get(8080)`?

aron-cf: I guess then we store the tunnel info in storage on the DO and wipe it when there's a new container instance?

packages/sandbox/Dockerfile
```diff
@@ -152,6 +152,38 @@ COPY --from=builder /app/packages/sandbox-container/dist/runtime/executors/javas
...
 COPY --from=builder /app/packages/sandbox-container/dist/index.js ./dist/
 
+# ============================================================================
+# cloudflared (Cloudflare Tunnel daemon) — enables sandbox.tunnels.*
+# Pinned version with sha256 verification per architecture.
+# Skipped in the musl/Alpine stage (no musl prebuilt).
+# ============================================================================
+ARG CLOUDFLARED_VERSION=2026.3.0
+ARG CLOUDFLARED_SHA256_AMD64=4a9e50e6d6d798e90fcd01933151a90bf7edd99a0a55c28ad18f2e16263a5c30
+ARG CLOUDFLARED_SHA256_ARM64=0755ba4cbab59980e6148367fcf53a8f3ec85a97deefd63c2420cf7850769bee
```
ghostwriternr, line 162: no-op: nice, i was just scrolling to find this, considering we're string matching to detect url etc.

packages/sandbox/src/tunnels/tunnels-handler.ts
ghostwriternr: yep!

packages/sandbox/src/sandbox.ts
```diff
@@ -3864,6 +3872,29 @@ export class Sandbox<Env = unknown> extends Container<Env> implements ISandbox {
...
+   * - `tunnels.create(port)` — spawn cloudflared, return a
+   *   `https://<words>.trycloudflare.com` URL pointing at `localhost:<port>`.
+   * - `tunnels.list()` — tunnels currently running in the container.
+   * - `tunnels.destroy(idOrInfo)` — tear down by id or by the record
+   *   returned from `create()`.
+   *
+   * Requires the RPC transport. Calling this on a route-based transport
+   * throws "RPC transport required".
+   */
+  get tunnels(): TunnelsHandler {
```
ghostwriternr, line 3949: no-op: this is indeed an intentional choice, and its right that the tunnels namespace is only accessible from the sandbox stub and not the session object, considering it's a sandbox-level operation

packages/sandbox/README.md
```diff
@@ -98,6 +98,40 @@ export default {
 };
 ```
 
+## Quick tunnels
```
ghostwriternr, line 101: nit: feel like we can leave this for the docs, considering the readme is kept super-lean here

ghostwriternr (commented):
super minor comments so far - looking v nice otherwise. the only open item to discuss is persistence/invalidation across containers stops & restarts. right now, the state is mainly on the container side which would be wiped on stop.

packages/sandbox/src/tunnels/tunnels-handler.ts
```diff
@@ -0,0 +1,216 @@
...
+
+      const work = (async (): Promise<TunnelInfo> => {
+        const map = await readMap();
+        const existing = map[port.toString()];
+        if (existing) {
+          cacheState = 'hit';
+          return existing;
+        }
+        const id = `quick-${shortId()}`;
+        const info = await host.client.tunnels.runQuickTunnel(id, port);
```
whoiskatrin: can `get()` and `destroy()` race on the same port here? like one request is creating or storing it while another is deleting it

packages/sandbox/src/tunnels/tunnels-handler.ts
```diff
@@ -0,0 +1,216 @@
...
+        outcome,
+        port,
+        cacheState,
+        durationMs: Date.now() - startTime,
+        error: caughtError
+      });
+    }
+  }
+
+  async function destroy(portOrInfo: number | TunnelInfo): Promise<void> {
```
whoiskatrin, line 182: for `destroy(tunnelInfo)`, should we compare the id too? wondering if an old tunnel object could accidentally delete a newer tunnel on the same port

packages/sandbox/src/tunnels/tunnels-handler.ts
```diff
@@ -0,0 +1,216 @@
...
+      // Atomic clear: same race window as get(), since destroyTunnel()
+      // also yields. Wrap the read-modify-write so a concurrent
+      // get(otherPort) write doesn't resurrect this entry. Storage is
+      // cleared *before* the container RPC for the same reason as
+      // portTokens (sandbox.ts:1795): a get(port) that races our destroy
+      // should see a cache miss and spawn fresh rather than reuse the
+      // record we're tearing down.
+      await host.storage.transaction(async (txn) => {
+        const current = (await txn.get<TunnelMap>(STORAGE_KEY)) ?? {};
+        delete current[port.toString()];
```
whoiskatrin: is it okay to remove the stored tunnel before `destroyTunnel()` finishes? i was wondering what happens if that call errors or times out, do we lose the record even though the tunnel might still be running?

packages/sandbox/src/tunnels/tunnels-handler.ts
```diff
@@ -0,0 +1,216 @@
...
+        outcome,
+        port,
+        tunnelId,
+        durationMs: Date.now() - startTime,
+        error: caughtError
+      });
+    }
+  }
+
+  async function list(): Promise<TunnelInfo[]> {
```
whoiskatrin, line 234: if `cloudflared` dies after startup, do `get()` and `list()` keep returning the cached record anyway? not sure if anything clears that state today

aron-cf: Yeah, this is a great shout. They will. I've got a note to add an event handler on the process to tear down the port if the process dies. I can take a look at that and see if it's a quick fix.

packages/sandbox/src/tunnels/tunnels-handler.ts
aron-cf: They can. I think probably the way to solve that is to have an operation queue per port. I'll have a think about it.

packages/sandbox/src/tunnels/tunnels-handler.ts
aron-cf: There can only ever be one tunnel per port. The id exists mostly for tracking the lifecycle in logging etc. I think that's probably the right level of granularity for now. If you're calling destroy you want to tear down the port. I don't think I would want it to no-op because the tunnel id cycled.

whoiskatrin: Cool, that's mainly what I wanted to know

packages/sandbox/src/tunnels/tunnels-handler.ts
aron-cf: Yeah, good shotu. I'm going to look at `transaction()` and see if we can put the create/destroy calls in there. I think that would give us the lock behavior we want. Otherwise the queue I mentioned in the other thread might be an option too.

aron-cf: Fixed in [c756973](https://github.com/cloudflare/sandbox-sdk/pull/695/commits/c7569737572b115cfa9bac328139e73a0343e067), transaction wasn't a great fit in the end. I've gone with a simple queue.

packages/sandbox/src/tunnels/tunnels-handler.ts
aron-cf: Fixed in [c756973](https://github.com/cloudflare/sandbox-sdk/pull/695/commits/c7569737572b115cfa9bac328139e73a0343e067), transaction wasn't a great fit in the end. I've gone with a simple queue.

packages/sandbox/src/tunnels/tunnels-handler.ts
aron-cf: Fixed in [c756973](https://github.com/cloudflare/sandbox-sdk/pull/695/commits/c7569737572b115cfa9bac328139e73a0343e067). We now callback from the container to the DO when a cloudflared process exits. It made the code a bit more verbose but I think it's a good pattern for us to start using.

whoiskatrin (approved):

PR #698, opened 2026-05-15
Recover RPC stream after OOM error

aron-cf:
When several RPC calls land concurrently on the `rpc` transport and the proxied container WebSocket fails — for example, under load from a few large `writeFile` payloads — every in-flight RPC rejects with `RPCTransportError: Peer closed WebSocket: 1011 ...`.

The Durable Object isolate stays wedged after that failure. Every subsequent RPC on the same isolate rejects with the same cached error, and only runtime eviction (idle or otherwise) recovers it.

## Root cause

Recovery in `ContainerControlClient` is driven by `setInterval(pollBusyState, 1000)`. The poller is the only path that calls `destroyConnection()` to null out the broken `ContainerControlConnection` (and with it the broken `DeferredTransport` / `RpcSession`) so the next `getConnection()` builds a fresh one. 

In a DO, `setInterval` callbacks don't fire while the isolate sits idle between requests — which is exactly the state the isolate enters the instant the failing batch returns its 500s. The poller never ticks again, `this.conn` keeps pointing at the dead connection, and `DeferredTransport.#error` is sticky for the lifetime of the transport, so every reuse of the cached stub throws synchronously without doing any I/O.

The wedge reproduces locally with `wrangler dev --var SANDBOX_TRANSPORT:rpc` and five concurrent `writeFile` calls sized just under the 32 MiB JSRPC argument limit. The Docker container itself stays healthy throughout; the failure is purely in-isolate state.

## Fix

Make failure recovery event-driven instead of polled. `ContainerControlConnection` now accepts an `onClose` option that fires from the WebSocket `close` and `error` listeners installed in `doConnect`, but only when an already-active connection transitions to disconnected. 

`ContainerControlClient` passes `onClose: () => this.destroyConnection()`, so the broken connection is torn down inside the same turn of the event loop as the WS close. The very next call to `getConnection()` builds a fresh `ContainerControlConnection` and the isolate recovers without depending on a timer firing. The existing busy-poll path is unchanged for idle-disconnect; only the failure-recovery branch moved.

The `wasEverConnected` flag in `ContainerControlClient` existed solely to distinguish "WS upgrade still in progress" from "peer disconnected mid-session" inside `pollBusyState`. Now that the disconnected case is handled synchronously via `onClose`, the flag is dead state: by the time the poller could observe `!isConnected()`, `this.conn` has already been nulled. So the field has now been removed.

## Verification

A new `tests/e2e/rpc-recovery-after-ws-close.test.ts` adds a regression test that reproduces the issue against main. The new code demonstrates the fix.

### DIFF
--- a/packages/sandbox/src/container-control/client.ts
+++ b/packages/sandbox/src/container-control/client.ts
@@ -394,9 +394,19 @@ export class ContainerControlClient {
     this.connOptions = {
       stub: options.stub,
       port: options.port,
+      localMain: options.localMain,
       logger: options.logger,
       retryTimeoutMs: options.retryTimeoutMs,
-      localMain: options.localMain
+      // Event-driven failure recovery: when the live WebSocket closes
+      // or errors, tear the connection down inside the same turn of
+      // the event loop so the next RPC call builds a fresh one. The
+      // 1Hz busy-poll fallback can't be relied on here — `setInterval`
+      // callbacks don't fire while the DO isolate sits idle between
+      // requests, which is exactly the state the isolate enters after
+      // every in-flight RPC rejects with a peer-closed error.
+      onClose: () => {
+        if (this.conn) this.destroyConnection();
+      }
     };
     this.idleDisconnectMs =
       options.idleDisconnectMs ?? DEFAULT_IDLE_DISCONNECT_MS;
--- a/packages/sandbox/src/container-control/client.ts
+++ b/packages/sandbox/src/container-control/client.ts
@@ -382,13 +382,6 @@ export class ContainerControlClient {
   private busyPollTimer: ReturnType<typeof setInterval> | null = null;
   /** Tracks whether we currently believe the session is busy. */
   private busy = false;
-  /**
-   * Set the first time the poller observes `conn.isConnected() === true`,
-   * cleared in `destroyConnection()`. Lets us distinguish "the WebSocket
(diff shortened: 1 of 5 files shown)


packages/sandbox/src/container-control/client.ts
```diff
@@ -381,20 +381,23 @@ export class ContainerControlClient {
...
+      // 1Hz busy-poll fallback can't be relied on here — `setInterval`
+      // callbacks don't fire while the DO isolate sits idle between
+      // requests, which is exactly the state the isolate enters after
+      // every in-flight RPC rejects with a peer-closed error.
+      onClose: () => {
+        if (this.conn) this.destroyConnection();
+      }
     };
     this.idleDisconnectMs =
       options.idleDisconnectMs ?? DEFAULT_IDLE_DISCONNECT_MS;
```
whoiskatrin, line 405: institute of dumb questions here, what happens if an old connection finishes after disconnect then closes later could it tear down the new connection too?

whoiskatrin (approved):

packages/sandbox/src/container-control/client.ts
aron-cf: Not gonna lie, it has broken my head trying to figure out if this can be triggered. In the end I just ended up unbinding the close listener on disconnect in 546ad17, this should mean that a staggered disconnect can't end up affecting a newer socket.

PR #703, opened 2026-05-18
Fix inconsistent interfaces between HTTP and RPC clients

aron-cf:
- **Implement SandboxAPI interfaces on HTTP sub-clients**
  Each HTTP client now declares implements on its corresponding
  SandboxAPI interface, so signature mismatches between the HTTP
  and RPC transports are caught at compile time.
  

- **Fix screenshotRegion signature to match SandboxDesktopAPI**
  The interface expects a single DesktopScreenshotRegionRequest object
  with region nested inside it. The HTTP client was taking region and
  options as separate arguments.
  
  The HTTP-only 'bytes' format overload is dropped from the public
  signature since it was never part of SandboxDesktopAPI and could
  not be used via the RPC transport.
  

- **Fix type() options to use delay instead of delayMs**
  TypeOptions had delayMs but SandboxDesktopAPI specifies
  { delay?: number }. Rename to align the two transports.
  

- **Fix getProcessStatus return type across all layers**
  The method returns per-process health (DesktopProcessHealth), not the
  full desktop status. The RPC stub was incorrectly delegating to
  status(), ignoring the name argument entirely. Fix the shared
  interface, the HTTP client, and the RPC stub to all return
  DesktopProcessHealth.
  

- **Stub listSessions on HTTP client as RPC-only**
  Satisfies SandboxUtilsAPI while making the transport
  requirement explicit at call time.

### DIFF
--- a/packages/sandbox/src/clients/desktop-client.ts
+++ b/packages/sandbox/src/clients/desktop-client.ts
@@ -1,4 +1,5 @@
 import type {
+  DesktopProcessHealth,
   DesktopScreenshotRegionRequest,
   DesktopScreenshotRequest,
   SandboxDesktopAPI
@@ -117,11 +118,7 @@ export interface Desktop {
   keyDown(key: KeyInput): Promise<void>;
   keyUp(key: KeyInput): Promise<void>;
   getScreenSize(): Promise<ScreenSizeResponse>;
-  getProcessStatus(
-    name: string
-  ): Promise<
-    BaseApiResponse & { running: boolean; pid?: number; uptime?: number }
-  >;
+  getProcessStatus(name: string): Promise<DesktopProcessHealth>;
 }
 
 /**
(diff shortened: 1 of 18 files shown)


packages/sandbox/src/clients/desktop-client.ts
```diff
@@ -220,78 +195,31 @@ export class DesktopClient extends BaseHttpClient {
...
-
-    return response;
+    return this.post<ScreenshotResponse>('/api/desktop/screenshot', data);
   }
 
   /**
-   * Capture a region screenshot as base64 (default).
+   * Capture a region screenshot as base64.
    */
   async screenshotRegion(
```
whoiskatrin, line 260: could this be a breaking change for existing callers? looks like `screenshotRegion(region, options)` and `{ format: 'bytes' }` worked before, and older js users wouldnt get a type error if that changed, theyd just start failing or get base64 back instead of bytes. is that expected or should we keep some kind of compat shim?

packages/sandbox/src/clients/desktop-client.ts
```diff
@@ -447,7 +375,7 @@ export class DesktopClient extends BaseHttpClient {
   async type(text: string, options?: TypeOptions): Promise<void> {
     await this.post<BaseApiResponse>('/api/desktop/keyboard/type', {
       text,
-      ...(options?.delayMs !== undefined && { delayMs: options.delayMs })
+      ...(options?.delay !== undefined && { delay: options.delay })
```
whoiskatrin: i noticed this starts sending `delay`, but `DesktopTypeRequest` in shared still has `delayMs`. i may be missing how the desktop side wires this up, but does that leave the http and container request shapes out of sync here?

packages/sandbox/src/clients/desktop-client.ts
aron-cf: 🤦 my bad, sorry I didn't review this as thoroughly as I should have done.

packages/sandbox/src/clients/desktop-client.ts
aron-cf: Fixed in [2d442fd](https://github.com/cloudflare/sandbox-sdk/pull/703/commits/2d442fd5b30d26a3f873c786a018041dbe744d65)

whoiskatrin (approved):

PR #706, opened 2026-05-19
Implement sessionless execution mode

scuffi:
## Summary

This PR makes sessionless execution a first-class, explicit execution mode across the SDK and container runtime. It introduces a canonical sessionless sentinel, adds `enableDefaultSession` so implicit execution can use either the default session or sessionless mode, and refactors container-side execution behind a unified `ExecutionService` so behavior is more consistent across commands, files, git, backup, and process operations.

## Key Changes

- adds shared sessionless execution types and exports, including `SESSIONLESS_SESSION_ID = 'none'`
- updates `Sandbox` execution-policy resolution and persists `enableDefaultSession` as sandbox configuration
- routes container execution through a unified `ExecutionService`
- updates file, git, backup, and process internals to use the new execution path
- tightens session propagation in the E2E test worker
- adds the user-facing changeset for the new execution policy

## Testing Added

Added and updated unit tests across the SDK and container runtime, plus lean end-to-end coverage in `tests/e2e/sessionless-execution-workflow.test.ts` and `tests/e2e/test-worker/index.ts` to validate default-session behavior, isolated/sessionless execution, explicit per-call session overrides, and `execStream` behavior under the new execution policy.


packages/shared/src/types.ts
```diff
@@ -439,6 +448,15 @@ export interface SessionOptions {
...
 
 // Sandbox configuration options
+/**
+ * Special session ID that forces a command or operation to run without a
+ * persistent shell session.
+ *
+ * Pass this as `sessionId` on supported APIs to opt a single operation into
+ * sessionless execution even when default sessions are enabled.
+ */
+export const SESSIONLESS_SESSION_ID = 'none';
```
aron-cf: can a user provide a session id or is it always a random string? If the former we should probably make this a little more obscure...

scuffi: a user can provide a session ID, but more explicitly they're meant to pass this (`none`) if they want a specific function call to use sessionless execution if default sessions are still enabled

aron-cf: I have a big concern about supporting a mixture. I would much rather you opt out when calling getSandbox() and all top level APIs are session less. I don't see a benefit to mixing/matching.

.changeset/sessionless-execution.md
```diff
@@ -0,0 +1,7 @@
+---
+'@cloudflare/sandbox': patch
+---
+
+Add sessionless execution mode with a configurable default-session policy.
+
+Set `enableDefaultSession: false` in `SandboxOptions` (or call `sandbox.setEnableDefaultSession(false)`) to run implicit operations without a persistent shell — each command gets a fresh process with no shared state. Use the new `SESSIONLESS_SESSION_ID` (`'none'`) constant as `sessionId` on individual calls to opt a single operation into sessionless mode without changing the sandbox-level policy.
```
aron-cf: Can we simplify and remove `setEnableDefaultSession` and `SESSIONLESS_SESSION_ID`, I don't think there's a clear benefit. We're trying to move away from the default session entirely so having a mixed mode doesn't give us a clean break.

packages/sandbox-container/src/services/execution-service.ts
aron-cf, line 1: I like this implementation.

packages/sandbox-container/tests/handlers/file-handler.test.ts
```diff
@@ -268,7 +268,8 @@ describe('FileHandler', () => {
   describe('handleDelete - POST /api/delete', () => {
     it('should delete file successfully', async () => {
       const deleteFileData = {
-        path: '/tmp/delete-me.txt'
+        path: '/tmp/delete-me.txt',
+        sessionId: 'none'
```
aron-cf: Can we use constants here instead of the string "none".

aron-cf (changes requested):
This is looking really good. I like the refactor to use an execution service. My main request is that we remove the ability to mix session and sessionless calls. If you want to chat about this in person we can.

tests/e2e/sessionless-execution-workflow.test.ts
```diff
@@ -0,0 +1,186 @@
...
+      .join('');
+    const complete = events.find((event) => event.type === 'complete');
+    const error = events.find((event) => event.type === 'error');
+
+    expect(stdout).toBe('missing');
+    expect(complete).toBeDefined();
+    expect(error).toBeUndefined();
+  }, 90000);
+
+  test('should reject explicit session IDs when default sessions are disabled', async () => {
```
aron-cf: Ah, maybe my last comment wasn't precise enough. I meant a user shouldn't be able to use the "default" session id when disabling default sessions. I'm not sure we should block the sessionId entirely, though I could be convinced that we should.

tests/e2e/test-worker/index.ts
```diff
@@ -544,11 +550,17 @@ console.log('Terminal server on port ' + port);
...
+          ? await executor.execStream(body.command, {
+              env: body.env,
+              cwd: body.cwd,
+              timeout: body.timeout
+            })
+          : await sandbox.execStream(body.command, {
+              env: body.env,
+              cwd: body.cwd,
+              timeout: body.timeout
+            });
```
aron-cf: What is this ternary doing?

packages/sandbox/src/sandbox.ts
```diff
@@ -568,6 +587,7 @@ export class Sandbox<Env = unknown> extends Container<Env> implements ISandbox {
   envVars: Record<string, string> = {};
   private logger: ReturnType<typeof createLogger>;
   private keepAliveEnabled: boolean = false;
+  private enableDefaultSession: boolean = true;
```
aron-cf: When we talked about it originally we discussed this. This flag shouldn't live on the DO. It's just a piece of state that affects the created sandbox interface. Does that make sense? You should be able to have:

```
const sandbox1 = getSandbox(env.Sandbox, "my-sandbox") // default session
const sandbox2 = getSandbox(env.Sandbox, "my-sandbox", { enableDefaultSession: false }) // stateless
const sandbox3 = getSandbox(env.Sandbox, "my-sandbox")
```

Those three sandboxes should be independent and the `enableDefaultSession` is purely configuration for the returned sandbox instance. This does mean holding that value in a closure in getSandbox and calling the internal methods appropriately.

If we store this on the DO then we're going to run into all kinds of conflicts.

packages/sandbox/src/sandbox.ts
```diff
@@ -1024,19 +1085,11 @@ export class Sandbox<Env = unknown> extends Container<Env> implements ISandbox {
     }
   }
 
-  /**
-   * RPC method to configure container startup timeouts. Idempotent once
-   * the values have been persisted: re-applying the same timeout set is a
-   * no-op. The transport retry budget is recomputed only when at least
-   * one timeout actually changes. Storage is written before the in-memory
-   * mirror and derived state are updated.
-   */
```
aron-cf, line 1096: What's with the pruning of all the comments?

packages/sandbox-container/tests/handlers/file-handler.test.ts
```diff
@@ -55,6 +55,8 @@ const mockContext: RequestContext = {
   sessionId: 'session-456'
 };
 
+const SESSIONLESS_SESSION_ID = 'none';
```
aron-cf: I'm finding it hard to read sessionless session id 😆 because it's not an id for a session

```suggestion
const DISABLE_SESSION_TOKEN = '__DISABLE_SESSION__';
```

packages/sandbox-container/tests/services/process-service.test.ts
```diff
@@ -64,6 +58,8 @@ const createMockProcess = (
   ...overrides
 });
 
+const SESSIONLESS_SESSION_ID = 'none';
```
aron-cf: Shall we put this in a file under the shared package?

aron-cf (changes requested):
Thanks for making that change, it feels conceptually simpler already. I've two comments now:

1. We need to move the `enableDefaultSession` option to be scoped per sandbox instance, not persisted on the DO. See comment inline.
2. We might want to enable `sessionId` for non default session id to be passed to the top level methods. I could be convinced otherwise but I think probably people use this shortcut to avoid creating a session object each time. I'm pretty sure the bridge code uses it.

If it helps we can sit for 15 mins and regroup on the high level requirements.

packages/sandbox-container/src/services/execution-service.ts
```diff
@@ -0,0 +1,666 @@
...
+
+  private resolveTarget(sessionId?: string): ExecutionTarget {
+    const resolved = this.canonicalizeExecutionSessionId(sessionId);
+
+    return resolved === DISABLE_SESSION_TOKEN
+      ? { kind: 'sessionless' }
+      : { kind: 'session', sessionId: resolved };
+  }
+
+  private canonicalizeExecutionSessionId(sessionId?: string): string {
```
aron-cf: nit: canonicalize is a clunky word.

```suggestion
  private resolveExecutionSessionId(sessionId?: string): string {
```

aron-cf (approved):
This looks good. Thanks for working thorough it.

packages/sandbox-container/src/services/execution-service.ts
scuffi: updated 🙏

PR #708, opened 2026-05-19
Enforce preview URL runtime activation

ghostwriternr:
Preview URLs currently mix two different concepts: persistent token authorization and live container runtime state. A preview URL token can survive container restart, but the user process serving that port does not. That means an old preview URL can start the container or reach a later container instance that never called `exposePort()` for that port.

This change separates token validity from current-runtime activation.

In this stack, “current runtime” means the currently running container instance/generation. A preview URL is active only if `exposePort()` was called for that port in that current container runtime.

`portTokens` remains persistent token authorization: it records whether a preview URL token is valid for a sandbox port across container restarts. That token staying valid does not mean a live process is currently serving the port, or that the URL may forward traffic in a later container runtime.

Preview URL traffic now enters the Sandbox Durable Object before forwarding. The Durable Object checks persistent token authorization and current-runtime activation, strips spoofed internal preview headers, then forwards only through `Container.fetchIfRunning()`, which does not start a stopped container.

Preview state updates use Durable Object storage transactions so concurrent expose/unexpose operations read and write coherent token and activation snapshots.

The response contract becomes:

- malformed or non-preview hostnames fall through without being handled as preview requests;
- invalid, revoked, or destroyed tokens return `404 INVALID_TOKEN`;
- valid tokens without current-runtime activation return `410 STALE_PREVIEW_URL`;
- valid tokens with activation also return `410 STALE_PREVIEW_URL` if the current container is no longer running and healthy;
- active preview URLs forward only when the container is already running and healthy.

`exposePort()` is now the explicit preview action that activates a port for the current runtime. Restarting a service on the same port is not enough to revive an old preview URL; callers must expose the port again.

The desktop preview integration now also goes through `exposePort()` activation. It can no longer synthesize a URL from a persisted token when current-runtime activation fails.

This PR only adds a runtime identity used to fence preview URL activation. It does not introduce a general lifecycle manager and does not change sessions, processes, interpreters, mounts, or backups.

Preview route parsing details are kept internal rather than exported as public API.

This PR depends on `@cloudflare/containers` support for `Container.fetchIfRunning()`. The dependency currently points at the preview package from cloudflare/containers#212 and must be replaced with a published version before this stack can merge.

The changeset is intentionally `minor` because preview URLs that previously auto-survived restart now return `410 STALE_PREVIEW_URL` until the port is exposed again in the new runtime.

---

This is **part 1 of 3** in the preview URL lifecycle stack:

- <kbd>&nbsp;3&nbsp;</kbd> #710 — Remove container-local preview port registry
- <kbd>&nbsp;2&nbsp;</kbd> #709 — Align preview port APIs with forwarding
- <kbd>&nbsp;1&nbsp;</kbd> #708 — Enforce preview URL runtime activation 👈
- `main`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1752,10 +1752,10 @@
       }
     },
     "node_modules/@cloudflare/containers": {
-      "version": "0.3.2",
-      "resolved": "https://registry.npmjs.org/@cloudflare/containers/-/containers-0.3.2.tgz",
-      "integrity": "sha512-hWobJrpXCgFJ/HYii6E49eegnnlUDWGacjLd0Fb6nQwTD005+T2eHUPI6qOEq9KPJOd7xTWkhkuXEKxDFvfyqQ==",
-      "license": "ISC"
+      "version": "0.3.4",
+      "resolved": "https://pkg.pr.new/@cloudflare/containers@212",
+      "integrity": "sha512-RrF4SdOCr4xL/WEyb6BOkTVoLPgs7Zsy5tCWOTXhUsCnoUj8OoSJ12OY5af6+49xQmxDpkYuVzs4iu+QjRWeXw==",
+      "license": "MIT OR Apache-2.0"
     },
     "node_modules/@cloudflare/kv-asset-handler": {
       "version": "0.4.2",
@@ -14260,7 +14260,7 @@
       "version": "0.10.3",
       "license": "Apache-2.0",
       "dependencies": {
-        "@cloudflare/containers": "^0.3.0",
+        "@cloudflare/containers": "https://pkg.pr.new/@cloudflare/containers@212",
         "aws4fetch": "^1.0.20",
         "capnweb": "^0.8.0",
         "hono": "^4.12.18"
--- a/package-lock.json
+++ b/package-lock.json
@@ -137,7 +137,7 @@
       "version": "1.0.0",
       "license": "MIT",
       "dependencies": {
-        "@cloudflare/containers": "^0.3.0"
+        "@cloudflare/containers": "^0.3.5"
       },
       "devDependencies": {
         "@cloudflare/sandbox": "*",
@@ -1752,9 +1752,9 @@
       }
     },
     "node_modules/@cloudflare/containers": {
-      "version": "0.3.4",
-      "resolved": "https://pkg.pr.new/@cloudflare/containers@212",
-      "integrity": "sha512-RrF4SdOCr4xL/WEyb6BOkTVoLPgs7Zsy5tCWOTXhUsCnoUj8OoSJ12OY5af6+49xQmxDpkYuVzs4iu+QjRWeXw==",
+      "version": "0.3.5",
+      "resolved": "https://registry.npmjs.org/@cloudflare/containers/-/containers-0.3.5.tgz",
+      "integrity": "sha512-P6jYEDkw1Q9qWRr9iFBxe1fozI5HfGMY6XrNg/jROPGZykcYrrzOluUqXv+q4N8gIoRXPCqJJ1FGALbTqnYTkg==",
       "license": "MIT OR Apache-2.0"
     },
     "node_modules/@cloudflare/kv-asset-handler": {
@@ -14260,7 +14260,7 @@
       "version": "0.10.3",
       "license": "Apache-2.0",
       "dependencies": {
-        "@cloudflare/containers": "https://pkg.pr.new/@cloudflare/containers@212",
+        "@cloudflare/containers": "^0.3.5",
         "aws4fetch": "^1.0.20",
         "capnweb": "^0.8.0",
         "hono": "^4.12.18"
--- a/packages/sandbox/package.json
+++ b/packages/sandbox/package.json
@@ -8,7 +8,7 @@
   "description": "A sandboxed environment for running commands",
   "type": "module",
   "dependencies": {
-    "@cloudflare/containers": "^0.3.0",
+    "@cloudflare/containers": "https://pkg.pr.new/@cloudflare/containers@212",
     "aws4fetch": "^1.0.20",
     "capnweb": "^0.8.0",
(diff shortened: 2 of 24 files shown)


aron-cf:
@ghostwriternr can we break the dependency on @cloudflare/containers here and update the code if/when the patch lands upstream.

turbo.json
```diff
@@ -83,7 +83,8 @@
         "!tests/e2e/browser/**",
         "vitest.e2e.config.ts",
         "packages/sandbox/Dockerfile"
-      ]
+      ],
+      "passThroughEnv": ["TEST_TRANSPORT", "TEST_WORKER_URL"]
```
scuffi, line 87: I don't think we want to passthrough `TEST_WORKER_URL`, it may break the perf tests. I'm looking into why currently

package-lock.json
```diff
@@ -14250,7 +14250,7 @@
       "version": "0.10.1",
       "license": "Apache-2.0",
       "dependencies": {
-        "@cloudflare/containers": "^0.3.0",
+        "@cloudflare/containers": "https://pkg.pr.new/@cloudflare/containers@212",
```
scuffi: I assume this is just for tests?

packages/sandbox/src/sandbox.ts
```diff
@@ -2420,6 +2396,145 @@ export class Sandbox<Env = unknown> extends Container<Env> implements ISandbox {
...
+    }
+
+    const body = await response
+      .clone()
+      .text()
+      .catch(() => null);
+    return (
+      body === 'Container is not running' ||
+      body === 'Container suddenly disconnected, try again'
+    );
```
scuffi: This feels pretty unstable, does it not return usable status codes or anything else incase containers change the wording?

scuffi (approved):
Looks good, left couple comments

packages/sandbox/src/sandbox.ts
```diff
@@ -2833,6 +2843,145 @@ export class Sandbox<Env = unknown> extends Container<Env> implements ISandbox {
...
+        port,
+        sandboxId,
+        containerStatus: validation.containerStatus,
+        reason: validation.reason,
+        method: request.method
+      });
+      return this.stalePreviewURLResponse();
+    }
+
+    const response = await this.fetchIfRunning(proxyRequest, port);
```
whoiskatrin: can the runtime change between this validation and `fetchifrunning()`? for example, if the current container stops and another one starts in that gap, it looks like the request might be sent to the new container even though that runtime did not expose the port yet.

packages/sandbox/src/sandbox.ts
```diff
@@ -4371,31 +4503,56 @@ export class Sandbox<Env = unknown> extends Container<Env> implements ISandbox {
...
-        );
-      }
       const sessionId = this.serializeExecutionContext(
         await this.resolveExecution()
       );
-      await this.client.ports.exposePort(port, sessionId, options?.name);
 
-      tokens[port.toString()] = { token, name: options?.name };
-      await this.ctx.storage.put('portTokens', tokens);
+      await this.client.ports.exposePort(port, sessionId, options.name);
```
whoiskatrin, line 4568: looks like we expose the port before finding out that the token is already taken. if `8080` has `shared` and someone tries `8081` with the same token, do we end up with `8081` exposed even though the call returns an error? maybe this needs a cleanup step when the token check fails.

turbo.json
ghostwriternr: this line is only relevant to `test:e2e:vitest`, so shouldn't affect perf tests? but i'm less aware of the perf test setup so if you still think we should remove, happy to.

package-lock.json
ghostwriternr: yea, i've removed it now and brought the logic into our sdk so we can test how effective it is before upstreaming. i've closed cloudflare/containers#212 for now.

packages/sandbox/src/sandbox.ts
ghostwriternr: yeah i hated this too - earlier when using upstream `fetchIfRunning`, this was inevitable. but now that it's inlined, we now validate DO-owned state and forward through `ctx.container.getTcpPort(port)` directly. staleness handling is now just based on the stored identity instead of parsing the response text.

packages/sandbox/package.json
```diff
@@ -8,7 +8,7 @@
   "description": "A sandboxed environment for running commands",
   "type": "module",
   "dependencies": {
-    "@cloudflare/containers": "^0.3.0",
+    "@cloudflare/containers": "https://pkg.pr.new/@cloudflare/containers@212",
```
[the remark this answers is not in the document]
ghostwriternr: closed the upstream PR, and inlined the fetchIfRunning into our sdk itself, so this is resolved. the dependency is pointing back to the latest containers package release now.

packages/sandbox/src/sandbox.ts
ghostwriternr: eep, you're right yes indeed. i've removed `fetchIfRunning` entirely [as suggested here](https://github.com/cloudflare/sandbox-sdk/pull/708#issuecomment-4492525535), but i've incorporated this fix into a new method inlined in our sdk itself. so now we carry the validated runtime identity into the forwarding step and check it again immediately before dispatch. if the container stops/restarts between, the runtime identity no longer matches and the request returns the stale-preview response instead of being sent to the new container.

packages/sandbox/src/sandbox.ts
ghostwriternr: this is a good catch, and the exact scenario you've described used to and would still happen with just the changes in this PR. but i remove the APIs exposing the container-local registry and code paths using it entirely in #710 which should resolve this case from ever being possible, so not fixed here. once all 3 PRs get merged, all such scenarios should disappear.

whoiskatrin (approved):

PR #709, opened 2026-05-19
Align preview port APIs with forwarding

ghostwriternr:
After preview forwarding becomes current-runtime scoped, the public preview-port APIs need to report the same state that forwarding uses. Otherwise users can see a port reported as exposed even though its preview URL is stale and cannot forward without an explicit `exposePort()` call in the current runtime.

In this stack, persistent token authorization means the preview token is still valid, while current-runtime activation means `exposePort()` has been called for that port in the currently running container runtime.

This change aligns the public preview-port surface with that model.

`getExposedPorts(hostname)` now reports only ports whose preview URLs can be used without another `exposePort()` call: a valid token exists, the sandbox has a current runtime, and the port activation was created for that runtime.

`isPortExposed(port)` now answers the same question for one port. It no longer treats persistent token authorization alone as proof that the preview URL can forward.

`unexposePort(port)` is now idempotent Durable Object-owned revocation. It clears token authorization and runtime activation without waking or reaching the container.

`exposePort()` and `unexposePort()` no longer depend on the old container-local exposed-port registry RPCs. The Sandbox Durable Object’s token and activation records are now the source of truth for deciding whether preview traffic may forward.

The desktop preview integration now also obtains URLs through `exposePort()`. It can no longer synthesize a URL from a persisted token when `exposePort()` fails to create current-runtime activation.

This PR intentionally does not delete the old container-local registry. It only stops the public Sandbox preview-port surface from relying on it. The registry deletion happens in the next PR once the new source of truth is in place.

---

This is **part 2 of 3** in the preview URL lifecycle stack:

- <kbd>&nbsp;3&nbsp;</kbd> #710 — Remove container-local preview port registry
- <kbd>&nbsp;2&nbsp;</kbd> #709 — Align preview port APIs with forwarding 👈
- <kbd>&nbsp;1&nbsp;</kbd> #708 — Enforce preview URL runtime activation
- `main`

### DIFF
--- a/.changeset/stale-preview-urls.md
+++ b/.changeset/stale-preview-urls.md
@@ -3,3 +3,5 @@
 ---
 
 Prevent stale preview URLs from waking or reaching sandbox runtimes. Invalid, revoked, or destroyed preview URLs return `404 INVALID_TOKEN`; authorized URLs that are not activated for the current runtime return `410 STALE_PREVIEW_URL` until the port is exposed again. Existing preview URLs that previously survived container restart now return `410 STALE_PREVIEW_URL` after a restart until the port is exposed again in the new runtime.
+
+`getExposedPorts()` and `isPortExposed()` now report only ports that are currently preview-forwardable in the active runtime. `unexposePort()` is now idempotent: revoking a port that is not currently exposed succeeds without contacting the container.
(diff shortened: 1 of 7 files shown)


scuffi (approved):
really cool approach 🫶

PR #712, opened 2026-05-19
Update vite-sandbox to use new tunnels API

aron-cf:
This updates the vite-sandbox to use the new tunnels API. This improves the usability as it now works locally, on workers.dev and on a production deployment.

---

<a href="https://app.devin.ai/review/cloudflare/sandbox-sdk/pull/712" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


examples/vite-sandbox/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM docker.io/cloudflare/sandbox:0.10.1
+FROM docker.io/cloudflare/sandbox-test:0.10.1
```
aron-cf: Will revert

examples/vite-sandbox/vite.config.js
```diff
@@ -10,12 +10,12 @@ function incrementCounter() {
...
+      const count = 0;
 
-      setInterval(() => {
-        count++;
-        writeFileSync(COUNTER_PATH, `export const count = ${count};\n`);
-      }, 1000);
+      // setInterval(() => {
+      //   count++;
+      //   writeFileSync(COUNTER_PATH, `export const count = ${count};\n`);
+      // }, 1000);
```
aron-cf: Will revert

examples/vite-sandbox/wrangler.jsonc
```diff
@@ -3,9 +3,13 @@
   "name": "sandbox-vite",
   "main": "src/worker.js",
   "assets": {
+    "directory": "./",
```
ghostwriternr: nit: i was curious what this one does, and it looks like this isn’t needed here because the [vite plugin auto-generates this config](https://developers.cloudflare.com/workers/wrangler/configuration/#assets) regardless of what's present and points to the built client output automatically (and that ends up being `"directory": "../client"` in this example). can remove, just to avoid confusion, but is also harmless to keep.

ghostwriternr (approved):
this change is such a major glow-up for doing previews, so nice!

examples/vite-sandbox/wrangler.jsonc
aron-cf: Oh, `wrangler dev` will fail without this, sometimes that's useful to run to figure out what's broken without the vite plugin. But it's not relevant to this example.

aron-cf:
```suggestion
```

PR #721, opened 2026-05-26
Reuse cloudflared release in musl image

ghostwriternr:
Reuse the official `cloudflared` Linux release asset in the Alpine/musl image instead of rebuilding `cloudflared` from source.

The upstream Linux release binaries are static Go binaries that run on Alpine. Using the release asset keeps this image on the same pinned, checksum-verified download path as the glibc variants and removes the extra Go builder stage.

Verification:

- Built the final musl image with Docker
- Verified the bundled `cloudflared` binary runs and is statically linked
- Verified the final image does not include the removed Go builder artifacts
- Verified `cloudflared tunnel --url` creates a working quick tunnel to an HTTP origin inside the Alpine image
- Pre-push hook passed typecheck, including E2E and perf typechecks

---

<a href="https://app.devin.ai/review/cloudflare/sandbox-sdk/pull/721" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/packages/sandbox/Dockerfile
+++ b/packages/sandbox/Dockerfile
@@ -2,15 +2,52 @@
 ARG BUN_VERSION=1
 # Node version — override via --build-arg NODE_VERSION=24
 ARG NODE_VERSION=24
-# cloudflared release — shared across glibc runtime-base and the musl builder.
-# Bump this together with the Go base image tag below.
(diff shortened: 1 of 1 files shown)


aron-cf:
This is excellent, thank you.

PR #722, opened 2026-05-26
Add support for named tunnels

aron-cf:
Currently, `sandbox.tunnels.get(port)` today returns a fresh `*.trycloudflare.com` URL on every container restart. Apps that want to use their own domain have to roll their own `cloudflared` orchestration on the side.

```ts
import { getSandbox } from '@cloudflare/sandbox';

export default {
 async fetch(request: Request, env: Env) {
   const sandbox = getSandbox(env.Sandbox, 'demo');
   await sandbox.startProcess('bun /app/server.js', { processId: 'app' });

   // Quick tunnel — zero config.
   const quick = await sandbox.tunnels.get(8080);
   // → https://random-words.trycloudflare.com

   // Named tunnel, creates Cloudflare Tunnel + CNAME DNS entry under the account (requires CLOUDFLARE_API_TOKEN).
   const named = await sandbox.tunnels.get(8080, { name: 'app' });
   // → https://app.example.com

   return new Response(named.url);
 }
};
```

The SDK aims to keep the credentials surface as small as possible. It reads `CLOUDFLARE_API_TOKEN` and attempts to infer the account id and zone id from the token itself; the user only needs to provide `CLOUDFLARE_ACCOUNT_ID` and `CLOUDFLARE_ZONE_ID` explicitly when inference is ambiguous. The provisioning flow is retry-friendly: partial failures leave Cloudflare-side resources in place so the next call rediscovers them via `findTunnelByName` instead of duplicating, and the same path covers respawning after a container restart.

The `websocket-tunnel` example demonstrates both modes — zero-config quick tunnel by default, named tunnel when `TUNNEL_NAME` and `CLOUDFLARE_API_TOKEN` are set. To verify locally, edit `examples/websocket-tunnel/.dev.vars`, run `npm run dev` from that directory, and visit `http://localhost:8787`.

Test coverage sits at three layers: container-runtime unit tests around the supervised `cloudflared` process, SDK unit tests covering the Cloudflare API client, the credentials resolver, name validation, the options-hash and retry-reuse paths, and the `destroyAll` hook, and a vitest end-to-end test that provisions against a real zone (gated on `TEST_TRANSPORT=rpc` and `CLOUDFLARE_API_TOKEN`).


packages/sandbox/src/tunnels/tunnels-handler.ts
```diff
@@ -272,32 +740,130 @@ export function createTunnelsHandler(host: TunnelsHandlerHost): TunnelsHandle {
...
+          const existing = map[port.toString()];
+          // Defensive: only clear if storage still references this exact
+          // tunnel id. Without this check, a sequence like "old
+          // cloudflared dies → new get() spawns fresh → old callback
+          // fires" would clobber the new record.
+          if (existing?.id === id) {
+            delete map[port.toString()];
+            await txn.put(STORAGE_KEY, map);
+            const meta = await readMetaMap(txn);
+            delete meta[port.toString()];
```
whoiskatrin: wont this lose the info we need to clean up a named tunnel? deleting it makes sense for quick tunnels, but for named tunnels the Cloudflare tunnel and DNS record are still around if `cloudflared` dies. once we delete `dnsRecordId` here, a later `sandbox.destroy()` cant remove them anymore. should named tunnels keep the metadata and get marked as needing a respawn, like they do after a container restart?

packages/sandbox/src/tunnels/tunnels-handler.ts
```diff
@@ -200,15 +605,78 @@ class TunnelsRpcTarget extends RpcTarget implements TunnelsHandler {
...
+          );
+          return;
+        }
+
+        const fetcher = this.#host.fetcher;
+        await Promise.allSettled([
+          metaBefore.dnsRecordId
+            ? deleteDNSRecord({
+                token: config.token,
+                zoneId: config.zoneId,
```
whoiskatrin: i think this breaks the config drift case. we store the account and zone the tunnel was created in, but on destroy we're deleting using the current config instead. if the zone changes, this tries to delete the record from the new zone and then forgets about the old one. should cleanup use the stored account and zone here?

aron-cf: Fixed in [19d5fd3](https://github.com/cloudflare/sandbox-sdk/pull/722/commits/19d5fd317b41930e2aa6ae0830c3b942e2f3963b)

packages/sandbox/src/tunnels/tunnels-handler.ts
aron-cf: Fixed in [21cd5e7](https://github.com/cloudflare/sandbox-sdk/pull/722/commits/21cd5e7e0463e08ca3cd12270c1a68a1152cdddc)

whoiskatrin (approved):

PR #727, opened 2026-05-27
Implement egress interception for S3 mounts

scuffi:
# Summary

When mounting S3-compatible storage with s3fs, the container needs direct access to real credentials, they're written to a password file on disk and used to sign every outgoing request. This means a compromised container process can read and exfiltrate the credentials, or use them to access storage outside the intended bucket scope.

This PR adds a credentialProxy option to mountBucket that removes credentials from the container entirely. Instead of giving s3fs real credentials, the Durable Object intercepts all outbound S3 requests at the network layer, re-signs them with the real credentials, and forwards them upstream. The container only ever holds dummy credentials that are useless outside the proxy.

The Durable Object supports AWS SigV4 signing (S3-compatible endpoints and R2) and GCS HMAC signing, with optimised FUSE defaults for R2 to ensure reliable read and write performance in production.

# Usage

To use the credential proxy and keep credentials out of the container the API is pretty simple, you just have to pass the `credentialProxy` parameter into the mount call (`false` by default), example:

```js
await sandbox.mountBucket('my-bucket', '/mnt/bucket', {
  endpoint: 'https://<account-id>.r2.cloudflarestorage.com',
  provider: 'r2',
  credentials: {
    accessKeyId: env.R2_ACCESS_KEY_ID,
    secretAccessKey: env.R2_SECRET_ACCESS_KEY
  },
  credentialProxy: true
});
```

### DIFF
--- a/packages/sandbox/src/sandbox.ts
+++ b/packages/sandbox/src/sandbox.ts
@@ -2091,6 +2091,9 @@ export class Sandbox<Env = unknown> extends Container<Env> implements ISandbox {
                 ? cleanupError.message
                 : String(cleanupError)
           });
+          this.activeMounts.delete(mountPath);
+          evictSigV4ClientCacheEntry(failedMount.mountId);
+          evictDirectoryMarkerCacheForMount(failedMount.mountId);
         }
       } else {
         this.activeMounts.delete(mountPath);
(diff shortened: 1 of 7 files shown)


packages/sandbox/src/storage-mount/s3-credential-proxy-handler.ts
```diff
@@ -0,0 +1,748 @@
...
+  if (mount.readOnly) {
+    const method = request.method.toUpperCase();
+    if (method !== 'GET' && method !== 'HEAD' && method !== 'OPTIONS') {
+      return new Response('Forbidden: bucket mount is read-only', {
+        status: 403
+      });
+    }
+  }
+
+  const realUrl = new URL(realPath + (url.search || ''), mount.endpoint);
```
whoiskatrin, line 723: would it make sense to have the proxy only sign requests within the bucket and prefix configured for this mount, right now it looks like someone with the mount id could send a request for a different bucket or another prefix and have it signed with the stored credentials, adding coverage for those cases would be great too

packages/sandbox/src/storage-mount/s3-credential-proxy-handler.ts
```diff
@@ -0,0 +1,748 @@
...
+  const requestInfo: CredentialProxyRequestInfo = {
+    authStrategy: mount.authStrategy,
+    bucket: mount.bucket,
+    contentLength: request.headers.get('content-length'),
+    method: request.method,
+    mountId,
+    query: [...url.searchParams.keys()].sort()
+  };
+
+  if (isZeroLengthDirectoryMarkerPUT(cleanRequest, realPath)) {
```
whoiskatrin, line 747: one thing i noticed is that empty directory markers seem to be kept only in the local cache rather than persisted upstream, so mkdir can succeed but the empty directory may disappear after a restart, remount, or when accessed from another sandbox, could we forward these marker writes to storage as well

packages/sandbox/src/sandbox.ts
```diff
@@ -1896,8 +2062,16 @@ export class Sandbox<Env = unknown> extends Container<Env> implements ISandbox {
         }
       }
 
-      // Clean up reservation on failure
+      // Clean up reservation on failure; reconfigure proxy without this mount
+      const failedMount = this.activeMounts.get(mountPath);
       this.activeMounts.delete(mountPath);
```
whoiskatrin: i wonder if we should hold onto the mount state until the proxy configuration has been successfully updated, if that update fails during cleanup it looks like the previous fetcher may still accept requests for this mount while we no longer have the state needed to retry removing it

aron-cf (approved):

PR #728, opened 2026-05-27
Allow more time for trycloudflare DNS to propagate in E2E

aron-cf:
cloudflared signals /ready once the tunnel is registered, but the
random *.trycloudflare.com hostname does not become routable until
Cloudflare's edge picks up the new record. CI runners regularly see
that lag exceed the previous 10s retry budget, which surfaced as
"fetchWithRetry failed ... fetch failed" in the quick-tunnel test.

Triple the budget to ~30s and sleep once before the first attempt so
the inevitable cold-cache miss does not eat a retry.

---

<a href="https://app.devin.ai/review/cloudflare/sandbox-sdk/pull/728" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/tests/e2e/tunnel-quick-workflow.test.ts
+++ b/tests/e2e/tunnel-quick-workflow.test.ts
@@ -128,12 +128,13 @@ console.log("Server listening on port " + server.port);
         expect(listed.map((t) => t.id)).toContain(tunnel.id);
 
         // 5. Fetch the marker through the public URL. The Cloudflare edge
-        //    can take a couple of seconds to propagate even after
-        //    cloudflared reports /ready, so retry briefly.
+        //    can take 10–20 seconds to register the *.trycloudflare.com
+        //    route after cloudflared reports /ready, so allow a generous
+        //    retry budget. CI has been observed to need >10s of polling.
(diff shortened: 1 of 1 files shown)


whoiskatrin (approved):

ghostwriternr (approved):

ghostwriternr:
wonder if this is something we can/should detect from the sdk itself so users don't have to write the same code every time, by polling for the public record (which will essentially make the roundtrip from the DO to the container but via the internet).

aron-cf:
> wonder if this is something we can/should detect from the sdk itself so users don't have to write the same code every time, by polling for the public record (which will essentially make the roundtrip from the DO to the container but via the internet).

@ghostwriternr we might have to. I didn't do this in the end for speed. I figured in many cases the time it takes the URL to propagate to a user would be longer than the DNS resolution. Whereas waiting for the DNS slows things down. Maybe it should be a flag on tunnels? `get(8080, { waitForDNS: true })`? or `tunnel.waitForDNS()` Then you could do:

```ts
const tunnel = sandbox.tunnels.get(8080).waitForDNS();
```

PR #730, opened 2026-05-29
Fix Office file binary detection

ghostwriternr:
Office Open XML MIME types contain \`xml\` in their vendor subtype, but the files
are ZIP-based binary documents. The existing MIME classifier treated them as
text, so reading \`.xlsx\` and \`.docx\` files returned UTF-8 content instead of
base64.

Classify textual MIME types by exact known types and structured \`+json\`/\`+xml\`
suffixes so JSON/XML-like files stay readable while Office documents are
returned as binary content.

---

<a href="https://app.devin.ai/review/cloudflare/sandbox-sdk/pull/730" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/packages/sandbox-container/src/services/file-service.ts
+++ b/packages/sandbox-container/src/services/file-service.ts
@@ -24,8 +24,31 @@ export interface SecurityService {
   validatePath(path: string): { isValid: boolean; errors: string[] };
 }
 
-// Maximum file size for RPC transfers is 32 MiB to prevent performance issues. For larger files, clients should use streaming APIs.
+// Maximum file size for RPC transfers is 32 MiB. Larger files should use streaming APIs.
 const MAX_RPC_FILE_SIZE = 32 * 1_048_576; // 32 MiB
+
+const TEXT_MIME_TYPES = new Set([
+  'application/ecmascript',
+  'application/javascript',
+  'application/json',
+  'application/json-seq',
+  'application/rtf',
+  'application/sql',
+  'application/toml',
+  'application/typescript',
+  'application/x-empty',
+  'application/x-httpd-php',
+  'application/x-javascript',
+  'application/x-sh',
+  'application/x-shellscript',
+  'application/x-typescript',
+  'application/x-yaml',
+  'application/xml',
+  'application/xml-dtd',
+  'application/xml-external-parsed-entity',
+  'application/yaml',
+  'inode/x-empty'
+]);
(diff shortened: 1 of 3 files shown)


packages/sandbox-container/src/services/file-service.ts
```diff
@@ -24,9 +24,32 @@ export interface SecurityService {
   validatePath(path: string): { isValid: boolean; errors: string[] };
 }
 
-// Maximum file size for RPC transfers is 32 MiB to prevent performance issues. For larger files, clients should use streaming APIs.
+// Maximum file size for RPC transfers is 32 MiB. Larger files should use streaming APIs.
 const MAX_RPC_FILE_SIZE = 32 * 1_048_576; // 32 MiB
 
+const TEXT_MIME_TYPES = new Set([
```
ghostwriternr, line 30: q for reviewers: will we regret replacing this hard-coded list here with a hard-pinned version of [mime-db](https://github.com/jshttp/mime-db)?

ghostwriternr: actually, that db doesn't seem to help us classify either so eh, nvm

packages/sandbox-container/src/services/file-service.ts
```diff
@@ -24,9 +24,32 @@ export interface SecurityService {
   validatePath(path: string): { isValid: boolean; errors: string[] };
 }
 
-// Maximum file size for RPC transfers is 32 MiB to prevent performance issues. For larger files, clients should use streaming APIs.
+// Maximum file size for RPC transfers is 32 MiB. Larger files should use streaming APIs.
```
aron-cf: This feels incorrect now. It's maximum size for HTTP transport. Larger files should use RPC transport with `readFile("<path>", { encoding: "none" })` to get a ReadableStream

aron-cf (approved):
This looks fine. FWIW a better way to handle files now is to use `readFile("my-path.docx", { encoding: "none" })`. This will return a readable stream of binary data with no encoding overhead.

https://developers.cloudflare.com/sandbox/api/files/#readfile

packages/sandbox-container/src/services/file-service.ts
ghostwriternr: good note, i agree. i've clarified the comment and error message to be clearer about this.

PR #732, opened 2026-05-29
Add exposed-port endpoint to Sandbox Bridge

ghostwriternr:
Add `POST /v1/sandbox/:id/exposed-port/:port` to the Sandbox Bridge to provide public HTTP access to services running inside a sandbox.

Consumers using the Bridge API need a REST interface for routing external traffic to internal container services such as web servers, API backends, or language servers. This endpoint creates or reuses a public endpoint for a service that is already listening inside the sandbox.

Empty requests return an ephemeral `*.trycloudflare.com` endpoint. Requests with a `name` field create or reuse a named endpoint under the configured Cloudflare zone. The endpoint is backed by Sandbox tunnels.

Tests Run:

- `npm run check`
- `npm test -w @cloudflare/sandbox-bridge`

### DIFF
--- a/packages/sandbox/src/bridge/openapi.ts
+++ b/packages/sandbox/src/bridge/openapi.ts
@@ -167,25 +167,25 @@ export const OPENAPI_SCHEMA = {
           }
         }
       },
-      ExposedPortRequest: {
+      TunnelRequest: {
         type: 'object',
         properties: {
           name: {
             type: 'string',
             description:
-              'Subdomain prefix for a named endpoint, such as `app`. Do not pass a full hostname. Omit to create or reuse an ephemeral endpoint.',
+              'Subdomain prefix for a named tunnel, such as `app`. Do not pass a full hostname. Omit to create or reuse an ephemeral tunnel.',
             example: 'app'
           }
         }
       },
-      ResolvedExposedPort: {
+      Tunnel: {
         type: 'object',
         required: ['id', 'port', 'url', 'hostname', 'createdAt'],
         properties: {
           id: { type: 'string' },
           port: {
             type: 'integer',
-            description: 'Container port served by the endpoint.',
+            description: 'Container port served by the tunnel.',
             example: 8080
           },
           url: {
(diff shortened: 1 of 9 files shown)


packages/sandbox/src/bridge/openapi.ts
```diff
@@ -379,6 +421,82 @@ export const OPENAPI_SCHEMA = {
         }
       }
     },
+    '/v1/sandbox/{id}/exposed-port/{port}': {
```
aron-cf: Can we call this `tunnel`?

```suggestion
    '/v1/sandbox/{id}/tunnel/{port}': {
```

packages/sandbox/src/bridge/openapi.ts
```diff
@@ -379,6 +421,82 @@ export const OPENAPI_SCHEMA = {
...
+      post: {
```
aron-cf, line 425: Do we need `DELETE` too? Feels like it should support this.

packages/sandbox/src/bridge/routes.ts
```diff
@@ -610,6 +621,66 @@ export function createBridgeApp(
...
+          throw err;
+        }
+        options = { name: body.name };
+      }
+    }
+
+    const sandbox = getSandbox(getSandboxNs(c.env), c.get('containerUUID'));
+
+    try {
+      const tunnel = await sandbox.tunnels.get(port, options);
```
aron-cf, line 683: Are the named tunnel errors (e.g. name without credentials) clear?

aron-cf (approved):
Looks great! But I'd like it to be called tunnel for parity with the sandbox feature.

packages/sandbox/src/bridge/openapi.ts
ghostwriternr: done, understood this helps us with implementation consistency and reduces confusion long-term.

packages/sandbox/src/bridge/openapi.ts
ghostwriternr: good point, added

packages/sandbox/src/bridge/routes.ts
ghostwriternr: the bridge preserves the error message from the underlying `sandbox.tunnels.get()`, so missing credentials return something like `tunnel failed: Named tunnels require CLOUDFLARE_API_TOKEN`... i've updated the README to include the named-tunnel env vars too.

PR #733, opened 2026-05-29
Allow backup bucket URL override

scuffi:
# Summary

Fixes backup/restore presigned URLs for R2 buckets in jurisdictions. Previously, backup URLs were always generated against the default R2 endpoint: `https://<account_id>.r2.cloudflarestorage.com`
For jurisdiction-specific buckets, this can fail because they require endpoints like: `https://<account_id>.eu.r2.cloudflarestorage.com`

This PR adds `BACKUP_BUCKET_ENDPOINT` so users can override the R2 endpoint used for backup and restore presigned URLs. It also validates that the endpoint is HTTPS and origin-only, with tests for invalid values.

# Example

`wrangler.jsonc`
```json
	"vars": {
		"BACKUP_BUCKET_NAME": "sandbox-backup-endpoint-repro",
		"BACKUP_BUCKET_ENDPOINT": "https://<account-id>.eu.r2.cloudflarestorage.com"
	},
	"r2_buckets": [
		{
			"binding": "BACKUP_BUCKET",
			"bucket_name": "my-backup-bucket",
			"jurisdiction": "eu"
		}
	],
```

### DIFF
--- a/packages/sandbox/src/sandbox.ts
+++ b/packages/sandbox/src/sandbox.ts
@@ -1131,7 +1131,7 @@ export class Sandbox<Env = unknown> extends Container<Env> implements ISandbox {
     // Read R2 presigned URL credentials for direct container-to-R2 backup transfers
     // R2 account id precedence: CLOUDFLARE_R2_ACCOUNT_ID > CLOUDFLARE_ACCOUNT_ID.
     // Token-derived fallback is intentionally not wired here because the
-    // backup path (requirePresignedUrlSupport) is synchronous; see
+    // backup path (requirePresignedURLSupport) is synchronous; see
     // tunnels/credentials.ts for the full chain that named tunnels use.
     this.r2AccountId =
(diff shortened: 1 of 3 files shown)


packages/sandbox/src/sandbox.ts
```diff
@@ -5566,21 +5616,37 @@ export class Sandbox<Env = unknown> extends Container<Env> implements ISandbox {
...
+    return new URL(
+      `${this.getBackupBucketEndpoint(accountId)}/${encodedBucket}/${encodedKey}`
     );
+  }
+
+  /**
+   * Generate a presigned GET URL for downloading an object from R2.
+   * The container can curl this URL directly without credentials.
+   */
+  private async generatePresignedGetUrl(r2Key: string): Promise<string> {
```
aron-cf: Nit but there's a mix of Url and URL here. I'd prefer URL

aron-cf (approved):

whoiskatrin (approved):

PR #736, opened 2026-06-03
Fix S3/R2 mount credential proxy when catch-all handler defined

### LINKED ISSUE #720: mountBucket() silently fails when a class-level outbound handler is set via setOutboundHandler (catch-all shadows per-host r2.internal interceptor)
## Summary

When a Sandbox class has a class-level outbound handler registered via
`setOutboundHandler(...)` (i.e. a catch-all for credential injection,
URL allowlisting, etc.), `mountBucket()`'s internally-installed per-host
interceptor for `r2.internal` is shadowed at the workerd routing layer.
All s3fs traffic to `r2.internal` reaches the catch-all handler instead
of the SDK's R2 egress proxy, which forwards it to the public internet,
fails (`r2.internal` is not a real DNS name), and the s3fs daemon dies
without surfacing an error. The mount appears to succeed; subsequent
reads/writes hit the underlying empty container directory.

This combination — a Sandbox subclass that both (a) sets a catch-all
outbound handler and (b) uses `mountBucket()` for workspace persistence —
is, I think, a common pattern (any deployment that wants to inject auth
or apply egress policy AND wants R2-backed persistence). I expected it
to work transparently.

## Environment

- `@cloudflare/sandbox`: `0.10.2`
- `@cloudflare/containers`: `0.3.3`
- `compatibility_date`: `2026-04-21`
- `compatibility_flags`: `["nodejs_compat"]`
- Transport: `rpc` (set on every `getSandbox()` call)
- Sandbox base image: `cloudflare/sandbox:0.10.2` (matched to npm)

## Repro

Minimal Sandbox subclass:

```ts
import { Sandbox as CFSandbox, getSandbox } from "@cloudflare/sandbox";

export class Sandbox extends CFSandbox {
  enableInternet = true;
  interceptHttps = true;
}

Sandbox.outboundHandlers = {
  async catchAll(req, env, ctx) {
    // any catch-all — even a passthrough that just calls fetch(req)
    return fetch(req);
  },
};

export default {
  async fetch(request, env) {
    const sandbox = getSandbox(env.Sandbox, "tenant-1", { transport: "rpc" });
    await sandbox.setOutboundHandler("catchAll", {});
    await sandbox.mountBucket("MY_BUCKET", "/workspace", {
      prefix: "/workspaces/tenant-1/",
    });
    // The mount call returns without throwing.
    // s3fs has actually died; /workspace is the empty container dir.
    return sandbox.exec("ls /workspace");
  },
};
```

(`MY_BUCKET` is an R2 binding in `wrangler.toml`.)

### Expected

`/workspace` is backed by the R2 bucket prefix. Writes propagate.

### Actual

`mountBucket` resolves successfully. `/workspace` is unmounted (an empty
directory in the container's overlay). Any prior writes from earlier
container instances are invisible. Adding a `console.log` in `catchAll`
reveals it receives the s3fs requests for `r2.internal/MY_BUCKET/...`,
forwards them via `fetch`, and gets DNS / connect failures.

## Root cause (best guess from reading `@cloudflare/containers` source)

`setOutboundHandler` promotes the Sandbox instance into "intercept-all"
mode (workerd-level `interceptAllOutboundHttp`). Once that's active,
the SDK's separate `interceptOutboundHttp("r2.internal", sdkFetcher)`
installation inside `mountBucket()` is unreachable — the workerd
routing table sends every host (including `r2.internal`) to the
catch-all fetcher.

The documented precedence (per-host > catch-all) only applies inside
a *single* `ContainerProxy` fetcher's `outboundByHostOverrides` table.
Two separate fetchers don't compose that way.

## Workarounds (we shipped option B)

**(A) Reimplement the R2 egress translator.** ~200 lines reproducing
the S3 REST API → R2 binding mapping the SDK already does internally.
Self-contained but duplicates non-trivial SDK code.

**(B) Borrow the SDK's handler.** Read `r2EgressHandler` out of the
shared `outboundHandlersRegistry` in `@cloudflare/containers` via
`Reflect.get(Container, "outboundHandlers", { name: "CloudflareSandboxR2EgressProxyTarget" })`,
register it on the user's Sandbox class alongside their catch-all, and
configure it per-host via `setOutboundByHost("r2.internal", "r2EgressMount", { buckets: ... })`.
Works today, but depends on a private class name. We added a startup
check that throws loudly with a clear error if the class is ever renamed.

## Proposed fixes

Either would resolve this; (1) is sufficient on its own and the smaller
change:

1. **Export `r2EgressHandler` as a public named export from `@cloudflare/sandbox`,**
   along with the params shape (`{ buckets: Record<string, { prefix?: string; readOnly?: boolean }> }`)
   and the recommended host (`r2.internal`). Users who set a catch-all
   could then explicitly compose the handlers:

   ```ts
   import { r2EgressHandler } from "@cloudflare/sandbox";

   Sandbox.outboundHandlers = {
     r2EgressMount: r2EgressHandler,
     catchAll: myCatchAll,
   };
   // before mountBucket:
   await sandbox.setOutboundByHost("r2.internal", "r2EgressMount", {
     buckets: { MY_BUCKET: { prefix } },
   });
   await sandbox.mountBucket("MY_BUCKET", "/workspace", { prefix });
   ```

2. **Or: make `mountBucket()` detect when the user's Sandbox class
   already has a `ContainerProxy` registered (i.e. they've set
   `outboundHandlers` / called `setOutboundHandler`) and register
   `r2EgressMount` on *that* class's handler set + call
   `setOutboundByHost` automatically,** instead of installing a
   separate `interceptOutboundHttp`. Zero-config from the user's POV.

3. **Documentation:** until one of the above ships, the README /
   `mountBucket()` JSDoc should call out that combining `mountBucket`
   with `setOutboundHandler` requires manual composition, with a
   pointer to the recommended pattern.

## Why this matters

The bug is silent — `mountBucket` resolves, `ls /workspace` returns
(empty), no exception surfaces to user code. The agent then writes
files that aren't persisted, and the data loss is only discovered
later. We caught it in staging by accident.

Happy to PR option (1) if there's interest — it's a small change
(promote the existing internal handler to a public export, add the
class name + handler name to docs). Let me know.

scuffi:
## Summary

Fixes bucket mounts when a Sandbox class also defines a catch-all outbound handler.

Previously, `mountBucket()` configured `r2.internal` and credential-proxy hosts with `setOutboundByHost()`, but dispatch still depended on the Containers `ContainerProxy` resolving SDK-internal handler names through its outbound handler registry. In catch-all interception scenarios, mount traffic could be routed incorrectly, causing `s3fs` to exit without establishing the mount.

This PR introduces an SDK-owned `ContainerProxy` that directly dispatches SDK-reserved mount hosts to the SDK mount handlers, while preserving normal Containers behavior for all other outbound traffic.

## What changed

- **`packages/sandbox/src/sandbox.ts`**
  - Adds SDK `ContainerProxy extends BaseContainerProxy`
  - Directly handles SDK mount hosts:
    - `r2.internal` → `r2EgressHandler`
    - credential-proxy hosts → `s3CredentialProxyHandler`
  - Falls back to `super.fetch(request)` for all non-SDK outbound traffic
  - Clarifies comments around `setOutboundByHost()` validation vs actual proxy dispatch

- **`packages/sandbox/src/index.ts`**
  - Exports the SDK `ContainerProxy` instead of re-exporting it from `@cloudflare/containers`

## Stack

This PR is stacked on top of #727 and fixes both R2 & S3 handlers (implemented in that PR), but kept separate to avoid bloating #727 

Closes https://github.com/cloudflare/sandbox-sdk/issues/720

### DIFF
--- a/.changeset/fix-catch-all-mount-proxy.md
+++ b/.changeset/fix-catch-all-mount-proxy.md
@@ -1,0 +1,5 @@
+---
+'@cloudflare/sandbox': patch
+---
+
+Fix bucket mounts when a Sandbox class defines a catch-all outbound handler by routing SDK-managed mount hosts through the SDK ContainerProxy.
--- a/packages/sandbox/src/index.ts
+++ b/packages/sandbox/src/index.ts
@@ -16,8 +16,6 @@ export { getSandbox, Sandbox } from './sandbox';
 
 // Legacy types are now imported from the new client architecture
 
-// Required export for egress intercepting
-export { ContainerProxy } from '@cloudflare/containers';
 // Export core SDK types for consumers
 export type {
   BackupOptions,
@@ -155,6 +153,8 @@ export { CodeInterpreter } from './interpreter.js';
 export { proxyTerminal } from './pty';
 // Re-export request handler utilities
 export { proxyToSandbox, type SandboxEnv } from './request-handler';
+// Required export for egress intercepting
+export { ContainerProxy } from './sandbox';
 // Export SSE parser for converting ReadableStream to AsyncIterable
 export {
   asyncIterableToSSEStream,
--- a/packages/sandbox/src/sandbox.ts
+++ b/packages/sandbox/src/sandbox.ts
@@ -1,4 +1,10 @@
-import { Container, getContainer, switchPort } from '@cloudflare/containers';
+import {
+  ContainerProxy as BaseContainerProxy,
+  Container,
+  getContainer,
+  type OutboundHandlerContext,
+  switchPort
+} from '@cloudflare/containers';
 import type {
   BackupOptions,
   BucketCredentials,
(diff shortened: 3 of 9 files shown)


ghostwriternr (approved):
LGTM! only minor comment is that i see you've updated test mocks right now but can you also just update tests to ensure `r2.internal` and `s3-credential-proxy.internal` go through the sdk proxy path by default as intended and unrelated outbound hosts work properly? just to ensure this doesn't regress in future

PR #737, opened 2026-06-03
Patch `startProcess()` leaves phantom running process

### LINKED ISSUE #725: Failed startProcess() leaves phantom running process
> *This was generated by AI during triage.*

**Describe the bug**

When `startProcess()` fails during streamed startup, `ProcessService` can leave behind a stored process record with status `running`, even though the caller received a startup error and no usable process was started.

This makes subsequent `listProcesses()` calls return a phantom running process.

**Affected code**

In `packages/sandbox-container/src/services/process-service.ts`, `executeCommandStream()` stores the process record before it asks `SessionManager` to start the streamed command:

```ts
const processRecord: ProcessRecord = {
  ...processRecordData,
  commandHandle: {
    sessionId,
    commandId: processRecordData.id
  }
};

await this.store.create(processRecord);

const streamResult = await this.sessionManager.executeStreamInSession(...);

if (!streamResult.success) {
  return streamResult as ServiceResult<ProcessRecord>;
}
```

If `executeStreamInSession()` returns a failure, or throws after `store.create()`, there is no cleanup or terminal-status update for `processRecord`.

Since the record was created with `status: 'running'`, `ProcessStore.list()` continues exposing it as active.

**To Reproduce**

A focused unit-test repro at the `ProcessService` seam is sufficient:

1. Construct `ProcessService` with an in-memory `ProcessStore` implementation.
2. Mock `sessionManager.executeStreamInSession()` to return a startup error:

   ```ts
   {
     success: false,
     error: { message: 'session unavailable', code: 'SESSION_ERROR' }
   }
   ```

3. Call:

   ```ts
   const result = await service.startProcess('node server.js', {
     processId: 'requested-server'
   });
   ```

4. Confirm `result.success === false`.
5. Call `await service.listProcesses()`.

**Actual behavior**

The failed startup remains listed as running:

```ts
[
  {
    id: 'requested-server',
    command: 'node server.js',
    status: 'running',
    pid: undefined,
    // ...
  }
]
```

I verified this with a temporary unit-test repro; it failed with the phantom record shown above.

**Expected behavior**

A process that fails before startup completes should not remain visible as `running`. Either:

- remove the record when startup fails, or
- record an appropriate terminal/error state that accurately reflects the failed startup.

**Impact**

This produces incorrect process lifecycle state and can interfere with callers that deduplicate background services based on `listProcesses()`.

One concrete impact is the OpenCode integration in `packages/sandbox/src/opencode/opencode.ts`, which searches for a matching process in `starting` or `running` status and reuses it instead of starting another server. A phantom running record from a failed startup can cause later attempts to reuse a nonexistent OpenCode server rather than launching a new one.

Other symptoms can include:

- stale entries in `listProcesses()`;
- `killAllProcesses()` trying to kill a nonexistent process;
- background-service startup logic becoming stuck on a failed process record;
- retention of failed startup records in the in-memory process store.

**Suggested fix**

Clean up the record on failure after `store.create()`:

```ts
const streamResult = await this.sessionManager.executeStreamInSession(...);

if (!streamResult.success) {
  await this.store.delete(processRecord.id);
  return streamResult as ServiceResult<ProcessRecord>;
}
```

The exception path after `store.create()` should receive equivalent cleanup. Alternatively, processes could begin in `starting` state and transition to `running` only after a valid `start` event, with startup failure persisted as `error`.

**Suggested regression tests**

Add tests in `packages/sandbox-container/tests/services/process-service.test.ts` covering:

1. `executeStreamInSession()` returning failure leaves no `running` process behind.
2. An exception after `store.create()` leaves no `running` process behind.
3. Successful streamed startup still returns and lists a running process.
4. Optionally, a higher-level OpenCode retry test confirming failed startup is not reused as an existing server.

**Version**

Observed on current `main` (`a2181337`, after `0.10.2`).

scuffi:
## Summary

When `startProcess()` fails during startup (e.g. the session is unavailable), callers receive an error but the process record remains stored with `status: 'running'`. Subsequent calls to `listProcesses()` return a phantom running process that will never complete or be cleaned up.

Both startup failure paths in `executeCommandStream()` returned a `{ success: false }` result and a thrown exception now mark the stored record as terminal `error` via a shared `markStartupFailed()` helper, which also preserves the failure message in `stderr`. This evicts the record from the in-memory store, so it no longer appears as `running`.

Closes #725

### DIFF
--- a/.changeset/fix-process-startup-phantom.md
+++ b/.changeset/fix-process-startup-phantom.md
@@ -1,0 +1,5 @@
+---
+'@cloudflare/sandbox': patch
+---
+
+Fix phantom `running` processes after a failed `startProcess` call. When the underlying session was unavailable or threw during startup, the process record was left in memory with status `running` and would appear in `listProcesses()` indefinitely. Failed startups are now correctly marked as terminal `error` records.
--- a/packages/sandbox-container/src/services/process-service.ts
+++ b/packages/sandbox-container/src/services/process-service.ts
@@ -105,6 +105,9 @@ export class ProcessService {
     options: ProcessOptions = {}
   ): Promise<ServiceResult<ProcessRecord>> {
     const startTime = Date.now();
+    let processRecord: ProcessRecord | undefined;
+    let storedProcessRecord = false;
+
     try {
       // 1. Validate command (business logic via manager)
       const validation = this.manager.validateCommand(command);
@@ -129,14 +132,16 @@ export class ProcessService {
         commandId: processRecordData.id
       };
 
-      const processRecord: ProcessRecord = {
+      processRecord = {
         ...processRecordData,
         commandHandle
       };
+      const storedRecord = processRecord;
       let executionSessionId = commandHandle.sessionId;
 
       // 4. Store record (data layer)
-      await this.store.create(processRecord);
+      await this.store.create(storedRecord);
+      storedProcessRecord = true;
 
       // 5. Execute command with streaming and bind it to the process record
       // Pass process ID as commandId for tracking and killing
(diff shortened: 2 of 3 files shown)


aron-cf (approved):
👏 nice fix

PR #739, opened 2026-06-04
Remove desktop functionality

scuffi:
## Summary
Removes the desktop client feature from the Sandbox SDK, including `sandbox.desktop`, `DesktopClient`, desktop-related types/errors, container services/routes, tests, examples, and the `-desktop` Docker image variant.

## Changes
- Removed desktop SDK exports and public API surface
- Removed desktop container services, handlers, worker, native wrapper, and routes
- Removed desktop E2E/test-worker routes and Docker image configuration
- Removed desktop Docker build/publish references and docs


whoiskatrin:
overall this looks pretty solid to me. only thing i’d double check is the DO migration in the e2e worker template

since `SandboxDesktop` already existed in the deployed migrations as `v6`, i think we probably need a new delete migration here instead of just removing that entry?

couple tiny things too: `EXPECTED_PR_APPS` can probably be 5 now instead of 6 and the agent docs still mention the desktop image/example

whoiskatrin (approved):

PR #740, opened 2026-06-04
Consolidate SDK streaming APIs

scuffi:
# Summary

Replaces the `exec()` return type with `ExecProcess` — a process handle that provides both streaming and buffered access to command output, aligned with the Containers SDK `ExecProcess` pattern.

`exec()` now returns synchronously. Because `ExecProcess` implements `PromiseLike<ExecResult>`, existing code that awaits the result continues to work with **zero migration**:

```ts
// Before AND after — identical, no changes needed
const result = await sandbox.exec("ls -la");
console.log(result.stdout);
```

New streaming and exit-code-only patterns are now available without a separate method:

```ts
// Stream stdout in real time
const proc = sandbox.exec("tail -f /var/log/app.log");
for await (const chunk of proc.stdout) {
  // process chunk
}

// Explicit buffered output
const result = await sandbox.exec("npm test").output();

// Just the exit code
const exitCode = await sandbox.exec("test -f /config").exitCode;
```

`execStream()` is now deprecated in favor of `exec().stdout`.

# Architecture

`ExecProcess` uses a **dual-path** design to preserve exact behavioral parity:

- **`output()` / `then()` (await)** — uses the existing buffered HTTP exec endpoint (`/api/execute`). This is the same code path as the old `exec()`, so timeout errors, exit codes, and stdout/stderr formatting are identical.
- **`.stdout` / `.stderr` / `.exitCode`** — lazily starts the SSE streaming endpoint (`/api/execute/stream`) only when first accessed. No extra container round-trip happens for the common `await sandbox.exec(cmd)` case.

### DIFF
--- a/packages/sandbox/src/sandbox.ts
+++ b/packages/sandbox/src/sandbox.ts
@@ -678,16 +678,26 @@ export function getSandbox<T extends Sandbox<any>>(
   // to ensure the returned session uses proxyTerminal instead of RPC's terminal.
   const enhancedMethods = {
     fetch: (request: Request) => stub.fetch(request),
-    exec: (command: string, execOptions?: ExecOptions) => {
-      const streamPromise = useDefaultSession
-        ? stub.execStream(command, execOptions)
-        : stub.execStreamWithSessionToken(
-            command,
-            DISABLE_SESSION_TOKEN,
-            execOptions
-          );
-      return createExecProcess(command, streamPromise, execOptions?.signal);
-    },
+    exec: (command: string, execOptions?: ExecOptions) =>
+      createExecProcess({
+        buffered: () =>
+          useDefaultSession
+            ? stub.exec(command, execOptions).then((r) => r)
+            : stub.execWithSessionToken(
+                command,
+                DISABLE_SESSION_TOKEN,
+                execOptions
+              ),
+        stream: () =>
+          useDefaultSession
+            ? stub.execStream(command, execOptions)
+            : stub.execStreamWithSessionToken(
+                command,
+                DISABLE_SESSION_TOKEN,
+                execOptions
+              ),
+        signal: execOptions?.signal
+      }),
     startProcess: (command: string, processOptions?: ProcessOptions) =>
       useDefaultSession || processOptions?.sessionId !== undefined
         ? stub.startProcess(command, processOptions)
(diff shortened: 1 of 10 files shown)


packages/sandbox/src/openai/index.ts
aron-cf, line 1: Can probably revert these changes? They look left over?

packages/shared/src/types.ts
```diff
@@ -75,6 +75,14 @@ export interface ExecOptions extends BaseExecOptions {
   origin?: 'user' | 'internal';
 }
 
+export type ExecStreamOptions = Omit<
+  ExecOptions,
+  'onOutput' | 'onComplete' | 'onError'
+> & {
+  stream: true;
```
aron-cf: Does TypeScript correctly identify the return type based on stream: true? I thought the ExecOptions would need a `stream?: never` too?

packages/sandbox/src/sandbox.ts
```diff
@@ -596,14 +606,23 @@ export function getSandbox<T extends Sandbox<any>>(
...
+    exec: (command: string, execOptions?: ExecOptions | ExecStreamOptions) => {
+      if (useDefaultSession) {
+        return stub.exec(command, execOptions);
+      }
+      if (execOptions?.stream && !hasExecCallbacks(execOptions)) {
+        return stub.execStreamWithSessionToken(
+          command,
+          DISABLE_SESSION_TOKEN,
+          execOptions
+        );
```
aron-cf: I want to make sure this is the right API before we go and add a new interface. Currently this is an SSE event stream right? So the caller needs to use parseSSEStream() to decode it? I think that's okay because JSRPC doesn't support streaming anything other than Uint8Array. But maybe we want to decode the stdout/stderr chunks...

packages/sandbox/src/interpreter.ts
```diff
@@ -44,8 +44,17 @@ export class CodeInterpreter {
...
-  ): Promise<Execution> {
+    options: RunCodeOptions & { stream: true }
+  ): Promise<ReadableStream>;
+  async runCode(code: string, options?: RunCodeOptions): Promise<Execution>;
+  async runCode(
+    code: string,
+    options: RunCodeOptions | (RunCodeOptions & { stream: true }) = {}
+  ): Promise<Execution | ReadableStream> {
+    if ('stream' in options && options.stream) {
+      return this.runCodeStream(code, options);
```
aron-cf: What's the output here? Stdout from the model? Or same as exec?

aron-cf (commented):
Some interim comments, I need to look into the code interpreter some more.

packages/sandbox/src/interpreter.ts
scuffi: This is the same output as the existing `runCodeStream()`, the stream is SSE-framed `ExecutionEvent` data from code execution: stdout, stderr, result, error/completion-style interpreter events. It's not model stdout, or the same schema as exec

PR #744, opened 2026-06-05
Fix bridge dropping credentialProxy property

scuffi:
## Summary

Fixes the bridge mount API dropping the `credentialProxy` option before forwarding mount requests to the SDK.

Previously, bridge/API callers could send `credentialProxy: true`, but the bridge option conversion omitted that property. As a result, the SDK received a normal remote mount request and performed a direct s3fs mount instead of configuring the credential proxy.

For bridge-based remote mounts with credentials available, `credentialProxy: true` was silently ignored and real credentials could be written into the container-side s3fs password file. Direct SDK usage was not affected.

_Note: There is no changeset included in this PR as the S3 proxy mount behaviour is not yet part of any official release version_


ghostwriternr (approved):
neat! only recommendation again is to add a regression test on the bridge side too pls.

PR #748, opened 2026-06-09
Retry transient WebSocket upgrade failures

ghostwriternr:
The Sandbox DO opens a WebSocket control-plane connection to the container. Until now, only HTTP 503 on the upgrade was treated as transient — every other 5xx terminated the connect attempt, even when the container was healthy and would have accepted the next attempt.

This widens the retry policy on the WebSocket upgrade leg only:

- 500/502/503/504 are retried against the existing 2-minute budget with exponential backoff.
- 400/401/403/404 stay terminal so client/validation errors fail fast.
- HTTP startup retry is unchanged — `BaseTransport` still only retries 503.
- The container runtime now returns 503 (instead of 500) when `server.upgrade()` itself fails, so producer and consumer agree a failed upgrade is retryable.

The asymmetry is deliberate: a 5xx on the WebSocket upgrade means we never established the channel and no application code ran, so retrying is safe. A 5xx on a normal HTTP request means the container actually attempted the work, so blindly retrying could double-execute. 503 is the one HTTP status that carries the specific "container still starting, no handler mounted" meaning.

The retry loop that previously lived inline in both upgrade paths is now a single internal helper, reused by the HTTP 503 startup path as well.

Out of scope: thrown fetch errors on the upgrade leg (abort/timeout before a Response) are not retried here, and no automatic `ctx.abort()` is added. Those belong to a follow-up control-plane liveness pass if persistent wedges remain after this change.

---

<a href="https://app.devin.ai/review/cloudflare/sandbox-sdk/pull/748" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/.changeset/retry-websocket-upgrades.md
+++ b/.changeset/retry-websocket-upgrades.md
@@ -1,0 +1,5 @@
+---
+'@cloudflare/sandbox': patch
+---
+
+Recover automatically from transient infrastructure failures when the SDK opens its WebSocket control connection to a sandbox. Previously, any 5xx response other than 503 on the upgrade would fail the SDK call even when the container was healthy.
--- a/packages/sandbox-container/src/server.ts
+++ b/packages/sandbox-container/src/server.ts
@@ -29,6 +29,10 @@ export type WSData =
 const logger = createLogger({ component: 'container' });
 const SERVER_PORT = 3000;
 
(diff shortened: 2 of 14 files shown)


whoiskatrin (approved):
LGTM 👍🏼

PR #751, opened 2026-06-11
Implement proxy mount read caching

scuffi:
## Summary

Reduce redundant upstream requests during credential-proxy mounted reads by adding a short-lived HEAD metadata response cache.

s3fs issues frequent HEAD requests for metadata (`getattr`) on every file open, stat, and read. With credential-proxy mounts, each HEAD request was forwarded upstream through the signing proxy, adding latency to repeated reads. This PR keeps the existing `aws4fetch`/`AwsClient` signing path and focuses the optimization on safe metadata caching.

### HEAD metadata cache (`s3-credential-proxy-handler.ts`)

- **Positive cache**: Cache successful HEAD responses for 60s.
- **Negative cache**: Cache 404 HEAD responses for 5s to avoid repeated existence checks for non-existent paths (s3fs probes `path/`, `path_$folder$` variants).
- **PUT priming**: After a successful PUT, synthesize and cache a HEAD-equivalent entry from request/response metadata such as `content-length`, `content-type`, `etag`, `last-modified`, and `x-amz-meta-*`.
- **Conservative bypasses**: Do not cache ranged, conditional, checksum-mode, SSE-C, or GCS customer-encryption HEAD requests.
- **Selective invalidation**: Mutating methods (`PUT`, `POST`, `DELETE`) invalidate cached metadata. GET requests preserve cached metadata.
- **Copy/multipart safety**: Do not prime from query-string PUTs or copy operations; multipart/query mutations invalidate affected metadata.
- **Size bound**: Cache is limited to 1,000 entries with TTL-aware eviction, falling back to FIFO eviction if still over limit.

### Cache lifecycle (`sandbox.ts`)

- `evictHeadMetadataCacheForMount` is called during unmount, mount-failure cleanup, and sandbox teardown, matching the existing SigV4 client and directory-marker cache cleanup paths.

### Request forwarding safety

- Strip hop-by-hop/proxy headers before forwarding credential-proxy requests upstream.
- Preserve SigV4 request bodies that do not include `content-length` instead of dropping the stream.
- Keep SigV4 signing on the existing `aws4fetch` `AwsClient` path.

## Benchmark results (from repro)

| Step | Direct S3 | Credential Proxy | Delta |
|------|-----------|-------------------|-------|
| read-small (1 KiB) | 133ms | 206ms | +73ms |
| read-large (512 KiB) | 69ms | 100ms | +31ms |
| read-large-repeat (5x) | 388ms | 285ms | **-103ms** |
| cached-head 20x reads | 2102ms | 1297ms | **-805ms** |
| list-files | 1898ms | 107ms | **-1791ms** |

Credential-proxy now avoids redundant upstream HEAD requests on repeated metadata reads.

### DIFF
--- a/packages/sandbox/src/storage-mount/s3-credential-proxy-handler.ts
+++ b/packages/sandbox/src/storage-mount/s3-credential-proxy-handler.ts
@@ -19,6 +19,7 @@ import type {
   OutboundHandlerContext
 } from '@cloudflare/containers';
 import type { BucketProvider } from '@repo/shared';
+import { AwsClient } from 'aws4fetch';
 import type { S3CredentialProxyParams } from './types';
 
 const PER_MOUNT_SUFFIX = '.s3-credential-proxy.internal';
@@ -82,7 +83,7 @@ const HEAD_METADATA_PUT_RESPONSE_HEADERS = new Set([
   'x-goog-storage-class'
 ]);
 
-const SIGV4_SKIPPED_FORWARD_HEADERS = new Set([
+const SKIPPED_FORWARD_HEADERS = new Set([
   'connection',
   'expect',
   'host',
@@ -94,18 +95,6 @@ const SIGV4_SKIPPED_FORWARD_HEADERS = new Set([
   'transfer-encoding',
   'upgrade',
   'user-agent'
-]);
-
-const SIGV4_UNSIGNABLE_HEADERS = new Set([
-  'authorization',
-  'connection',
-  'content-length',
-  'content-type',
-  'expect',
-  'presigned-expires',
-  'range',
-  'user-agent',
-  'x-amzn-trace-id'
 ]);
 
 export const DUMMY_AUTH_HEADERS = new Set([
@@ -146,13 +135,12 @@ type CredentialProxyDiagnosticEvent = CredentialProxyRequestInfo & {
 };
 
 type SigV4ClientCacheEntry = {
+  client: AwsClient;
   accessKeyId: string;
   secretAccessKey: string;
   endpoint: string;
   provider: BucketProvider | null;
   region: string;
-  date: string;
-  key: ArrayBuffer;
 };
 
 type HeadMetadataCacheEntry = {
@@ -397,7 +385,7 @@ function buildCleanHeaders(original: Headers): Headers {
     // Strip dummy signing headers and the intercepted proxy host; x-amz-content-sha256
     // is stripped here too (it's in DUMMY_AUTH_HEADERS) so it can be re-added below
     // only when the original value is a genuine payload hash (not a dummy signing value).
-    if (!DUMMY_AUTH_HEADERS.has(lower) && lower !== 'host') {
+    if (!DUMMY_AUTH_HEADERS.has(lower) && !SKIPPED_FORWARD_HEADERS.has(lower)) {
       clean.set(k, v);
     }
   }
@@ -518,14 +506,13 @@ async function withCredentialProxyDiagnostics(
   }
 }
 
-async function getSigV4SigningKey(
+function getSigV4Client(
   mountId: string,
   endpoint: string,
   provider: BucketProvider | null,
   credentials: { accessKeyId: string; secretAccessKey: string },
-  region: string,
-  date: string
-): Promise<ArrayBuffer> {
+  region: string
+): AwsClient {
   const cached = sigV4ClientCache.get(mountId);
   if (
     cached &&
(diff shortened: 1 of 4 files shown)


aron-cf (commented):
Thanks for digging into the HEAD traffic. Did you look at tweaking s3fs’s own stat cache? I don't really want to add an additional layer when the one in s3fs should be doing this already (unless I've missed something).

Right now sandbox-sdk sets these defaults for R2-backed s3fs mounts:

```ts
stat_cache_expire: '60',
enable_noobj_cache: true,
multipart_size: '5'
```

s3fs also supports `max_stat_cache_size`, `stat_cache_expire`, and negative caching (`enable_negative_cache` / `disable_negative_cache`; negative cache appears to be enabled by default in current s3fs).

Have we tried to repro with larger s3fs options first? for example:

```ts
s3fsOptions: [
  'stat_cache_expire=300',
  'max_stat_cache_size=100000',
  'enable_negative_cache'
]
```

packages/sandbox/src/storage-mount/s3-credential-proxy-handler.ts
```diff
@@ -28,6 +28,74 @@ export const DIAGNOSTICS_PATH = '/__sandbox_credential_proxy_diagnostics__';
 const DEFAULT_SLOW_REQUEST_MS = 1000;
 const ERROR_RESPONSE_BODY_LIMIT = 2048;
 const MAX_DIAGNOSTIC_EVENTS = 500;
+const HEAD_METADATA_CACHE_TTL_MS = 60_000;
```
aron-cf, line 31: This feels risky, if something updates/deletes the object, we keep serving the old HEAD result for up to a minute. Could we make this shorter or configurable... this works well for cases where this bucket is not changing often but if the bucket is shared with another writer then this will get problematic.

packages/sandbox/src/storage-mount/s3-credential-proxy-handler.ts
```diff
@@ -28,6 +28,74 @@ export const DIAGNOSTICS_PATH = '/__sandbox_credential_proxy_diagnostics__';
...
+const NEGATIVE_HEAD_METADATA_CACHE_TTL_MS = 5_000;
```
aron-cf, line 32: Ditto here, this means an object created externally will be missing for 5s.

packages/sandbox/src/storage-mount/s3-credential-proxy-handler.ts
```diff
@@ -93,6 +169,173 @@ export function evictDirectoryMarkerCacheForMount(mountId: string): void {
...
+export function evictHeadMetadataCacheForMount(mountId: string): void {
+  const prefix = `${mountId}:`;
+  for (const key of Array.from(headMetadataCache.keys())) {
+    if (key.startsWith(prefix)) {
+      headMetadataCache.delete(key);
+    }
+  }
+}
+
+function getHeadMetadataCacheKey(
```
aron-cf, line 181: The cache key is scoped by `mountId`, not by the underlying bucket/endpoint identity, this doesn't feel right. If you have the same bucket mounted in two places the cache should be the same right?

packages/sandbox/src/storage-mount/s3-credential-proxy-handler.ts
```diff
@@ -93,6 +169,173 @@ export function evictDirectoryMarkerCacheForMount(mountId: string): void {
...
+  mountId: string,
+  realPath: string,
+  url: URL
+): string {
+  return `${mountId}:${realPath}${url.search}`;
+}
+
+function getCachedHeadMetadataResponse(
```
aron-cf, line 189: This  defeats s3fs’s open-time revalidation. s3fs intentionally drops its own stat cache on reopen and sends a HEAD; with this cache, that HEAD may be answered locally instead of actually revalidating against the bucket. I think this behaviour in s3fs is probably what led to this PR. But it makes me think that this cache layer should be opt-in and extremely configurable.

packages/sandbox/src/storage-mount/s3-credential-proxy-handler.ts
```diff
@@ -93,6 +169,173 @@ export function evictDirectoryMarkerCacheForMount(mountId: string): void {
...
+        : HEAD_METADATA_CACHE_TTL_MS),
+    headers: Array.from(
+      response.headers as unknown as Iterable<[string, string]>
+    ),
+    status: response.status,
+    statusText: response.statusText
+  });
+}
+
+function cacheHeadMetadataFromPUT(
```
aron-cf, line 250: This is neat. How do we ensure that we're accurately matching what the provider actually responds with? Feels like this might need an integration test to verify our code matches the various provider implementations.

packages/sandbox/src/storage-mount/s3-credential-proxy-handler.ts
```diff
@@ -813,8 +1064,36 @@ export const s3CredentialProxyHandler: OutboundHandler<
...
+        requestInfo,
+        debugConfig,
+        ctx.containerId,
+        realPath,
+        () => Promise.resolve(cachedHeadResponse)
+      );
+    }
+  } else if (isMutatingMethod(method)) {
     deleteDirectoryMarkerCacheEntry(mountId, realPath);
+    deleteHeadMetadataCacheEntriesForObject(mountId, realPath);
```
aron-cf, line 1096: For non-HEAD mutating requests, we invalidate before forwarding. There’s a possible race where a HEAD during an in-flight DELETE/POST/copy/multipart operation can cache an old upstream metadata, and then the mutation succeeds. Should we also invalidate _after_ successful mutations?

packages/sandbox/src/storage-mount/s3-credential-proxy-handler.ts
aron-cf, line 1024: The directory marker cache looks like it has no TTL. External changes can leave us with stale directory metadata for the lifetime of the mount unless something local invalidates it. Can we add the same invalidation pattern?

aron-cf (changes requested):
After looking at s3fs I think this is probably useful for common cases where a bucket is mounted and owned by a single container or the bucket is largely read only.

It's a bit risky for anything intended to be shared collaboratively. I'd suggest we roll this out under an opt-in flag, with configurable timeouts and document which usecases will benefit and which wont.

PR #766, opened 2026-06-18
Implement Sandbox extension framework

scuffi:
## Summary

Adds an opt-in extension framework so capabilities can be attached to a `Sandbox` subclass instead of being bolted onto core or baked into the container image.

Everything uses one base class, `SandboxExtension`, wired in with a `withX(this)` factory. There are two flavours, decided purely by whether you pass a tarball:

- **SDK-only**: orchestrates existing container APIs (`commands`, `files`, `processes`, etc). No extra runtime.
- **Sidecar**: ships an npm-style `.tgz`. The container provisions it lazily on first use, spawns it as a supervised process, and talks to it over capnweb on a private unix socket.

The interpreter is the first real consumer (stacked PR #767).

## Usage

SDK-only extension, just drive the container APIs:

```ts
import { Sandbox as BaseSandbox } from '@cloudflare/sandbox';
import {
  SandboxExtension,
  type SandboxLike
} from '@cloudflare/sandbox/extensions';

class Greeter extends SandboxExtension {
  constructor(sandbox: SandboxLike) {
    super(sandbox);
  }

  async greet(name: string, sessionId: string) {
    const { stdout } = await this.client.commands.execute(
      `echo "hi ${name}"`,
      sessionId
    );
    return stdout.trim();
  }
}
export const withGreeter = (s: SandboxLike) => new Greeter(s);

export class Sandbox extends BaseSandbox<Env> {
  greeter = withGreeter(this);
  greet(name: string, sessionId: string) {
    return this.greeter.greet(name, sessionId);
  }
}
```

Sidecar extension, same shape, but pass tarball bytes and grab a typed capnweb remote with `this.sidecar<T>()`:

```ts
import sidecarTarballBytes from './sidecar-package.tgz';
import {
  SandboxExtension,
  type SandboxLike
} from '@cloudflare/sandbox/extensions';
import type { MySidecarAPI } from './shared';

class MyExtension extends SandboxExtension {
  constructor(sandbox: SandboxLike) {
    super(sandbox, { tarball: new Uint8Array(sidecarTarballBytes) });
  }

  async run(input: string) {
    const api = await this.sidecar<MySidecarAPI>();
    return api.run(input);
  }
}
export const withMyExtension = (s: SandboxLike) => new MyExtension(s);
```

Streaming is just a typed callback parameter on a sidecar method. capnweb stubs the callback across both hops, so there's no separate streaming protocol to learn.

## Changes

**SDK** (`@cloudflare/sandbox/extensions`)
- `SandboxExtension`: `this.client` for container APIs, plus `sidecar()` / `sidecarHealth()` / `stopSidecar()` for sidecar-backed extensions (these throw if no package was passed).
- Hash-first connect: the SDK sends only the tarball hash, and only resends bytes if the host process hasn't provisioned that hash yet.

**Sidecar authoring** (`@cloudflare/sandbox/sidecar`)
- Extend `SandboxSidecar`, call `serveSandboxSidecar(...)`. It listens on the `EXT_SOCKET` unix socket the host injects and serves a capnweb session per connection (`__ping__` reserved for health).

**Container** (`packages/sandbox-container/src/extensions/`)
- `ExtensionHost` provisions by tarball content hash, derives identity from the embedded `package.json`, runs `bun add ./extension.tgz`, and spawns the resolved bin under Bun.
- Sidecars spawn lazily, are supervised, and restart transparently on the next `sidecar()` call after a crash. Unique socket path per spawn plus generation guards keep lifecycle transitions safe.
- capnweb over a private unix socket, no exposed ports or HTTP.

## Compatibility

Additive. New opt-in subpaths only; no change to existing `Sandbox` behaviour. Third-party publishing/install tooling is deliberately out of scope for now: sidecar tarballs are bundled with the app, and a public authoring flow can build on the same hash-first provisioning later. Changeset: `patch` (`add-sandbox-extensions`).


packages/sandbox-container/src/control-plane/api.ts
```diff
@@ -1129,3 +1135,67 @@ class TunnelsRPCAPI extends RpcTarget {
     return this.#svc.list();
   }
 }
+
+// ===========================================================================
+// Extensions (dynamic sidecar bridge)
+// ===========================================================================
+
+/**
+ * Level 2 dynamic capnweb surface for sidecar extensions.
```
aron-cf: Try and avoid sticking plan/document milestones in comments

packages/sandbox-container/src/extensions/provision.ts
```diff
@@ -0,0 +1,420 @@
...
+ * Read `package/package.json` from an npm-style tarball. Tarballs produced by
+ * `npm pack` / `bun pm pack` always nest contents under a `package/` prefix.
+ *
+ * Implemented in-process to avoid shelling out to `tar` and to keep the host
+ * runtime independent of the container image's userland tools.
+ */
+function readPackageJSONFromTarball(
+  tarballBytes: Uint8Array
+): SandboxExtensionPackageJSON {
+  const inflated = gunzipSync(tarballBytes);
```
aron-cf: Is there an async version of this that processes a stream? There's tar stream package for example that let's you iterate over a tarball without buffering everything into memory. Generally this is going to be fine on the container as there's usually a decent amount of memory but still it's good to get into the habit.

packages/sandbox-container/src/extensions/socket-transport.ts
```diff
@@ -0,0 +1,104 @@
+import type { Socket } from 'node:net';
```
aron-cf, line 1: We're using a lot of node apis instead of the native bun implementations. I don't think we need the portability and the Bun implementations are going to be significantly as they're not going through a compatibility layer. Can we switch them out?

packages/sandbox-container/src/extensions/extension-host.ts
```diff
@@ -0,0 +1,432 @@
...
+ * session B); capnweb proxies methods and callback stubs across both hops.
+ *
+ * Startup-optimised: unused extensions cost ~nothing. A connect for an
+ * unknown hash with no tarball attached fails with
+ * {@link ExtensionTarballRequiredError} so the SDK can resend with bytes;
+ * once provisioned, the hash alone is enough for every subsequent connect.
+ */
+export class ExtensionHost {
+  readonly #instancesByHash = new Map<string, ExtensionInstance>();
+  readonly #provisioningByHash = new Map<string, Promise<ExtensionInstance>>();
```
aron-cf, line 69: If we keep this in memory each container restart is going to redownload/install the plugins. I wonder if instead track this on disk. Either by keeping this mapping on disk or scanning existing files.

packages/sandbox-container/src/extensions/socket-transport.ts
```diff
@@ -0,0 +1,104 @@
...
+ * sees the same `send(string)` / `receive(): Promise<string>` interface it
+ * uses for every other transport.
+ *
+ * Lifetimes:
+ * - One transport per socket per capnweb session.
+ * - On socket close/error, in-flight `receive()` callers reject and any
+ *   subsequent receive throws the cached error. capnweb tears the session
+ *   down from there.
+ */
+export class SocketTransport implements RpcTransport {
```
aron-cf, line 26: This is great! Only thing I'd say is it's probably worth adding some basic back-pressure support to pause/resume the socket when it gets overloaded. If you build this on top of ReadableStream/WriteableStream/TransformStream etc you'll get this out of the box.

aron-cf (changes requested):
@scuffi I think this is pretty close. I'd be happy to merge this as-is if you want to address comments as a follow up. I don't have any suggestions to change the API right now, just some internal performance/security bits.

I think the big piece here is that the entire socket communication & filesystem is open, but for the moment so is the rest of the container side implementation so I think that just needs to be addressed wholesale.

packages/sandbox-container/src/extensions/socket-transport.ts
scuffi: I've switched out in favour of Bun in most cases, but there are still some cases that don't have 1:1 APIs and will require a bigger re-work. I'll have more of a look into this in future

aron-cf (approved):
Looks freaking great, good job, let's merge it and iterate.

PR #767, opened 2026-06-18
Migrate code interpreter into a `SandboxExtension`

scuffi:
## Summary

Moves the code interpreter out of the core SDK and the container binary into a standalone, opt-in extension: `@cloudflare/sandbox/interpreter`. Its entire container-side runtime (the process pool + the Python/JavaScript executors) now ships as a **sidecar** that the container provisions on first use, instead of being compiled into the SDK or baked into the container image.

This doubles as the **reference implementation** for the extension framework.

## Why

- Keeps the core SDK and container image lean, you only pay for the interpreter if you opt in.
- Validates the extension framework end-to-end against a real, non-trivial feature.
- Establishes a copyable, self-contained extension layout.

## How it's structured

The whole extension is self-contained under `packages/sandbox/src/interpreter/`:

- `index.ts` => `Interpreter` (extends `SandboxExtension`), `withInterpreter()`, and `buildInterpreterManifest()`.
- `sidecar/` => the runtime that runs in the container: frame-protocol `server.ts`, the process pool, and the Python/JS executors.
- `build.ts` => bundles the sidecar + executors into the npm-style `.tgz` (a content-hash version drives idempotent re-provisioning).
- The sidecar runs under Node/Bun, so it's excluded from the Workers `tsconfig` and typechecked via its own `tsconfig.sidecar.json`.

## Calling extensions from a Worker

This PR also updates the Sandbox RPC proxy path so extensions can be called naturally from Workers, for example:
 
```ts
await sandbox.interpreter.runCode("1 + 1");
```
 
Worker -> DO RPC does not reliably property-pipeline nested calls like `sandbox.interpreter.runCode(...)`. To preserve the extension DX, `getSandbox()` now synthesises nested proxies for extension namespaces and forwards those calls to a top-level DO method:
 
```ts
callExtension(extensionName, method, args)
```
 
`Sandbox.callExtension()` only dispatches to actual `SandboxExtension` instances on the DO and rejects unknown extensions/methods. This mirrors the existing `callTunnels` workaround for nested RPC access.

So now we can attach the extension on your `Sandbox` subclass and call it directly as a nested namespace, no delegate methods required:

```ts
import { Sandbox as BaseSandbox } from '@cloudflare/sandbox';
import { withInterpreter } from '@cloudflare/sandbox/interpreter';

export class Sandbox extends BaseSandbox<Env> {
  interpreter = withInterpreter(this);
}

const context = await sandbox.interpreter.createCodeContext({ language: 'python' });
const result = await sandbox.interpreter.runCode('print("hello")', { context });
```

Under the hood `getSandbox()` maps `sandbox.<extension>.<method>(...)` to a top-level DO dispatch, so this works across the Worker/DO boundary without relying on property pipelining. You can still add your own delegate methods (e.g. `sandbox.runCode(...)`) if you want a flatter API, but they're optional.

One rule worth calling out: extension methods called from a Worker have to return serializable data, so `runCode()` now resolves an `ExecutionResult` (plain object) rather than an `Execution` class instance.

## Breaking change (`minor`)

The interpreter API moves off `Sandbox` directly and onto the extension. `createCodeContext`, `runCode`, `runCodeStream`, `listCodeContexts`, and `deleteCodeContext` (and their types) are otherwise unchanged, aside from `runCode` returning `ExecutionResult` directly. Python execution still requires the `-python` image variant.

> **Stacked on #766** (the extension framework implementation). Review that first, this PR builds on it and is the first real consumer of the framework.


examples/code-interpreter/Dockerfile
```diff
@@ -1 +1 @@
-FROM docker.io/cloudflare/sandbox:0.12.1-python
+FROM cloudflare/sandbox-test:0.12.1-python
```
aron-cf: Back this out

packages/sandbox-container/build.ts
```diff
@@ -3,13 +3,9 @@
...
  * - dist/sandbox: Standalone binary for /sandbox entrypoint
  * - dist/index.js: Legacy JS bundle for backwards compatibility
- * - dist/runtime/executors/javascript/node_executor.js: JS executor
  */
 
-import { mkdir } from 'node:fs/promises';
-
-// Ensure output directories exist
-await mkdir('dist/runtime/executors/javascript', { recursive: true });
+export {};
```
aron-cf, line 10: What is this doing? If nothing is exported shall we just remove it?

packages/sandbox/src/extensions/index.ts
```diff
@@ -171,8 +175,36 @@ export abstract class SandboxExtension extends RpcTarget {
...
   if (typeof error !== 'object' || error === null) return false;
-  const name = (error as { name?: unknown }).name;
-  return name === EXTENSION_TARBALL_REQUIRED;
+  const candidate = error as {
+    name?: unknown;
+    message?: unknown;
+    code?: unknown;
+  };
+  if (candidate.name === EXTENSION_TARBALL_REQUIRED) return true;
+  if (candidate.code === EXTENSION_TARBALL_REQUIRED) return true;
```
aron-cf: Why two checks? Which one is supposed to use `EXTENSION_TARBALL_REQUIRED`?

packages/sandbox/src/interpreter/sidecar/executors/javascript/node_executor.ts
```diff
@@ -52,19 +52,19 @@ rl.on('line', async (line: string) => {
     let stdout = '';
     let stderr = '';
 
-    (process.stdout.write as any) = (
+    (process.stdout.write as unknown) = (
       chunk: string | Buffer,
-      encoding?: BufferEncoding,
+      _encoding?: BufferEncoding,
```
aron-cf: Should we not be using this encoding when decoding the chunk into a string?

packages/sandbox/src/interpreter/build.ts
```diff
@@ -0,0 +1,156 @@
+import { createHash } from 'node:crypto';
+import { readFile, writeFile } from 'node:fs/promises';
+import { dirname, join } from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { gzipSync } from 'node:zlib';
```
aron-cf: This is going to run under the bun runtime, so again I think where possible we should use the bun standard library before dropping down into nodejs

packages/sandbox/src/interpreter/build.ts
```diff
@@ -0,0 +1,156 @@
...
+  )} KB)`
+);
+
+interface TarEntry {
+  path: string;
+  content: string | Uint8Array;
+  mode?: number;
+}
+
+function createTarball(entries: TarEntry[]): Buffer {
```
aron-cf: I think it's okay to use a library for this?

packages/sandbox/src/interpreter/tgz-module.ts
```diff
@@ -0,0 +1,4 @@
+declare module '*.tgz' {
```
aron-cf, line 5: What does this do?

packages/sandbox/src/sandbox.ts
```diff
@@ -977,6 +1028,27 @@ export class Sandbox<Env = unknown> extends Container<Env> implements ISandbox {
...
+    ];
+    if (!(extension instanceof SandboxExtension)) {
+      throw new Error(`Unknown sandbox extension: ${extensionName}`);
+    }
+    const fn = getConcreteExtensionMethod(extension, method);
+    if (!fn) {
+      throw new Error(`Unknown extension method: ${extensionName}.${method}`);
+    }
+    return fn.apply(extension, args);
+  }
```
aron-cf, line 1050: Does this exist because calling methods on the extension doesn't work correctly over RPC?

aron-cf (commented):
Looking good. Some questions inline. Can you add a section to the PR description about the non-interpreter changes to the Sandbox class. It looks like we had to make some modifications to the extensions API to support calling the interpreter over RPC but it's not clear from the code why.

packages/sandbox-container/build.ts
scuffi: added a comment - we need it so TS marks it as a module

packages/sandbox/src/extensions/index.ts
scuffi: updated

packages/sandbox/src/interpreter/build.ts
scuffi: updated to bun apis

packages/sandbox/src/interpreter/tgz-module.ts
scuffi: added a comment - TS declaration so we can import the sidecar package

packages/sandbox/src/sandbox.ts
scuffi: yes exactly, I'll update the PR description more to outline these changes

aron-cf (approved):
🚀 shipit!

PR #775, opened 2026-06-22
Ignore exec AbortSignal when RPC flag is unset

### LINKED ISSUE #764: Passing AbortSignal to exec() throws across DO RPC
### Version
`@cloudflare/sandbox@0.12.1`

### Summary
`ExecOptions` accepts `signal?: AbortSignal`, but passing one to `getSandbox(...).exec(command, { signal })` throws before the command runs:

    DataCloneError: AbortSignal serialization is not enabled.

The options object is structured-cloned across the Worker→DO RPC boundary, and `AbortSignal` isn't cloneable, so the call fails rather than the signal being honored or ignored.

### Repro
From a Worker with a Sandbox binding:

```ts
const sandbox = getSandbox(env.Sandbox, 'repro');
const controller = new AbortController();
await sandbox.exec('echo hi', { signal: controller.signal });
// -> DataCloneError: AbortSignal serialization is not enabled.
```

### Expected
Passing `signal` shouldn't throw — either it cancels the command or it's a documented no-op on this path.

aron-cf:
Fixes https://github.com/cloudflare/sandbox-sdk/issues/764

Calling `exec()` with an abort signal threw before the command ever ran:

```ts
const sandbox = getSandbox(env.Sandbox, 'repro');
const controller = new AbortController();
await sandbox.exec('echo hi', { signal: controller.signal });
// -> DataCloneError: AbortSignal serialization is not enabled.
```

The options object is structured-cloned when it crosses from the Worker into
the Durable Object. An `AbortSignal` is only cloneable when the
`enable_abortsignal_rpc` compatibility flag is set, so without that flag the
call failed instead of either honoring the signal or ignoring it. This is
[issue #764](https://github.com/cloudflare/sandbox-sdk/issues/764).

**The fix**

Before the call crosses the boundary, the software development kit now checks
whether an abort signal can actually be serialized. It does this with a local
`structuredClone` probe rather than letting the real call fail, because a
failed call also leaves an unhandled rejection inside the runtime's remote
procedure call machinery that the caller cannot reach.

When the flag is missing, the signal is dropped so the command still runs, and
a single warning explains how to turn cancellation on. When the flag is set,
the signal is passed through and honored exactly as before. The same guard is
applied to `execStream()`, which accepts a signal through the same options
shape. The warning fires once per isolate, since the flag cannot change while
the isolate is alive.

**How to verify**

With the flag enabled, cancellation works as expected:

```jsonc
// wrangler.jsonc
{
  "compatibility_flags": ["enable_abortsignal_rpc"]
}
```

Without the flag, the previously broken snippet from the issue now completes
instead of throwing, and the logs contain a one-time warning pointing at the
flag.

**Testing**

Compatibility flags are fixed for the lifetime of an isolate, so a single test
run cannot toggle the flag on and off. To cover both states, the regression
test runs under two test projects in the same suite: one boots the pool without
`enable_abortsignal_rpc` and one boots it with the flag. A binding tells each
run which state it is in so the same assertions describe both directions.

The test confirms three things. The raw, unguarded boundary still throws when
the flag is off, which documents the underlying issue. Going through
`getSandbox().exec()` never surfaces the clone error to the caller. And the
warning about the flag is emitted only when the flag is off.

A deliberate side effect of the flag-off case is an internal rejected promise
created by the runtime that no test code can await. The suite drops only that
exact error through the `onUnhandledError` hook, so every other unhandled
error across the whole suite still fails the run.

### DIFF
--- a/.changeset/exec-abortsignal-rpc.md
+++ b/.changeset/exec-abortsignal-rpc.md
@@ -1,0 +1,19 @@
+---
+'@cloudflare/sandbox': patch
+---
+
+Stop `exec()` from throwing when passed an `AbortSignal` without the
+`enable_abortsignal_rpc` compatibility flag. Previously this failed with
+`DataCloneError: AbortSignal serialization is not enabled.` because the signal
+could not cross the Worker → Durable Object boundary. The command now runs and
+a one-time warning explains how to enable cancellation:
+
+```jsonc
+// wrangler.jsonc
+{
+  "compatibility_flags": ["enable_abortsignal_rpc"]
+}
+```
+
+With the flag set, the signal is honored as before; without it, the signal is
+ignored instead of breaking the call.
--- a/packages/sandbox/src/sandbox.ts
+++ b/packages/sandbox/src/sandbox.ts
@@ -619,6 +619,68 @@ function applySandboxConfiguration(
   return Promise.all(operations).then(() => undefined);
 }
 
+/**
+ * Tracks whether the AbortSignal-not-serializable warning has already been
+ * emitted in this isolate, so we warn once per isolate rather than on every
+ * call. The compatibility flag is fixed for the isolate's lifetime, so a
+ * single notice is enough.
+ */
+let warnedAboutAbortSignalRPC = false;
+
+/**
+ * Detects whether an `AbortSignal` can cross the Worker -> Durable Object RPC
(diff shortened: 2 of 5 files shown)


ghostwriternr:
`AbortSignal` has been introduced on the `next` branch and will be supported on versions published from the `next` branch and subsequently be a part of the stable releases when it becomes the primary release version. Closing this PR as superseded by those changes.

Issue #158, opened 2025-10-28
DOs and containers getting reset during deployment/rollouts needs to be handled better


ghostwriternr:
https://developers.cloudflare.com/containers/platform-details/rollouts/

mikenomitch:
Hmmm... so what would we need beyond the following:

- End user needs to be able to nicely configure (and docs explain) the grace period setting
- The Container itself should handle SIGTERM and log that it got a kill signal

I think you want everything in the container to keep running normally right?

aron-cf:
@ghostwriternr nine-months on where are we with this issue? It's currently not very actionable. I'm going to close it, but feel free to expand the description and re-open.

ghostwriternr:
Agreed, will re-open later if this continues to be relevant, but I think a lot of the lifecycle related work that I've been landing last month and continuing to do (as recently as #763) are all in this general direction - so the blanket issue here is no longer necessary.

PR #779, opened 2026-06-24
Recover from sandbox runtime replacement on stable

ghostwriternr:
The Cloudflare runtime can replace a sandbox's container while SDK-owned work is in flight — most often during a deployment rollout (see #158). When that happens on the stable line today, in-flight operations surface raw transport failures: `createDir(...)` and other setup calls fail with `RPCTransportError: RPC session was shut down by disposing the main stub`, `restoreBackup()` returns a generic 500, and `tunnels.get()` can hand back a URL that no longer routes. These are the same failures we see intermittently in CI, and they affect users on the latest release now.

The proper fixes for this live on `next` (#763 backup, #776 tunnels), but they build on the beta transport removal and large backup/tunnel decompositions and won't reach `main` for several weeks. This PR ports the *behavior* of those fixes to the stable line as a surgical hotfix, reusing the same public names and durable concepts (`CONTAINER_UNAVAILABLE`, `OPERATION_INTERRUPTED`, runtime-identity fencing, tunnel `runId`) so the eventual `next` → `main` merge reconciles concepts rather than fighting stable-only terminology.

## Behavior changes

- **Container not yet ready vs. interrupted mid-operation are now distinct.** A failed `/rpc` upgrade discards the poisoned connection, preserves locally-raised SDK errors, and surfaces a retryable `CONTAINER_UNAVAILABLE` (503) once bounded upgrade retries are exhausted. This covers the file-watch / `createDir` setup failures seen during rollout.
- **`restoreBackup()` recovers from runtime replacement.** Restore is fenced by sandbox lifetime and runtime identity and retried boundedly when the runtime is replaced mid-restore. Recovery exhaustion surfaces as `OPERATION_INTERRUPTED` (409) instead of a generic 500.
- **`tunnels.get(port)` returns a usable quick tunnel after replacement.** Quick-tunnel provisioning is now idempotent across retries, and `tunnels.list()` no longer returns quick URLs known to be stale after restart reconciliation.

No public method signatures change. The new errors and a few tunnel run types are additive exports.

## Scope and what this deliberately does not do

This is a hotfix, not the `next` architecture. To stay stable-safe it keeps the existing files in place: recovery lands in `sandbox.ts` and `tunnels-handler.ts` rather than extracting a `BackupService` or splitting the tunnel subsystem into service/provisioner/rpc-target modules. Recovery is bounded (at most two retry attempts) and scoped to SDK-owned operations — there are no blanket retries around `exec()`, writes, renames, or other user operations that may have side effects. Operations are not retried across `destroy()` or a sandbox lifetime change.

One known limitation: named-tunnel durable cleanup intent (Track B on `next`) is out of scope here; this hotfix does not regress current stable named-tunnel behavior but also does not add the stronger cleanup guarantees landing on `next`.

### DIFF
--- a/.changeset/stable-runtime-recovery.md
+++ b/.changeset/stable-runtime-recovery.md
@@ -1,0 +1,5 @@
+---
+'@cloudflare/sandbox': patch
+---
+
+Improve recovery when a sandbox runtime is replaced during active work, reducing transient failures for stable SDK users.
--- a/packages/sandbox-container/src/control-plane/api.ts
+++ b/packages/sandbox-container/src/control-plane/api.ts
@@ -1,6 +1,8 @@
 import type {
   CheckChangesRequest,
   CheckChangesResult,
+  EnsureTunnelRunRequest,
+  EnsureTunnelRunResult,
   ExecutionError,
   FileEncoding,
   FileInfo,
@@ -9,6 +11,8 @@ import type {
   OutputMessage,
   Result,
   SandboxAPI,
+  StopTunnelRunRequest,
+  StopTunnelRunResult,
   TunnelInfo,
   WatchRequest
 } from '@repo/shared';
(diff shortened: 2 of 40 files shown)


ghostwriternr:
Superseded by a reviewable 3-PR stack that splits this work by failure domain (each piece compiles and tests independently, in dependency order):

- #780 — Classify stable container unavailability (base `main`)
- #781 — Recover stable backup restores (stacked on #780)
- #782 — Recover stable quick tunnels (stacked on #781)

Same final tree as this branch; closing in favor of the stack.

PR #780, opened 2026-06-24
Classify stable container unavailability

ghostwriternr:
**Stack:** 1 of 3 — base of the stable runtime-recovery series. Followed by quick-tunnel and backup-restore recovery, which build on the error contract introduced here.

The Cloudflare runtime can start, replace, or briefly drop a sandbox's container while the SDK is connecting to it — most often during a deployment rollout (see #158). Today these moments surface as raw transport failures: setup calls like `createDir(...)` fail with `RPCTransportError: RPC session was shut down by disposing the main stub`, and a default-session `createSession` RPC that lands after a container stop fails the caller outright. These are the same failures we see intermittently in CI, and they affect users on the latest release now.

This PR establishes the shared lifecycle error vocabulary for the series and fixes the admission/connection layer:

- **`CONTAINER_UNAVAILABLE` (503, retryable)** — raised when the container is not yet ready or was replaced during a connection attempt. A failed `/rpc` upgrade now discards the poisoned connection and preserves locally-raised SDK errors instead of leaking transport internals.
- **`OPERATION_INTERRUPTED` (409)** — in-flight RPC calls whose session is disposed now surface as an operation interruption with `admitted: "unknown"` and `retryable: false`, instead of leaking capnweb's `main stub` transport message. This gives callers a handleable contract: retry only idempotent work, or verify sandbox state first.
- **Default-session recovery** — when a `createSession` RPC is invalidated by a container generation change, the default session retries once against the new container rather than failing the originating call.
- **Platform update classification** — Durable Object code-update resets (`Durable Object reset because its code was updated` / `This script has been upgraded`) are classified as lifecycle interruptions. Enhanced `getSandbox()` methods now surface them as `OPERATION_INTERRUPTED` with `admitted: "unknown"` instead of leaking platform strings.
- **E2E initialization recovery** — the test helper retries only its default `echo ready` probe when it receives `CONTAINER_UNAVAILABLE`, `OPERATION_INTERRUPTED`, or a classified platform-update reset, modeling the safe retry pattern users should apply to idempotent operations.

No public method signatures change; the new errors are additive exports.

## Why this targets the stable line

The proper fixes live on `next` (#763 backup, #776 tunnels), but they build on the beta transport removal and large backup/tunnel decompositions and won't reach `main` for several weeks. This series ports the *behavior* to stable as a surgical hotfix, reusing the same public names (`CONTAINER_UNAVAILABLE`, `OPERATION_INTERRUPTED`) so the eventual `next` → `main` merge reconciles concepts rather than fighting stable-only terminology.


packages/sandbox/src/container-control/client.ts
```diff
@@ -311,6 +330,34 @@ function buildTransportErrorResponse(
...
+    message: `Sandbox operation ${context.operation} was interrupted while the runtime connection was closing`,
+    context: interruptedContext,
+    httpStatus: getHttpStatus(ErrorCode.OPERATION_INTERRUPTED),
+    suggestion: getSuggestion(
+      ErrorCode.OPERATION_INTERRUPTED,
+      interruptedContext as unknown as Record<string, unknown>
+    ),
+    timestamp: new Date().toISOString()
+  };
+}
```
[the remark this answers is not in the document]
ghostwriternr: Mapped in-flight `peer_closed` and `connection_failed` transport failures to `OPERATION_INTERRUPTED` when the RPC translator has operation context. `upgrade_failed` stays on the pre-admission/container-unavailable path. Added regression coverage in `packages/sandbox/tests/rpc-client-retry.test.ts`.

.changeset/stable-container-unavailable.md
```diff
@@ -0,0 +1,5 @@
+---
+'@cloudflare/sandbox': patch
+---
+
+Classify sandbox runtime and platform update interruptions as structured errors instead of surfacing raw transport or Durable Object messages. This gives applications clear retry guidance for idempotent operations and reduces transient startup failures when a sandbox is starting up or being replaced during deployment rollouts.
```
aron-cf: Can we make this shorter and lead with the bigger item which is reducing errors.

```suggestion
Reduces startup failures when a sandbox is booting or being replaced during container deployment. As well as providing structured errors for sandbox runtime and platform interruptions including retry guidance for idempotent operations.
```

packages/sandbox/src/container-control/connection.ts
```diff
@@ -20,12 +20,72 @@ import type {
...
+/**
+ * Attempt to parse a structured CONTAINER_UNAVAILABLE JSON body from a
+ * non-101 upgrade response. Returns a typed ContainerUnavailableError when
+ * the body contains a recognized code, or null if the response is not
+ * JSON / not a container availability error.
+ *
+ * The response body is consumed only once via `clone()` so the caller still
+ * holds the original Response.
+ */
+async function tryParseContainerUnavailable(
```
aron-cf, line 51: This is super nice

packages/sandbox/src/platform-errors.ts
aron-cf, line 1: 😍

aron-cf (approved):
This is great. Really nice work. I think it was worth extracting it and backporting it.

PR #781, opened 2026-06-24
Recover stable backup restores

ghostwriternr:
**Stack:** 2 of 3 — builds on the lifecycle error contract from the container-unavailable PR. Open that one first.

When the Cloudflare runtime replaces a sandbox's container while a backup restore is in progress, `restoreBackup()` fails with a generic 500 and leaves the caller unsure whether the restore committed. This is one of the runtime-replacement failure modes seen during deployment rollouts (see #158).

This PR runs restore through a lifecycle that makes it recoverable:

- Each attempt is **fenced by sandbox lifetime and runtime identity**, so an attempt that started against a now-replaced runtime is detected rather than silently writing to the wrong container.
- Runtime replacement mid-restore is **retried boundedly** (at most two recovery attempts).
- Restore is **not retried across `destroy()`** or a sandbox lifetime change.
- When bounded recovery is exhausted, the caller gets a structured **`OPERATION_INTERRUPTED` (409)** with retry guidance instead of an opaque failure.

## Scope

This is a stable hotfix, not the `next` architecture (#763). Recovery lands in the existing `sandbox.ts` restore path plus a narrow lifecycle runner and operation-record store, rather than extracting a `BackupService`. Recovery is scoped to this SDK-owned operation — there are no blanket retries around `exec()`, writes, renames, or other user operations that may have side effects.

### DIFF
--- a/.changeset/stable-backup-restore-recovery.md
+++ b/.changeset/stable-backup-restore-recovery.md
@@ -1,0 +1,5 @@
+---
+'@cloudflare/sandbox': patch
+---
+
+Recover `restoreBackup()` when the sandbox runtime is replaced mid-restore. Restore is fenced by sandbox lifetime and runtime identity and retried boundedly when the runtime changes during the operation; if recovery is exhausted it surfaces as a structured `OPERATION_INTERRUPTED` error instead of a generic 500. Restores are not retried across `destroy()`.
--- a/.changeset/stable-backup-restore-recovery.md
+++ b/.changeset/stable-backup-restore-recovery.md
@@ -2,4 +2,4 @@
 '@cloudflare/sandbox': patch
 ---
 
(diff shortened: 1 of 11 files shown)


.changeset/stable-backup-restore-recovery.md
```diff
@@ -0,0 +1,5 @@
+---
+'@cloudflare/sandbox': patch
+---
+
+Recover `restoreBackup()` when the sandbox runtime is replaced mid-restore. Restore is fenced by sandbox lifetime and runtime identity and retried boundedly when the runtime changes during the operation; if recovery is exhausted it surfaces as a structured `OPERATION_INTERRUPTED` error instead of a generic 500. Restores are not retried across `destroy()`.
```
aron-cf: I think this can simply be:
```suggestion
Restoring a backup now survives a durable object restart.
```

What about creating a backup though? Is that already fixed?

aron-cf (commented):
Minor comments, I can't figure out why CI is broken. It looks legit but I'm not sure what part of this change broke it...

scuffi (approved):

PR #786, opened 2026-06-26
Improve scaling of bridge warm pool under load

aron-cf:
Benchmarking the bridge warm pool surfaced two reliability problems and one speed problem. A cold burst of one hundred sandboxes failed roughly a third of the time, a primed pool still developed a latency tail beyond ten seconds for requests that overflowed the warm target, and priming an empty pool took over a minute because containers started one at a time. Underneath all of this, bridge requests were largely invisible in traces.

The pool now accepts its capacity ceiling up front through the `WARM_POOL_MAX_INSTANCES` variable instead of discovering it by failing a start and parsing the error, so its capacity math is correct from the first request. It refills the instant a warm container is taken rather than waiting for the next background sweep, with a guard so concurrent requests trigger at most one refill at a time and never exceed the ceiling. It also fills in small parallel batches instead of one container at a time, with capacity re-checked between batches; the batch size defaults to five and is tunable through `WARM_POOL_SCALE_BATCH_SIZE`. Together these took the same hundred-sandbox burst from a dozen slow sandboxes to none, with wall-clock time roughly halved.

Every bridge request is now wrapped in a custom span named for its operation and annotated with the sandbox identifier, the container identifier, and call-specific metadata such as the command, file path, tunnel port, or session. Tracing is enabled in the worker configuration with a tunable sampling rate, the container instance size is raised to `standard-1`, and the span instrumentation falls back to doing nothing where the runtime does not support it.

To verify, set `WARM_POOL_TARGET` and `WARM_POOL_MAX_INSTANCES`, deploy, and watch `GET /v1/pool/stats` during a burst: the reported ceiling is correct before any start is attempted and the warm count recovers immediately rather than after the refresh interval. Spans for each operation appear in the dashboard carrying the sandbox identifier and metadata. The change is covered by unit tests for the seeded ceiling, eager refill and its concurrency guard, batched scale-up, and the tracing helper, with the existing bridge and software development kit suites unchanged.

The streaming command and terminal endpoints record their setup metadata but close their spans when the handler returns rather than when the stream completes, a limitation of the callback-scoped span interface.

### DIFF
--- a/.changeset/bridge-custom-spans.md
+++ b/.changeset/bridge-custom-spans.md
@@ -1,0 +1,9 @@
+---
+'@cloudflare/sandbox': patch
+---
+
+Instrument the bridge API with Cloudflare custom spans. Each endpoint now emits
+a `bridge.<operation>` span annotated with the sandbox ID, container UUID, and
+operation-specific metadata (command, file path, port, session ID, and so on),
+so requests are traceable end to end. Enable tracing with
+`observability.traces.enabled` in your Worker configuration.
--- a/.changeset/warm-pool-batched-scale-up.md
+++ b/.changeset/warm-pool-batched-scale-up.md
@@ -1,0 +1,7 @@
+---
+'@cloudflare/sandbox': patch
+---
+
+Fill the warm pool in bounded-parallel batches instead of one container at a
+time, so it primes and recovers far faster after a burst. Tune the batch size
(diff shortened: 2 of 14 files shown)


scuffi (approved):
nice 🔥 just the one devin bug reported worth looking at imo

PR #789, opened 2026-06-26
Unify exec around single process handle

scuffi:
## Summary

This PR consolidates command execution around a single `sandbox.exec()` primitive that mirrors the Cloudflare Containers `ctx.container.exec()` contract while preserving Sandbox-specific features such as sessions, process re-attach, log replay, and readiness helpers.

`exec()` now returns a `SandboxProcessPromise`: a thenable that resolves to a `SandboxProcess` handle and also exposes `.output()`, `.text()`, `.json()`, and `.kill()` directly for the common buffered-output case.

```ts
const result = await sandbox.exec('npm test').output();
const text = await sandbox.exec('cat README.md').text();

const server = await sandbox.exec('npm run dev', { processId: 'dev' });
await server.waitForPort(3000);

const proc = await sandbox.exec('cat', { stdin: 'pipe' });
const writer = proc.stdin!.getWriter();
await writer.write(new TextEncoder().encode('hello\n'));
await writer.close();
console.log(await proc.text());
```

A new `sandbox.run()` helper keeps the buffered persistent-shell semantics that legacy `exec()` provided (`cd`, aliases, functions, exported variables persist across calls).

```ts
await sandbox.run('cd /workspace/app');
const pwd = await sandbox.run('pwd');
```

The old execution surface split closely-related functionality across multiple APIs:

- `exec()` for buffered output
- `execStream()` for SSE streaming
- `startProcess()` for background processes
- `getProcessLogs()` / `streamProcessLogs()` for log replay
- `killProcess(id)` for process control

The new shape gives users one mental model: `exec()` returns a process handle. You can buffer it, stream it, kill it, wait on it, or re-attach by id. This also positions the SDK to swap the backend to Cloudflare Containers' native exec runtime later without another public API break.

### New / changed

- `sandbox.exec(cmd, options)` returns `SandboxProcessPromise`.
- `session.exec(cmd, options)` returns `SandboxProcessPromise` and is session-pinned.
- `sandbox.run(cmd, options)` returns the legacy buffered `ExecResult` and preserves persistent shell state.
- `session.run(cmd, options)` does the same for a specific session.
- `sandbox.getProcess(id)` returns `SandboxProcess | null`.
- `sandbox.listProcesses()` returns `SandboxProcess[]` snapshots.
- `stdin` is supported:
  - `stdin: 'pipe'` -> `proc.stdin` is a `WritableStream<Uint8Array>`
  - `stdin: ReadableStream<Uint8Array>` -> stream is piped into the command
  - `stdin: string` -> utf-8 encoded into the command

### Removed from public API

- `startProcess()` -> use `exec(cmd, { processId })`
- `execStream()` -> use `(await exec(cmd)).stdout` / `.stderr`
- `execWithSessionToken()` -> use `exec(cmd, { sessionId })`
- `killProcess(id)` -> use `(await getProcess(id))?.kill()`
- `getProcessLogs(id)` -> use `(await getProcess(id))?.getLogs()`
- `streamProcessLogs(id)` -> use `(await getProcess(id))?.stdout`
- callback options (`stream`, `onOutput`, `onComplete`, `onError`) -> consume streams / promises on `SandboxProcess`

## Migration examples

```ts
// Before
const result = await sandbox.exec('echo hello');
console.log(result.stdout);

// After
const result = await sandbox.exec('echo hello').output();
console.log(result.stdout);
```

```ts
// Before
const server = await sandbox.startProcess('npm run dev');
await server.waitForPort(3000);

// After
const server = await sandbox.exec('npm run dev', { processId: 'dev-server' });
await server.waitForPort(3000);
```

```ts
// Before
await sandbox.killProcess('dev-server');

// After
(await sandbox.getProcess('dev-server'))?.kill();
```


packages/sandbox-container/src/services/execution-service.ts
```diff
@@ -60,6 +60,13 @@ export interface ExecutionStreamOptions extends ExecutionOptions {
   onEvent: (event: ExecEvent) => Promise<void>;
   commandId: string;
   background?: boolean;
+  /**
+   * Optional standard input stream piped into the spawned command. The
+   * stream is consumed by the container side; bytes flow until the stream
+   * ends or the command exits. Only honoured by streaming execution — the
+   * buffered `execute()` path always reads stdin from `/dev/null`.
```
aron-cf: What does "only honoured by streaming execution" mean?

packages/sandbox-container/src/session.ts
```diff
@@ -489,6 +502,13 @@ export class Session {
       // This ensures synchronization: shell writes PID, we read it (blocking)
       await this.createPidPipe(pidPipe);
 
+      // Create stdin FIFO if a stdin stream was provided. Must exist on
+      // disk before the bash script attempts `< "$stdinPipe"` (the redirect
+      // is a path open, not a synthesized fd).
```
aron-cf: Can you elaborate on this setup. I'm not against the write/read from disk but curious why we went that direction over a more bash like mkfifo pipe.

packages/sandbox/src/openai/index.ts
```diff
@@ -95,12 +95,14 @@ export class Shell implements OpenAIShell {
...
-        stdout = result.stdout;
-        stderr = result.stderr;
+        const result = await this.sandbox
+          .exec(command, {
+            timeout: action.timeoutMs,
+            cwd: this.cwd
+          })
+          .output({ encoding: 'utf8' });
+        stdout = result.stdout as string;
+        stderr = result.stderr as string;
```
aron-cf: If these casts are needed something has gone wrong with the typing of the result. It should be based on a type override based on the encoding argument

tests/e2e/test-worker/index.ts
```diff
@@ -19,9 +19,146 @@ import {
...
+// ---------------------------------------------------------------------------
+// Legacy SSE shims for the unified `SandboxProcess` handle.
+//
+// E2E tests pre-date the exec unification and consume the old `execStream` /
+// `streamProcessLogs` SSE frame shapes (`{ type: 'stdout' | 'stderr' | 'exit'
+// | 'complete' | 'error', data?, exitCode?, processId? }`). The shims below
+// re-encode the raw stdout/stderr byte streams + exitCode into those frames
+// so the E2E tests can stay unchanged. These should be deleted when the E2E
+// tests are migrated to consume `proc.stdout` / `proc.stderr` directly.
+// ---------------------------------------------------------------------------
```
aron-cf, line 36: Can we do the migration here so that we don't have legacy cleanup. Unless this is going to be an immediate follow up.

packages/sandbox/src/process.ts
```diff
@@ -0,0 +1,777 @@
...
+// Dependency surface
+//
+// The set of operations `SandboxProcessImpl` needs from the owning Sandbox.
+// Decoupled into an interface so the class can be unit-tested without
+// instantiating a Durable Object, and so we don't grow `sandbox.ts` further.
+// ---------------------------------------------------------------------------
+
+export interface SandboxProcessDeps {
+  /** Open the multiplexed log stream for a process (replay-then-tail). */
+  openLogStream(processId: string): Promise<ReadableStream<Uint8Array>>;
```
aron-cf: I think we should shorten this to just `id`, process id is too close to a system `pid`. Name would be another option, but that's a wider change.

packages/sandbox/src/process.ts
```diff
@@ -0,0 +1,777 @@
...
+  command: string;
+  sessionId?: string;
+  startTime: Date;
+  /** Initial status reported by the container after `startProcess`. */
+  status: ProcessStatus;
+  /**
+   * `pipe` for owners (live multiplexed stream), `replay` for re-attach.
+   * Either way we read the same `streamProcessLogs` endpoint; the flag is
+   * carried for diagnostics.
+   */
```
aron-cf: This comment seems detached from the types?

packages/sandbox/src/process.ts
```diff
@@ -0,0 +1,760 @@
...
+interface DemuxResult {
+  stdout: ReadableStream<Uint8Array> | null;
+  stderr: ReadableStream<Uint8Array> | null;
+  exitCode: Promise<number>;
+}
+
+function demultiplexLogStream(
+  source: ReadableStream<Uint8Array>,
+  options: {
+    stdout: 'pipe' | 'ignore';
```
aron-cf, line 92: I think currently when ignore is provided we actually still keep the logs around, it just means output will set stderr and stdout to null. I think it should mean drop the logs entirely...

packages/sandbox/src/process.ts
```diff
@@ -0,0 +1,760 @@
...
+    stderr: 'pipe' | 'ignore' | 'combined';
```
aron-cf, line 93: I think currently combined only applies to the streams. It should also apply to output as well.

packages/sandbox/src/process.ts
```diff
@@ -0,0 +1,760 @@
+/**
+ * Unified `SandboxProcess` implementation.
+ *
+ * Wraps the existing `ProcessService` plumbing (start / kill / getProcess /
```
aron-cf, line 4: I think getProcess currently isn't storing the output options, it always constructs the result with pipe.

aron-cf, line 4: Can we verify that the kill signal is actually passed to and used by the container implementation. I don't think it is currently.

aron-cf (approved):
I've left some comments inline, but in generally I think it's worth getting this in and iterating.

PR #797, opened 2026-06-30
Extract preview URL handling into a core module

ghostwriternr:
Move all preview URL functionality out of the Sandbox Durable Object into a dedicated module.

Preview URL handling had accumulated inside `sandbox.ts`: port exposure and token authorization, runtime activation checks, stale-runtime classification, proxy request construction, container TCP forwarding, and URL parsing/construction were all tangled into the DO facade. That made both sandbox.ts and the preview logic harder to read, test, and reason about in isolation.

This keeps preview URLs as core Sandbox functionality rather than an extension, because they touch Worker ingress, DO routing, `fetch()` interception, lifecycle hooks, container TCP forwarding, and DO storage — concerns that do not fit the RPC-only `SandboxExtension` model. The Sandbox DO retains lifecycle ownership (startup, activity/in-flight accounting, fetch interception); `PreviewService` owns the preview domain logic behind explicit dependencies, with focused helpers for proxy metadata, forwarding request construction, tokens, route parsing, and persisted state.

The change is behavior-preserving. The public API is unchanged: `exposePort`, `unexposePort`, `getExposedPorts`, `isPortExposed`, `validatePortToken`, and `proxyToSandbox` keep their existing contracts, and existing test suites continue to assert the same behavior. New unit tests cover route parsing/URL construction and the preview proxy protocol helpers.


aron-cf (approved):
👍

PR #798, opened 2026-07-01
Add sandbox container labels option

### LINKED ISSUE #790: Expose or document container labels for Sandbox SDK sandboxes
**Describe the bug**

`@cloudflare/containers` supports start-time container labels via `ContainerStartConfigOptions.labels`, and the Cloudflare Containers analytics docs expose labels for querying metrics/usage with `labels_has` and `label(name: "...")`.

The Sandbox SDK does not appear to expose a first-class way to set those labels when using `getSandbox()` and the SDK's lazy auto-start flow. `SandboxOptions` has options such as `sleepAfter`, `keepAlive`, `containerTimeouts`, and `transport`, but no `labels`/`metadata` field. 

**To Reproduce**

1. Use `@cloudflare/sandbox` in a Worker and create/reuse a sandbox:

```ts
const sandbox = getSandbox(env.Sandbox, "tenant-workspace", {
  sleepAfter: "30m"
});
```

2. Try to pass labels for analytics attribution:

```ts
const sandbox = getSandbox(env.Sandbox, "tenant-workspace", {
  sleepAfter: "30m",
  labels: {
    tenantId: "tenant_123",
    workload: "code-workspace"
  }
});
```

3. Observe that `SandboxOptions` has no `labels` field, so this is not typed/supported.
4. Trigger the first lazy-starting operation, such as `exec()`, `createSession()`, or `containerFetch()`.
5. Compare with `@cloudflare/containers`, where `ContainerStartConfigOptions` includes `labels?: Record<string, string>` and the start implementation applies `options?.labels ?? this.labels`.

**Expected behavior**

Expose labels through Sandbox options:

```ts
const sandbox = getSandbox(env.Sandbox, id, {
  labels: {
    tenantId: "tenant_123",
    workload: "code-workspace"
  }
});
```

aron-cf:
Fixes https://github.com/cloudflare/sandbox-sdk/issues/790
Container instances created through the Sandbox SDK could not be tagged with custom metadata. The underlying Cloudflare Containers platform supports key-value labels that flow into the analytics pipeline, where they can be used to filter and group usage metrics by dimensions such as tenant, environment, or workload. The SDK exposed sleep and transport options through `SandboxOptions` but offered no way to set labels, so callers had no first-class path to attribute container usage back to their own entities. This addresses the request in cloudflare/sandbox-sdk#790.

This change adds a `labels` field to `SandboxOptions`, letting callers attach labels at the same call site they already use to configure a sandbox. Labels are stored on the sandbox and applied when the container next starts, which keeps them consistent with how the platform consumes them. Changing labels while a container is already running is allowed, but the new values only take effect on the following start.

```ts
const sandbox = getSandbox(env.Sandbox, 'tenant-workspace', {
  sleepAfter: '30m',
  labels: {
    tenantId: 'tenant_123',
    workload: 'code_workspace'
  }
});
```

Label validation remains owned by the underlying Cloudflare Containers runtime, keeping Sandbox aligned with the platform label contract.

You can verify the behavior locally by calling `getSandbox()` with labels and confirming the container starts as normal. In local development the applied labels are visible on the underlying container through `docker inspect`.

Unit tests cover option forwarding, cache deduplication for repeated label configurations, order-independent label comparison, clearing labels with an empty object, Durable Object persistence/restoration, and clone behavior so caller-side mutation does not affect stored labels.

The package README and the `SandboxOptions` type documentation now describe the `labels` option and its start-time behavior.

---

<a href="https://app.devin.ai/review/cloudflare/sandbox-sdk/pull/798" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


ghostwriternr (approved):

ghostwriternr:
fixed some of the language around validation from the PR desc, rest looks great 🙌

PR #799, opened 2026-07-01
Fix for race condition in RPC connection

### LINKED ISSUE #794: Concurrent sandbox creation fails with "disposing the main stub" and leaves orphaned Running containers
**Describe the bug**

A Worker exposes HTTP endpoint that creates a sandbox per request. When several requests hit it concurrently, some of the creations fail and return HTTP 500 with:

```
RPC session was shut down by disposing the main stub
```

This happens with the account at very few running containers (10 instances), well below the instance cap. After the batch, the Containers dashboard shows container instances still in the **Running** state that do not correspond to any successfully created sandbox. They stay Running on their own and were only removed by an explicit destroy call using their id.

---

### 1. Setup

Start with the account at 0 running container instances. Deploy the Worker below.

### 2. Worker code

One HTTP RPC endpoint. The `create_workspace` case mints a fresh id, gets a sandbox, and runs the create:

```ts
// apps/sandbox-proxy/src/rpc/handler.ts
case 'create_workspace': {
  const workspaceId = crypto.randomUUID();
  const sandbox = getSandbox(env.Sandbox, workspaceId);
  try {
    const result = await workspaceOps.createWorkspace(env.BACKUP_BUCKET, sandbox, workspaceId, request.params);
    return { workspaceId, status: 'ready', ...result };
  } catch (error) {
    const cleanupSandbox = getSandbox(env.Sandbox, workspaceId);
    await workspaceOps
      .destroyWorkspace(env.BACKUP_BUCKET, cleanupSandbox, workspaceId)
      .catch((cleanupError) =>
        console.error(`Failed to clean up after a failed create_workspace (${workspaceId})`, cleanupError),
      );
    throw error;
  }
}
```

`createWorkspace` then makes the first container-touching SDK calls — `mkdir` (which starts the container), then `restoreBackup`. The failure surfaces on one of these:

```ts
// apps/sandbox-proxy/src/workspace.ts
export const createWorkspace = async (bucket, sandbox, workspaceId, options) => {
  await sandbox.mkdir(WORKSPACE_DIR, { recursive: true });
  const snapshot = await resolveSnapshotById(bucket, options.snapshotId);
  await sandbox.restoreBackup(snapshot);
  // ...
};
```

### 3. Test code

Sign the control-plane JWT once, then fire 10 POSTs of `{ method: 'create_workspace', params: { snapshotId } }` at the same time. Building all fetches before the first `await` keeps them concurrent:

```js
const requests = Array.from({ length: 10 }, () =>
  fetch(endpoint, {
    method: 'POST',
    headers: { 'content-type': 'application/json', authorization: `Bearer ${token}` },
    body: JSON.stringify({ method: 'create_workspace', params: { snapshotId } }),
  }).then(async (res) => ({ status: res.status, body: await res.text() })),
);
const results = await Promise.all(requests);
```

### 4. Logs

One run (each line is one of the 10 concurrent requests; `start`/`end` are ms from batch start):

```
[00] FAIL http=500 start=+0ms   end=+7544ms   RPC session was shut down by disposing the main stub
[01] OK   http=200 start=+13ms  end=+4005ms   workspaceId=4119ca0c-... status=ready
[02] FAIL http=500 start=+14ms  end=+5270ms   RPC session was shut down by disposing the main stub
[03] OK   http=200 start=+14ms  end=+5684ms   workspaceId=216890c6-... status=ready
[04] OK   http=200 start=+14ms  end=+4912ms   workspaceId=1014ec88-... status=ready
[05] OK   http=200 start=+14ms  end=+5614ms   workspaceId=84233253-... status=ready
[06] OK   http=200 start=+14ms  end=+4882ms   workspaceId=ffc8fe21-... status=ready
[07] OK   http=200 start=+14ms  end=+7545ms   workspaceId=19c6cab9-... status=ready
[08] FAIL http=500 start=+15ms  end=+10975ms  RPC session was shut down by disposing the main stub
[09] OK   http=200 start=+15ms  end=+5267ms   workspaceId=b97eaeef-... status=ready
---- summary: 7/10 ok · 3/10 failed ----
```

Failed responses return quickly (~2–5s); successful ones take longer (~5–10s).

Counts across the batches run so far (only a few batches were run):

| Worker SDK | Image  | succeeded / 10 | left Running |
|------------|--------|----------------|--------------|
| 0.11.0     | 0.11.x | 6              | 3            |
| 0.11.0     | 0.12.2 | 8              | 1            |
| 0.12.2     | 0.12.2 | 7              | 1            |

Running the 10 creations sequentially (one at a time) produces no failures.

### 5. Dashboard

After the batch, one or two instances remain in the **Running** state with no matching successfully created sandbox.

---

## Expected behavior

A concurrent batch of 10 creations, with the account well below the instance cap, should create all 10. A creation that cannot complete should not leave a container instance in the Running state.

---

## Environment

- `@cloudflare/sandbox` 0.12.2 (also observed on 0.11.0)
- container image `cloudflare/sandbox:0.12.2`
- instance type `basic`

aron-cf:
Fixes https://github.com/cloudflare/sandbox-sdk/issues/794

Concurrent sandbox startup (e.g. creating many workspaces from a snapshot at once) intermittently failed, and the errors callers saw were misleading. This PR fixes two related problems on the RPC transport: an idle-disconnect race, and the way container **capacity/startup** failures are surfaced.

## 1. Idle-disconnect race disposing the main stub

The RPC client tracks busy/idle by polling capnweb session stats. During concurrent startup those stats can briefly look like the idle baseline (`imports=1, exports=1`) while a method promise is still in flight, so the client could arm the idle-disconnect timer and dispose the capnweb main stub before the pending call resolved — surfacing as:

```text
RPC session was shut down by disposing the main stub
```

**Fix:** pending RPC method calls are now part of the client's busy/idle state via an `activeCalls` counter. A wrapped method increments it before invoking, marks the session busy, and only decrements on settle. The session transitions to idle (and arms disconnect) only when `activeCalls === 0` **and** capnweb stats are at the idle threshold — preserving the stats-based behavior for long-lived stream exports.

## 2. Capacity/startup failures were masked

When the Containers platform could not admit a container ("There is no container instance…", "Maximum number of running container instances exceeded"), the failure reached callers as a generic `utils.createSession` interruption — or, later, as a raw `RPC session was shut down…` string — with no actionable signal.

**Fixes:**

- **Explicit, bounded start.** On the RPC transport the control connection now starts the container explicitly (via `startAndWaitForPorts`) **before** opening the WebSocket, under a short instance-get budget. Each attempt fails fast under capacity pressure and hands back to the connection's own retry loop (exponential backoff within the startup budget) instead of one attempt blocking ~30s and letting the DO be evicted mid-wait. Port readiness still uses the full timeout so a genuinely-booting app isn't cut short.
- **Typed errors, never raw internals.** The start hook always throws a typed `SandboxError`: a retryable `ContainerUnavailableError` for capacity/admission failures (carrying the real platform message and a `reason` — `no_container_instance_available` vs `max_container_instances_exceeded`, via the new exported `ContainerUnavailableReason`), or an `INTERNAL_ERROR`-coded error carrying the real cause otherwise. Detection is case-insensitive and realm-safe.
- **Honest interruption semantics.** A queued RPC call is only reported as `OPERATION_INTERRUPTED` when the session actually established a connection to a running container and was then interrupted. If the container never started, the underlying typed error is surfaced instead of a masked interruption.
- **HTTP path parity.** `Sandbox.containerFetch` emits a structured `CONTAINER_UNAVAILABLE` 503 (preserving the original platform message) for these failures.

## Caveat

This does not change the platform reality that under `max_instances` pressure some concurrent starts can't get an instance. Those now surface as clean, retryable `ContainerUnavailableError`s across the retry budget rather than misleading interruptions or raw transport strings.


ghostwriternr (approved):
simplified the changesets, overall looks great! some of these changes need to go back into `next` too

PR #809, opened 2026-07-03
Move extensions into their own directory

scuffi:
# Summary

Extensions have been living inside `packages/sandbox/src/` since we introduced the framework, which made it hard to tell at a glance what's an extension versus core SDK code. This moves the two real `SandboxExtension`s we ship (`interpreter` and `git`) out to a top-level `extensions/` directory.

Purely a file move. Same npm package, same subpath imports (`@cloudflare/sandbox/interpreter`, `@cloudflare/sandbox/git`), same `dist/` layout, byte-identical sidecar tarball. `openai`/`opencode`/`xterm` stayed put since they aren't `SandboxExtension`s yet.

Most of the diff is wiring: tsdown entries, tsconfig `include` + path aliases so `@cloudflare/sandbox/{extensions,sidecar,errors}` resolves to source during dev, matching vitest aliases, and the sidecar build/test scripts pointing at the new paths.

Two small bits that aren't strictly moves:

- `gitCredentialProxyHandler` is baked into `ContainerProxy` (core), not really part of the extension, so it moved to `src/storage-mount/outbound/` instead of into `extensions/git/` — otherwise we'd have a circular workspace import.
- Added `@cloudflare/sandbox/errors` as a public subpath (just re-exports the existing module) so the git extension can import `createErrorFromResponse` without reaching into private paths.

Skill doc updated with the new layout and authoring checklist. All 863 unit tests + 56 sidecar tests pass, and every workspace typechecks, including the E2E test worker that consumes both subpaths.


aron-cf (approved):
Thanks!

.changeset/extensions-out-of-src.md
```diff
@@ -0,0 +1,5 @@
+---
+'@cloudflare/sandbox': patch
+---
+
+Move the built-in extension sources (`interpreter/`, `git/`) out of `packages/sandbox/src/` and into a top-level `extensions/` directory so it's clear which code is an extension versus core SDK. Public subpath imports (`@cloudflare/sandbox/interpreter`, `@cloudflare/sandbox/git`) and their APIs are unchanged.
```
aron-cf: This doesn't affect external users, so no changeset required.

PR #827, opened 2026-07-17
Stabilize process streams and releases

ghostwriternr:
Keep process log and port-watch subscriptions connected until they settle. Worker RPC stream capabilities do not reliably preserve retained event delivery, so observations now pull events explicitly and hold runtime control until the local subscription finishes or is canceled.

Serialize branch releases so an older build cannot replace a newer Worker and container deployment. Also clean up partially initialized E2E sandboxes, make the terminal assertion insensitive to asynchronous prompts, and declare the internal execution package version required by preview publishing.

The rewritten history removes the diagnostic probes and reverted experiments used to isolate the Worker RPC stream boundary. Its final tree is unchanged from the previously reviewed branch.


ghostwriternr:
Temporarily closing to run the existing privileged cleanup for the saturated disposable E2E environment. Will reopen after cleanup completes.

PR #835, opened 2026-07-22
Update Devin Outpost to conditionally issue R2 head request

aron-cf:
A Devin session container runs `devin-checkpoint restore` on start, which issues a `HEAD /checkpoint` to the internal proxy to decide whether to restore a saved workspace. `CheckpointProxy` answered that by calling `DEVIN_CHECKPOINTS.head(key)`. For any session that has never suspended, `key` (`sessions/<sessionId>.tar.zst`) doesn't exist yet — R2's `.head()` returns `null` and the proxy correctly returns `404`, but R2's observability layer still records the underlying `HeadObject` miss as an **error-level `r2_head` span** (`code 10007`, "The specified key does not exist"). The result is a noisy error span on every fresh session even though nothing failed.

This makes the session's Durable Object the source of truth for checkpoint existence, so the restore probe never issues a `HeadObject` against a missing key:

```
DevinWorker (DO):
  recordCheckpointSaved()  -> storage.put("checkpointPresent", true)   // set after a successful PUT
  hasCheckpoint()          -> storage.get("checkpointPresent") === true
  stop(reason="terminated")-> deleteCheckpoint(...); storage.delete("checkpointPresent")

CheckpointProxy:
  HEAD -> status = (await worker(sessionId).hasCheckpoint()) ? 200 : 404   // was: DEVIN_CHECKPOINTS.head(key)
  PUT  -> ...put(key, body); await worker(sessionId).recordCheckpointSaved()
  GET  -> unchanged (streams from R2)
```

`CheckpointProxy` reaches the DO via `env.DevinWorker.get(idFromName(sessionId))`, the same deterministic per-session id the reconciler uses. The container scripts (`devin-checkpoint`, `start-devin-worker`) are untouched, so restore behavior is identical — a fresh session still gets `404 → "no checkpoint found" → start fresh`, just without the spurious error span. `GET` still reads from R2 and only runs after a `200` HEAD, so it only touches keys known to exist.

> [!NOTE]
> This code will be removed once the backup/restore flow is able to use native snapshotting.

---

<a href="https://app.devin.ai/review/cloudflare/sandbox-sdk/pull/835" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/devin/src/index.ts
+++ b/devin/src/index.ts
@@ -4,6 +4,8 @@ import { reconcile } from './reconcile';
 
 export { CheckpointProxy } from './persistence';
 
+const CHECKPOINT_PRESENT_KEY = 'checkpointPresent';
+
 /**
  * One Durable Object per Devin session. The DO is deliberately dumb: it knows
  * how to start/stop its Cloudflare Container, but it does not call Devin and it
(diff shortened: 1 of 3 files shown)


aron-cf:
@scuffi not sure if this is needed, but I kept seeing this error in the traces when the R2 HEAD request returns 404, which made me think things were broken.

image

aron-cf:
Alternatively we could do a `.list()` call with a prefix to filter only the bucket. I _think_ not making any request at all is marginally nicer.

scuffi:
imo we should `.list()` or something to validate it exists, or we will get errors if people delete buckets, migrate, etc, feels safer, wdyt?

aron-cf:
@agent-think can you review this PR and associated context and open an alternative PR that implements the proposed `.list()` approach instead of calling `.head()` retain the same regression tests introduced in this PR.

PR #841, opened 2026-07-22
Fix stale preview URLs

### LINKED ISSUE #829: STALE_PREVIEW_URL on every preview request despite healthy container (0.11.0)
## Summary

`sandbox.exposePort()` succeeds (its `assertActive(runtime)` passes), but every request to the returned preview URL returns `410 STALE_PREVIEW_URL` ("Preview URL is stale because the sandbox runtime is not active") — even while the same sandbox's container is verifiably healthy and serving traffic.

## Environment

- `@cloudflare/sandbox` 0.11.0 (worker-side and DO)
- Container image: `cloudflare/sandbox:0.11.0-desktop`
- Worker deployed with a wildcard route (`*.example.com/*`) + `run_worker_first: true`; `proxyToSandbox` middleware intercepts preview hostnames
- Sandbox accessed via `getSandbox(env.Sandbox, taskId)` (also tried `{ keepAlive: true }`)

## Repro

1. Custom domain worker; wildcard DNS exists and TLS is valid (one level under the zone apex).
2. `const sandbox = getSandbox(env.Sandbox, id);`
3. `const { url } = await sandbox.exposePort(6080, { hostname: "example.com" })` → returns `https://6080-<id>-<token>.example.com/` (HTTP 200).
4. `GET <url>vnc.html` → `410 {"error":"Preview URL is stale because the sandbox runtime is not active","code":"STALE_PREVIEW_URL"}`.

The stale response persists for 4+ minutes of polling and across repeated `exposePort()` re-registrations (same token returned each time). `keepAlive: true` does not change it.

## Expected

Preview URL forwards to the container port once the container is running.

## Notes

- The same sandbox executes commands fine via `exec` (container boots, agent responds), so the runtime is active in practice.
- From reading the DO code, `fetchPreviewIfRunning` returns stale when `!container?.running || state.status !== "healthy"` or `!currentRuntime.isActive(runtime)`. Since `exposePort()`'s `assertActive(runtime)` passes, the registration-side state disagrees with the proxy-side state somewhere (possibly a runtime-id mismatch after DO migration/restart, or the health gate never reaching `healthy` on the desktop image).
- Same behavior with the SDK's `getDesktopStreamUrl(hostname)` convenience on the desktop image.
- Happy to provide a minimal repro repo if useful.

aron-cf:
Fixes #829: `exposePort()` succeeds but every request to the preview URL returns `410 STALE_PREVIEW_URL` after a container restart — even while the container is healthy and serving traffic.

**Root cause.** Preview forwarding is gated on a *runtime identity* that rotates on every container start: `onStart()` calls `currentRuntime.markStarted()`, minting a fresh id. `exposePort()` writes a runtime-scoped activation stamped with the id current at that instant, and `onStop()` clears activations. Nothing re-derived the activation on start, so after any restart the proxy path saw `runtime.owns(activation) === false` and returned stale:

```ts
// validatePreviewURLForRuntime()
if (!runtime.owns(activation)) return { status: 'stale', reason: 'runtime-mismatch' };
```

Because `exec`/`containerFetch` *starts* the container (rotating the id via `onStart`), health-checking with `exec` actively re-invalidated the just-exposed port — which is why re-exposing appeared not to help and staleness persisted for minutes. Durable auth (`portTokens`) already survives restarts; only the runtime binding was missing.

**Fix.** Re-derive activations from the durable `portTokens` for the new runtime as part of `onStart()`:

```ts
override async onStart() {
  const runtime = await this.currentRuntime.markStarted();
  await this.reactivateExposedPortsForRuntime(runtime); // new
  ...
}

private async reactivateExposedPortsForRuntime(runtime: RuntimeIdentity) {
  await this.ctx.storage.transaction(async (txn) => {
    const tokens = await this.readPortTokens(txn);
    const activations: PreviewPortActivations = {};
    for (const [port, entry] of Object.entries(tokens))
      activations[port] = runtime.scope({ token: entry.token });
    await this.writeActivePreviewPorts(activations, txn);
  });
}
```

This runs only for a container that is already starting, so it never wakes a stopped runtime, and it never resurrects unexposed/destroyed ports (those are absent from `portTokens`) — preserving the security intent of #708.

---

<a href="https://app.devin.ai/review/cloudflare/sandbox-sdk/pull/841" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/.changeset/preview-url-survives-restart.md
+++ b/.changeset/preview-url-survives-restart.md
@@ -1,0 +1,11 @@
+---
+'@cloudflare/sandbox': patch
+---
+
+Keep preview URLs working across container restarts. Previously, `exposePort()`
+succeeded but preview requests returned `410 STALE_PREVIEW_URL` after any
+container restart (including `exec` calls that woke a stopped container),
+because the runtime-scoped activation was cleared on stop and only re-created by
+a manual `exposePort()` call. Exposed ports are now automatically reactivated
+for the new runtime when the container starts, so a preview URL keeps forwarding
+without re-exposing. Ports that were unexposed or destroyed are not resurrected.
--- a/packages/sandbox/src/sandbox.ts
+++ b/packages/sandbox/src/sandbox.ts
@@ -2876,7 +2876,8 @@ export class Sandbox<Env = unknown> extends Container<Env> implements ISandbox {
   override async onStart() {
     this.logger.debug('Sandbox started');
 
-    await this.currentRuntime.markStarted();
+    const runtime = await this.currentRuntime.markStarted();
+    await this.reactivateExposedPortsForRuntime(runtime);
 
     // Fire-and-forget: version check is observability, not load-bearing.
     this.checkVersionCompatibility().catch((error) => {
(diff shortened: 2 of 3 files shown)


aron-cf:
@ghostwriternr I've added a fix for the reported issue, however I do want to check that this is a bug, rather than the port code acting as intended...

ghostwriternr:
Closing this PR for now as discussed async. The preview URL behaviour described is an intentional change i.e. if the container restarts, the processes running within are lost anyway until manual restarts so any exposed ports were potentially pointing to processes that don't exist, so we don't automatically re-expose.

But the issue suggests the `getSandbox` -> `exposePort` -> hit the URL flow fails, which is not ideal and definitely a bug. The fix for that is elsewhere, and should be debugged separately.

PR #866, opened 2026-08-13
Fix credential proxy copy scope

scuffi:
# Summary

This makes sure copy requests through `credentialProxy` stay inside the mounted bucket and prefix.

It checks the S3/R2 and GCS copy headers before signing, including multipart copies, while keeping normal in-scope copies working.


aron-cf (approved):
Thanks for fixing

PR #867, opened 2026-08-13
Backport credential proxy copy fix

scuffi:
Backports #866 to `next`.

This keeps copy requests through `credentialProxy` inside the mounted bucket and prefix. It's the same fix and tests, adjusted for the storage mount files on `next`.


aron-cf (approved):
Thanks for fixing

